PR #2, opened 2015-03-11
Review for DM-1576

TallJimbo:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+    fluxKey = schema[fluxField].asKey()
+    try:
+        fluxErrKey = schema[fluxErrField].asKey()
+    except Exception:
+        fluxErrKey = None
+    return (fluxKey, fluxErrKey)
+
+class LoadReferenceObjectsConfig(pexConfig.Config):
+    pixelMargin = pexConfig.RangeField(
+        doc = "Padding to add to 4 all edges of the bounding box (pixels)",
```
TallJimbo: "all 4" (or maybe "all four") rather than "4 all"

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+        dtype = int,
+        default = 50,
+        min = 0,
+    )
+    defaultFilter = pexConfig.Field(
+        doc = "Default reference catalog filter to use if filter not specified in exposure; " + \
+            "if blank then filter must be specified in exposure",
```
TallJimbo: We may have clashed on this before, in which case I'm not going to fight you, but I find:
```
("line 1..... "
 "...line 2")
```
more readable than ` + \` for multi-line strings.  It's also more efficient, I believe, since it makes the concatenation happen during compilation to .pyc, not at runtime, but of course that's not a big deal in these sorts of cases.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+    - photometric (optional): is the object usable for photometric calibration?
+    - resolved (optional): is the object spatially resolved?
+    - variable (optional): does the object have variable brightness?
+
+    @section meas_algorithms_loadReferenceObjects_Config       Configuration parameters
+
+    See @ref LoadReferenceObjectsConfig for a base set of configuration parameters.
+    Most subclasses will add configuration variables.
+    """
+    __metaclass__ = abc.ABCMeta
```
TallJimbo: Are we getting anything out of the `abc` module that we couldn't just get from doing `raise NotImplementedError`() ourselves?  Or do you think this is just more idiomatic?

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+    @copydoc loadObjectsInBBox
+
+    @section meas_algorithms_loadReferenceObjects_Schema       Schema of the reference object catalog
+
+    Reference object catalogs are instances of lsst.afw.table.SimpleCatalog with the following schema
+    (other fields may also be present):
+    - coord: position of star on sky (an lsst.afw.coord.IcrsCoord)
+    - centroid: position of star on an exposure, if relevant (an lsst.afw.Point2D)
+    - hasCentroid: is centroid usable?
+    - *referenceFilterName*_flux: brightness in the specified reference catalog filter: 10^(-0.4*mag)
```
TallJimbo: Given you've got code for Janskys elsewhere (not maggies), I think the `10^(-0.4*mag)` is a relic.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+    - *referenceFilterName*_fluxSigma (optional): brightness standard deviation;
+        omitted if no data is available; possibly nan if data is available for some objects but not others
+    - *cameraFilterName*_camera_flux: brightness in specified camera filter (magnitude)
```
TallJimbo: Up at the top of the file, you have some code that deals with fields like "*_camFlux", but you've got "*_camera_flux" here.  I think those probably should be consistent.  I much prefer "*_camera_flux".

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+    - *cameraFilterName*_camera_fluxSigma (optional): brightness standard deviation
+        in specified camera filter (magnitude); omitted if no data is available;
+        possibly nan if data is available for some objects but not others
+    - default_flux (optional): brightness to use if no camera filter is available (magnitude);
+        omitted unless defaultFilter is specified in the config
```
TallJimbo: What does "(magnitude)" mean here?

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+        """Add aliases for camera filter fluxes to the schema
+
+        If self.config.defaultFilter then adds these aliases:
+            camFlux:      <defaultFilter>_flux
+            camFluxSigma: <defaultFilter>_fluxSigma, if the latter exists
+
+        For each camFilter: refFilter in self.config.filterMap adds these aliases:
+            <camFilter>_camFlux:      <refFilter>_flux
+            <camFilter>_camFluxSigma: <refFilter>_fluxSigma, if the latter exists
+
```
TallJimbo: More "camFlux" vs. "camera_flux" here.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -0,0 +1,252 @@
...
+        aliasMap = schema.getAliasMap()
+
+        def addAliasesForOneFilter(filterName, refFilterName):
+            """Add aliases for a single filter
+
+            @param[in] filterName  camera filter name, or ""
+                the name is <filterName>_camFlux or camFlux if filterName is None
+            @param[in] refFilterName  reference filter name; <refFilterName>_flux must exist
+            """
+            camFluxName = filterName + "_camFlux" if filterName is not None else "camFlux"
```
TallJimbo: And here.

PR #7, opened 2015-04-22
Wrap matplotlib.pyplot import in RuntimeError in addition to ImportError

wmwv:
Passes buildbot 
http://lsst-buildx.ncsa.illinois.edu:8010/builders/DM_stack/builds/5664

Moderate discussion of how to do plotting "right" in general and what should be caught, etc., and what should be wrapped by a general plotting import function at
https://jira.lsstcorp.org/browse/DM-2588
But I recommend that that discussion be preserved for a future general fix, and this simple catching of both ImportException, RuntimeException be accepted for a pragmatic fix for this.

### DIFF
--- a/python/lsst/meas/algorithms/objectSizeStarSelector.py
+++ b/python/lsst/meas/algorithms/objectSizeStarSelector.py
@@ -25,7 +25,7 @@ import numpy
 try:
     import matplotlib.pyplot as pyplot
     fig = None
-except ImportError:
+except (ImportError, RuntimeError):
     pyplot = None
 
 import lsst.pex.config as pexConfig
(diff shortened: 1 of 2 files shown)


python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -25,7 +25,7 @@
 try:
     import matplotlib.pyplot as pyplot
     fig = None
-except ImportError:
+except (ImportError, RuntimeError):
```
timj: DM-3856 is triggering an `IOError` with the font cache.

jdswinbank:
Closing PR without merging; see https://jira.lsstcorp.org/browse/DM-2588?focusedCommentId=42997

PR #8, opened 2015-05-12
Added setMatchDistance function and a unit test

boutigny:
Clarify test logic a bit

### DIFF
--- a/python/lsst/meas/algorithms/setMatchDistance.py
+++ b/python/lsst/meas/algorithms/setMatchDistance.py
@@ -20,22 +20,23 @@ from __future__ import absolute_import, division, print_function
 # the GNU General Public License along with this program.  If not, 
 # see <http://www.lsstcorp.org/LegalNotices/>.
 #
+__all__ = ["setMatchDistance"]
 
-from .algorithmsLib import *
-from .defects import *
-from .psfDeterminerRegistry import *
-from .setMatchDistance import *
-from .starSelectorRegistry import *
-from .findCosmicRaysConfig import *
-from .detection import *
-from .gaussianPsfFactory import *
-from .loadReferenceObjects import *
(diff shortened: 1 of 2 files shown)


tests/testSetMatchDistance.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Use involves setting one class attribute:
+    * MatchClass: match class, e.g. ReferenceMatch or SourceMatch
+
+    This test is a bit messy because it exercises two templates of MatchClass
+
+    The test creates source and reference object catalogs that intentionally have
+    some separation in on-sky coordinates. The reference catalog is set using
+    a uniform grid of pixel positions and a simple WCS to compute on-sky coordinates.
+    The source catalog is created by using a distored version of the same grid
```
boutigny: distored --> distorted

python/lsst/meas/algorithms/setMatchDistance.py
```diff
@@ -0,0 +1,42 @@
...
+# You should have received a copy of the LSST License Statement and 
+# the GNU General Public License along with this program.  If not, 
+# see <http://www.lsstcorp.org/LegalNotices/>.
+#
+__all__ = ["setMatchDistance"]
+
+def setMatchDistance(matches):
+    """Set the distance field of the matches in a match list to the distance in radians on the sky
+
+    @warning the coord field of the source in each match must be correct
```
boutigny: Shouldn't you add a try / except block here to warn the caller that the coord field is missing ?

r-owen: I think the existing message from afwTable suffices. I tried changing the name of a field from "coord" to "coords" and got this error:
{code}
KeyError: "Field 'coords' not found in Schema."
{code}
The message doesn't say which catalog is missing the field, but the traceback does show that.

PR #10, opened 2015-05-27
meas_algorithms tickets/DM-2535: switch from compound fields to FunctorKeys

TallJimbo:


include/lsst/meas/algorithms/KernelPsfFactory.h
```diff
@@ -13,6 +13,7 @@
 #include "lsst/afw/table/io/InputArchive.h"
 #include "lsst/afw/table/io/OutputArchive.h"
 #include "lsst/afw/table/io/CatalogVector.h"
+#include "lsst/afw/table/aggregates.h"
```
r-owen: That's a lot of detailed imports. If it does not cause too much build slowdown or bloat I suggest importing table.h

src/DoubleGaussianPsf.cc
```diff
@@ -29,6 +29,7 @@
 #include "lsst/afw/table/io/OutputArchive.h"
 #include "lsst/afw/table/io/InputArchive.h"
 #include "lsst/afw/table/io/CatalogVector.h"
+#include "lsst/afw/table/aggregates.h"
 #include "lsst/afw/detection/PsfFormatter.h"
```
r-owen: Same suggestion as above: include table.h if reasonable

src/SingleGaussianPsf.cc
```diff
@@ -37,6 +37,7 @@
 #include "lsst/afw/table/io/InputArchive.h"
 #include "lsst/afw/table/io/OutputArchive.h"
 #include "lsst/afw/table/io/CatalogVector.h"
+#include "lsst/afw/table/aggregates.h"
```
r-owen: And again...table.h if reasonable

tests/ticket2986.py
```diff
@@ -46,8 +46,10 @@ def test(self):
 
         new = ccds.addNew()
         new.set("id", 0)
-        new.set("bbox.min", afwGeom.Point2I(0,0))
-        new.set("bbox.max", afwGeom.Point2I(1024,1024))
+        new.set("bbox_min_x", 0)
+        new.set("bbox_min_y", 0)
+        new.set("bbox_max_x", 1024)
+        new.set("bbox_max_y", 1024)
```
r-owen: Is there a nicer way to do this, e.g. using a bbox key and setting it from a bbox (I realize those are a pain to construct, but I think it would still be clearer and more robust than setting 4 fields independently).

include/lsst/meas/algorithms/KernelPsfFactory.h
TallJimbo: This is still probably only 50% of the code in table.h, and we certainly don't want all of that.  We could add InputArchiive.h and OutputArchive.h to CatalogVector.h, as any header that uses the latter will always want the former, but I figured that was going against the spirit of the consensus of RFC-52 (even though that's about EUPS dependencies, not headers).

In any case, in general we should absolutely not be using the full-package headers anywhere in our C++ code - that's the path to unnecessary code bloat.  IMO those should only be used in Swig files, where we're essentially forced to import all headers in a package, and if we someday address that problem we should remove them.

tests/ticket2986.py
TallJimbo: If this wasn't test code, I'd agree, and create a pair of Point FunctorKeys - and in the long run, we probably want a Box FunctorKey, though that's beyond the scope of this issue.

But even if we had that, I don't it'd be more concise than what's already there, and in this context the performance advantage of setting them at the same time goes away.

TallJimbo:
Merged to master on tickets/DM-1766.

PR #13, opened 2015-09-09
Review of DM-3387 ("Make use of good pixel count when building CoaddPsfs")

jdswinbank:


src/CoaddPsf.cc
```diff
@@ -152,9 +152,18 @@ CoaddPsf::CoaddPsf(
 {
     afw::table::SchemaMapper mapper(catalog.getSchema());
     mapper.addMinimalSchema(afw::table::ExposureTable::makeMinimalSchema(), true);
+
+    // copy the field "goodpix", if available, since computeAveragePosition will use it, if available
+    try {
+        afw::table::Key<int> goodPixKey = catalog.getSchema()["goodpix"]; // auto does not work
+        mapper.addMapping(goodPixKey, true);
+    } catch (pex::exceptions::NotFoundError &) {}
```
jdswinbank: A pretty trivial point, but: you've used "if available" twice one one line above, plus a further twice in the commit message, plus one instance of "if it was available". None of them are actually wrong, but all that repetition makes things pretty hard to read! Could you spend a few seconds rephrasing for clarity?

tests/CoaddPsf.py
```diff
@@ -28,27 +28,16 @@
    python CoaddPsf.py
 """
 
-import os, sys
-from math import *
-import numpy
```
jdswinbank: Thank you for all the cleaning up you've done in this file!

tests/CoaddPsf.py
```diff
@@ -534,6 +510,48 @@ def testValidPolygonPsf(self):
...
+        for i, (xOff, yOff, numGoodPix) in enumerate((
+            (30.0, -20.0, 25),
+            (32.0, -21.0, 10),
+            (28.0, -19.0, 30),
+        )):
+            xwsum -= xOff * numGoodPix
+            ywsum -= yOff * numGoodPix
+            wsum += numGoodPix
+            record = mycatalog.getTable().makeRecord()
+            record.setPsf(measAlg.DoubleGaussianPsf(25, 25, 10, 1.00, 0.0))
```
jdswinbank: When I see numbers hard-coded like this in a test case, I often find it helpful if there's an accompanying comment explaining why those particular numbers were chosen -- are they completely arbitrary placeholders, or do they have some significance regarding the test conditions? In this case, it's pretty clear what's happening so I don't really object, but maybe it's worth considering.

tests/CoaddPsf.py
```diff
@@ -534,6 +510,48 @@ def testValidPolygonPsf(self):
...
+        crval1, crval2 = 0.0, 0.0
+        bboxSize = afwGeom.Extent2I(2000, 2000)
+        crpix = afwGeom.PointD(bboxSize/2.0)
+        crval = afwCoord.Coord(afwGeom.Point2D(crval1, crval2))
+        wcsref = afwImage.makeWcs(crval, crpix, cd11, cd12, cd21, cd22)
+
+        schema = afwTable.ExposureTable.makeMinimalSchema()
+        schema.addField("weight", type="D", doc="Coadd weight")
+        schema.addField("goodpix", type="I", doc="Number of good pixels")
+        mycatalog = afwTable.ExposureCatalog(schema)
```
jdswinbank: Being especially picky, I might question whether some of the variable names here match the coding standard. Should it be `myCatalog`, `wSum`, etc?

r-owen: I agree, but they are pre-existing names and things like crpix are a borderline case. I'd prefer to leave it alone.

src/CoaddPsf.cc
r-owen: OK

tests/CoaddPsf.py
r-owen: OK

PR #14, opened 2015-10-12
Only take median position when points exist.

timj:
Previously, when attempted to take the median position of a list of points we
relied on the (old, pre-1.10) NumPy behaviour of providing NaN as the median
of an empty array. Newer NumPy will raise an ImportError instead. Here, we
explicitly check that we have at least one point before taking the median.

### DIFF
--- a/python/lsst/meas/algorithms/objectSizeStarSelector.py
+++ b/python/lsst/meas/algorithms/objectSizeStarSelector.py
@@ -192,7 +192,12 @@ def _kcenters(yvec, nCluster, useMedian=False, widthStdAllowed=0.15):
             break
 
         for i in range(nCluster):
-            centers[i] = func(yvec[clusterId == i])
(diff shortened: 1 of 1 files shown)


python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -192,7 +192,10 @@ def _kcenters(yvec, nCluster, useMedian=False, widthStdAllowed=0.15):
             break
 
         for i in range(nCluster):
-            centers[i] = func(yvec[clusterId == i])
+            if (clusterId == i).any():
```
timj: This seems like magic without a comment explaining what is going on. The `clusterId == i` is also used later on so maybe it would look less magic if a variable was used to avoid the repetition?

jdswinbank: Fair point!

PR #15, opened 2016-01-26
U/lauren/dm 4821

rearmstr:


rearmstr:
I would really like to have a better name than footprintBackground.  When I wrote the code, I didn't have anything better, but maybe you can come up with something more descriptive?

rearmstr:
I would probably squash the second commit.  It was just a mistake in the config settings so I don't think we need to preserve it.

laurenam:
The only descriptive names I can think of get quite long. What about tempLocalBackground?

laurenam:
I'll squash those two commits.

rearmstr:
I think tempLocalBackground is much better.

laurenam:
Ok, I'll change it.  I'm inclined to make the name change a separate commit as it may help anyone looking to see if this functionality was ported over.  This way, at least searching through git history for the original name could lead them to the answer.  Do let me know if you disagree!

rearmstr:
Sounds good.

jdswinbank:
Rebased version of this has been merged from tickets/DM-4821.

PR #17, opened 2016-02-22
Review for DM-5161: Support a full background model when detecting cosmic rays

jdswinbank:


src/CR.cc
```diff
@@ -82,7 +82,13 @@ struct IdSpanCompar : public std::binary_function<const IdSpan::ConstPtr, const
         } else if(a->id > b->id) {
             return false;
         } else {
-            return (a->y < b->y) ? true : false;
+            if (a->y < b->y) {
```
jdswinbank: This is now "sort by ID, then by y, then by x" -- can you update the docstring to reflect that please?

laurenam: Updated.

PR #22, opened 2016-03-19
DM-5532

r-owen:


python/lsst/meas/algorithms/starSelector.py
```diff
@@ -0,0 +1,154 @@
...
+                    psfCandidate.setHeight(self.config.kernelSize + 2*self.config.borderWidth)
+
+                im = psfCandidate.getMaskedImage().getImage()
+                vmax = afwMath.makeStatistics(im, afwMath.MAX).getValue()
+                if not np.isfinite(vmax):
+                    continue
+                psfCandidateList.append(psfCandidate)
+                goodStarCat.append(star)
+            except Exception as err:
+                self.log.warn("Failed to make a psfCandidate from star %d: %s" % (star.getId(), err))
```
TallJimbo: It might be better to just wrap the call to `makePsfCandidate` in a try block, as I don't think we expect it to be possible that the rest could throw (and hence we'd want the exception to propagate up and cause an early failure if it did, so we could fix the bug).

r-owen: Good idea.

python/lsst/meas/algorithms/starSelector.py
```diff
@@ -0,0 +1,154 @@
...
+        ],
+    )
+
+
+class StarSelectorTask(pipeBase.Task):
+    """!Base class for star selectors
+    """
+    __metaclass__ = abc.ABCMeta
+
+    usesMatches = False #
```
TallJimbo: Missing comment, or orphaned `#`?

python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -317,18 +339,15 @@ def selectStars(self, exposure, sourceCat, matches=None):
...
 
-        # create a log for my application
-        logger = pexLogging.Log(pexLogging.getDefaultLog(), "meas.algorithms.objectSizeStarSelector")
-
         detector = exposure.getDetector()
         pixToTanXYTransform = None
         if detector is not None:
-            tanSys = detector.makeCameraSys(cameraGeom.TAN_PIXELS)
+            tanSys = detector.makeCameraSys(TAN_PIXELS)
             pixToTanXYTransform = detector.getTransformMap().get(tanSys)
```
TallJimbo: Is there any way we could share this distortion logic across multiple star selectors?  I suspect virtually all would want to make use of it.

Maybe an intermediate base class, or utility code somewhere else?

Could be a new issue if it's substantial work, but we should at least create the issue and describe it while these classes are fresh in your mind.

jhoblitt:
Can one of the admins verify this patch?

python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -450,37 +469,12 @@ def selectStars(self, exposure, sourceCat, matches=None):
...
-                        ds9.dot("o", source.getX() - mi.getX0(), source.getY() - mi.getY0(),
-                                size=4, frame=frame, ctype=ds9.CYAN)
-                except Exception as err:
-                    logger.logdebug("Failed to make a psfCandidate from source %d: %s" % (source.getId(), err))
-
-        return psfCandidateList
-
-starSelectorRegistry.register("objectSize", ObjectSizeStarSelector)
+        starCat = SourceCatalog(sourceCat.schema)
+        for isStellar, source in zip(stellar, [s for g, s in zip(good, sourceCat) if g]):
```
TallJimbo: Logic in this loop argument is complex enough that it probably merits a comment.

r-owen: In my opinion the code is confusing and needs a cleanup pass. However, I will break this one line into two (make the list, then use it). It is still not obvious to me why `stellar` and `good` have the same length (e.g. from reading the code above).

python/lsst/meas/algorithms/objectSizeStarSelector.py
r-owen: I'm going to ask you to file a separate ticket because I don't understand what part of it would be of interest to other star selectors.

PR #24, opened 2016-05-04
DM-5323

r-owen:


python/lsst/meas/algorithms/utils.py
```diff
@@ -171,11 +171,14 @@ def showPsfCandidates(exposure, psfCellSet, psf=None, frame=None, normalize=True
                         stdev = numpy.sqrt(afwMath.makeStatistics(im.getVariance(), afwMath.MEAN).getValue())
                         afwMath.randomGaussianImage(bim.getImage(), afwMath.Random())
                         bim *= stdev
-                        var = bim.getVariance(); var.set(stdev**2); del var
+                        var = bim.getVariance()
+                        var.set(stdev**2)
+                        del var
```
PaulPrice: How about just:

    bim.getVariance().set(stdev**2)

python/lsst/meas/algorithms/utils.py
```diff
@@ -737,14 +744,17 @@ def showPsfMosaic(exposure, psf=None, nx=7, ny=None,
         i = 0
         with ds9.Buffering():
             for cen, shape in zip(centers, shapes):
-                bbox = mos.getBBox(i); i += 1
+                bbox = mos.getBBox(i)
+                i += 1
```
PaulPrice: Use `enumerate`.

python/lsst/meas/algorithms/utils.py
```diff
@@ -737,14 +744,17 @@ def showPsfMosaic(exposure, psf=None, nx=7, ny=None,
...
                 xc, yc = cen[0] + bbox.getMinX(),  cen[1] + bbox.getMinY()
                 if showCenter:
                     ds9.dot("+", xc, yc,  ctype=ds9.BLUE, frame=frame)
 
                 if showEllipticity:
                     ixx, ixy, iyy = shape
-                    ixx *= scale; ixy *= scale; iyy *= scale
```
PaulPrice: @RobertLuptonTheGood likes putting these kinds of things on one line, since they go together.  I think you need to fight it out with him.

tests/cr.py
```diff
@@ -93,7 +90,8 @@ def setUp(self):
                 else:
                     self.XY0 = afwGeom.PointI(280, 2750)
                     self.nCR = 2
-                self.mi = self.mi.Factory(self.mi, afwGeom.BoxI(self.XY0, afwGeom.ExtentI(256, 256), afwImage.LOCAL))
+                self.mi = self.mi.Factory(self.mi, afwGeom.BoxI(self.XY0, afwGeom.ExtentI(256, 256),
+                                        afwImage.LOCAL))
```
PaulPrice: Line up with the open-paren.

tests/measure.py
```diff
@@ -99,7 +100,9 @@ def __eq__(self, other):
 
     def setUp(self):
         ms = afwImage.MaskedImageF(afwGeom.ExtentI(31, 27))
-        var = ms.getVariance(); var.set(1); del var
+        var = ms.getVariance()
+        var.set(1)
+        del var
```
PaulPrice: `ms.getVariance().set(1)`

tests/measure.py
```diff
@@ -190,11 +194,13 @@ def setUp(self):
 
         if False:                       # use full image, trimmed to data section
             self.XY0 = afwGeom.PointI(32, 2)
-            self.mi = self.mi.Factory(self.mi, afwGeom.BoxI(self.XY0, afwGeom.PointI(2079, 4609)), afwImage.LOCAL)
+            self.mi = self.mi.Factory(self.mi, afwGeom.BoxI(self.XY0, afwGeom.PointI(2079, 4609)),
+                                    afwImage.LOCAL)
```
PaulPrice: Line up with open-paren.  Here and below.

tests/measure.py
```diff
@@ -227,12 +233,14 @@ def testDetection(self):
...
-        bctrl.setNySample(int(self.mi.getHeight()/bgGridSize) + 1);
+        bctrl = afwMath.BackgroundControl(afwMath.Interpolate.NATURAL_SPLINE)
+        bctrl.setNxSample(int(self.mi.getWidth()/bgGridSize) + 1)
+        bctrl.setNySample(int(self.mi.getHeight()/bgGridSize) + 1)
         backobj = afwMath.makeBackground(self.mi.getImage(), bctrl)
 
-        img = self.mi.getImage(); img -= backobj.getImageF(); del img
+        img = self.mi.getImage()
+        img -= backobj.getImageF()
+        del img
```
PaulPrice: These are annoying.  In the past I've written it as:

    self.mi.getImage().__isub__(backobj.getImageF())

Not particularly beautiful, but neither is three lines.

tests/psfIO.py
```diff
@@ -125,19 +122,22 @@ def setUp(self):
...
+            del im
+            del smi
 
         psf = roundTripPsf(4, algorithms.DoubleGaussianPsf(self.ksize, self.ksize,
-                                                             self.FWHM/(2*sqrt(2*log(2))), 1, 0.1))
+                                                             self.FWHM/(2*math.sqrt(2*math.log(2))), 1, 0.1))
 
-        self.cellSet = afwMath.SpatialCellSet(afwGeom.BoxI(afwGeom.PointI(0, 0), afwGeom.ExtentI(width, height)), 100)
+        self.cellSet = afwMath.SpatialCellSet(afwGeom.BoxI(afwGeom.PointI(0, 0),
+                                            afwGeom.ExtentI(width, height)), 100)
```
PaulPrice: Align with open-paren.

tests/psfIO.py
```diff
@@ -302,26 +306,30 @@ def testGetPcaKernel(self):
...
-            
+
     def testCandidateList(self):
         if False and display:
             ds9.mtv(self.mi)
 
             for cell in self.cellSet.getCellList():
                 x0, y0, x1, y1 = \
-                    cell.getBBox().getX0(), cell.getBBox().getY0(), cell.getBBox().getX1(), cell.getBBox().getY1()
+                    cell.getBBox().getX0(), cell.getBBox().getY0(), \
```
PaulPrice: Prefer parens to backslashes.

tests/measure.py
RobertLuptonTheGood: Does
 
    self.mi.getImage()[:] -= backbone.getImageF()

not work?  I don't think it'll add any overhead

tests/testDoubleGaussianPsf.py
```diff
@@ -40,7 +38,8 @@ class DoubleGaussianPsfTestCase(unittest.TestCase):
     def setUp(self):
         FWHM = 5
         self.ksize = 25                      # size of desired kernel
-        self.psf = measAlg.DoubleGaussianPsf(self.ksize, self.ksize, FWHM/(2*sqrt(2*log(2))), 1, 0.1)
+        self.psf = measAlg.DoubleGaussianPsf(self.ksize, self.ksize,
+                                            FWHM/(2*math.sqrt(2*math.log(2))), 1, 0.1)
```
PaulPrice: Alignment.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+
+__all__ = ("SubtractBackgroundConfig", "SubtractBackgroundTask")
+
+class SubtractBackgroundConfig(pexConfig.Config):
+    """!Config for SubtractBackgroundTask and lsst::afw::math::makeBackground
+
+    @warning This config is automatically converted to an lsst::afw::math::BackgroundControl
+    for lsst::afw::math::makeBackground, so do not change the field names or types.
```
PaulPrice: This seems fragile, especially since `BackgroundControl` is in a separate package.  How about moving this `SubtractBackgroundConfig` into afw as `BackgroundConfig`?  If necessary, you could subclass it here.  That would keep together things that go together.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+        dtype=str, default="REDUCE_INTERP_ORDER",
+        allowed={
+            "THROW_EXCEPTION": "throw an exception if there are too few points",
+            "REDUCE_INTERP_ORDER": "use an interpolation style with a lower order.",
+            "INCREASE_NXNYSAMPLE": "Increase the number of samples used to make the interpolation grid.",
+            },
+    )
+    binSize = pexConfig.RangeField(
+        doc="how large a region of the sky should be used for each background point",
+        dtype=int, default=256, min=10,
```
PaulPrice: The `min` seems somewhat arbitrary.  I suggest it could be as small as `1`.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+            },
+    )
+    ignoredPixelMask = pexConfig.ListField(
+        doc="Names of mask planes to ignore while estimating the background",
+        dtype=str, default = ["BAD", "EDGE", "DETECTED", "DETECTED_NEGATIVE", "NO_DATA",],
+        itemCheck = lambda x: x in afwImage.MaskU().getMaskPlaneDict().keys(),
+    )
+    isNanSafe = pexConfig.Field(
+        doc="Ignore NaNs when estimating the background",
+        dtype=bool, default=False,
```
PaulPrice: I suggest that `True` is a better default.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+    weighting = pexConfig.Field(
+        doc="Use inverse variance weighting in calculation (valid only with useApprox=True)",
+        dtype=bool, default=True,
+    )
+
+    def validate(self):
+        pexConfig.Config.validate(self)
+        # Allow None to be used as an equivalent for "NONE", even though C++ expects the latter.
+        if self.algorithm is None:
+            self.algorithm = "NONE"
```
PaulPrice: Do we necessarily `validate` before freezing?  One doesn't normally think of validation as changing things.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+        @param[in] statsKeys  key names used to store the mean and variance of the background
+            in the exposure's metadata (a pair of strings); if None then use ("BGMEAN", "BGVAR");
+            ignored if stats is false
+
+        @return an lsst.pipe.base.Struct containing:
+        - background  full background model (initial model with changes), an lsst.afw.math.BackgroundList
+
+        @throw RuntimeError if fitBackground returns a background of None
+        """
+        displayBackground = lsstDebug.Info(__name__).displayBackground
```
PaulPrice: A `run` method should be as compact as possible, with any real work delegated to other methods.  This makes it simpler to subclass and override.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+
+        if background is None:
+            background = afwMath.BackgroundList()
+
+        maskedImage = exposure.getMaskedImage()
+
+        fitBg = self.fitBackground(maskedImage)
+        if not fitBg:
```
PaulPrice: Why might `fitBackground` return something that doesn't evaluate as `True`?  You should let that method do any `raise` that needs to be done.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+        background.append(fitBg)
+        del fitBgImg
+
+        if displayBackground > 1 or stats:
+            netBgImg = background.getImage()
+
+        if displayBackground > 1:
+            ds9.mtv(netBgImg, title="background", frame=3)
+
+        # Record statistics of the background in the bgsub exposure metadata.
```
PaulPrice: Is this a feature you've added?  I don't think it's really going to be useful, at least as implemented.  The problem is that we do multiple background subtractions in the course of running processCcd or similar, and so it's not clear what values go with what operation.  I suggest dropping this functionality.  At the very least, it needs to go in its own method.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+            s = afwMath.makeStatistics(netBgImg, afwMath.MEAN | afwMath.VARIANCE)
+            bgmean = s.getValue(afwMath.MEAN)
+            bgvar = s.getValue(afwMath.VARIANCE)
+            meta.addDouble(mnkey, bgmean)
+            meta.addDouble(varkey, bgvar)
+
+        if displayBackground:
+            ds9.mtv(exposure, title="subtracted")
+
+        return pipeBase.Struct(
```
PaulPrice: Why return a `Struct` with a single element?  I know that it adds some future-proofing against changes where you might decide to return more information.  But I'm not sure you want that future-proofing, but instead you want things to fail because it's changed.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+            background = background,
+        )
+
+    def fitBackground(self, maskedImage, nx=0, ny=0, algorithm=None):
+        """!Estimate the background of a masked image
+
+        This is a thin layer on lsst.afw.math.makeBackground
```
PaulPrice: It's not that thin!

python/lsst/meas/algorithms/subtractBackground.py
r-owen: No. I think it is a wart. I was preserving existing functionality.

python/lsst/meas/algorithms/subtractBackground.py
r-owen: Run methods always return a Struct and it is exactly for being able to return additional items (such as statistics) in the future.

python/lsst/meas/algorithms/subtractBackground.py
r-owen: Good point. I just preserved the existing comment, but will change it.

python/lsst/meas/algorithms/subtractBackground.py
PaulPrice: I suggest dropping it in an additional commit.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+        #       afw/Approximate.cc.
+        #       See ticket DM-2920: "Clean up code in afw for Approximate background
+        #       estimation" (which includes a note to remove the following and the
+        #       similar checks in pipe_tasks/matchBackgrounds.py once implemented)
+        #
+        # Check that config setting of approxOrder/binSize make sense
+        # (i.e. ngrid (= shortDimension/binSize) > approxOrderX) and perform
+        # appropriate undersampleStlye behavior.
+        if self.config.useApprox:
+            if self.config.approxOrderY not in (self.config.approxOrderX,-1):
```
PaulPrice: Space after comma.

python/lsst/meas/algorithms/subtractBackground.py
r-owen: I agree, but was trying to preserve the existing behavior of getBackground in fitBackground. I'll look at it harder to see if we can get away with raising in fitBackground instead. Even if fitBackground needs to return None, this test should be `if fitBg is None` instead of `if not fitBg`

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -0,0 +1,260 @@
...
+                raise ValueError("Error: approxOrderY not in (approxOrderX, -1)")
+            order = self.config.approxOrderX
+            minNumberGridPoints = self.config.approxOrderX + 1
+            if min(nx,ny) <= self.config.approxOrderX:
+                self.log.warn("Too few points in grid to constrain fit: min(nx, ny) < approxOrder) "+
```
PaulPrice: Space before `+`.

examples/subtractBackgroundExample.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
```
PaulPrice: 2016

python/lsst/meas/algorithms/subtractBackground.py
r-owen: I think we always validate before freezing, but it is still disturbing to be changing things in validate. I was preserving existing behavior, but I don't like it.

examples/subtractBackgroundExample.py
```diff
@@ -0,0 +1,64 @@
...
+"""
+from __future__ import absolute_import, division, print_function
+import os.path
+
+import lsst.utils
+import lsst.afw.image as afwImage
+import lsst.afw.math as afwMath
+from lsst.meas.algorithms import SubtractBackgroundTask
+
+def loadData(psfSigma=1.5):
```
PaulPrice: Drop `psfSigma`.

python/lsst/meas/algorithms/subtractBackground.py
r-owen: I think that is out of scope for this ticket, which consists of refactoring, not changing behavior.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -105,7 +105,99 @@ def validate(self):
...
+
+    @section meas_algorithms_subtractBackground_Initialize  Task initialisation
+
+    @copydoc \_\_init\_\_
+
+    @section meas_algorithms_subtractBackground_IO  Invoking the Task
+
+    Call `run` to fit the background and subtract it.
+
+    Call `fitBackground` if to fit the background without subtracting it
```
PaulPrice: Remove `if`.

python/lsst/meas/algorithms/subtractBackground.py
r-owen: Good suggestion. It would be even safer to make config that matches BackgroundControl an element of this config. That buries things a level down, which is annoying, but is probably necessary (though it will require more refactoring).

python/lsst/meas/algorithms/subtractBackground.py
r-owen: To say more about this: it seems very likely to me that most of our exposures will end up with metadata that contains statistics for the initially estimated background, but those values will not match the final background that was subtracted. Ick.

tests/measure.py
r-owen: Sure. It's out of scope, but too trivial to justify another ticket so I'll implement your suggested modernizations in a separate commit.

PR #26, opened 2016-05-05
Use and enforce Astropy compliant units

pschella:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -306,15 +306,15 @@ def makeMinimalSchema(filterNameList, addFluxSigma=False,
...
+                units = "count",
             )
         if addFluxSigma:
             for filterName in filterNameList:
                 schema.addField(
                     field = "%s_fluxSigma" % (filterName,),
                     type = numpy.float64,
                     doc = "flux uncertainty in filter %s" % (filterName,),
-                    units = "?",
+                    units = "count",
```
TallJimbo: I don't think the units in reference catalogs are actually counts.  We had a long debate on what they should be, but I don't remember the outcome; maybe @r-owen does?

timj: All your options are listed in http://www.ivoa.net/documents/VOUnits/20140523/VOUnits-REC-1.0-20140523.pdf

TallJimbo: Well we're using *some* unit already, and I'm not proposing we change it right now if it's not compliant (but it's much better to leave it blank than set it incorrectly).

r-owen: The units are Janskys, as per http://lsst-web.ncsa.illinois.edu/doxygen/x_masterDoxyDoc/classlsst_1_1meas_1_1algorithms_1_1load_reference_objects_1_1_load_reference_objects_task.html#meas_algorithms_loadReferenceObjects_Schema

PR #27, opened 2016-05-09
Change from boost::math to std

pschella:


TallJimbo:
Aside from `<cmath>` vs. ndarray, include ordering was correct before the last commit (and is incorrect now).

PR #28, opened 2016-05-15
meas_algorithms DM-6078: structural changes to fix SDSS apcorr

TallJimbo:


TallJimbo:
Just noticed an unsurprising test failure on this branch that I'd somehow missed.  Will have a commit to fix it shortly.

python/lsst/meas/algorithms/measureApCorr.py
```diff
@@ -0,0 +1,250 @@
...
+    """!Configuration for MeasureApCorrTask
+    """
+    refFluxName = lsst.pex.config.Field(
+        doc = "Field name prefix for the flux other measurements should be aperture corrected to match",
+        dtype = str,
+        default = "slot_CalibFlux",
+    )
+    starSelector = lsst.pex.config.ConfigurableField(
+        target = UsePreviousStarSelectorTask,
+        doc = "Selector that sets the stars that aperture corrections will e measured from."
```
r-owen: will e -> will be

python/lsst/meas/algorithms/measureApCorr.py
```diff
@@ -0,0 +1,250 @@
...
+            try:
+                self.toCorrect[name] = FluxKeys(name, schema)
+            except KeyError:
+                # if a field in the registry is missing, just ignore it.
+                pass
+        self.makeSubtask("starSelector", schema=schema)
+
+    def run(self, exposure, catalog):
+        """!Measure aperture correction
+
```
r-owen: Please add doxygen @param for `exposure` and `catalog` arguments

tests/psfSelectTest.py
```diff
@@ -189,21 +189,6 @@ def plantSources(x0, y0, nx, ny, sky, nObj, wid, detector, useRandom=False):
 
     return expos, goodAdded, expos0, goodAdded0
 
-#################################################################
-# quick and dirty detection (note: we already subtracted background)
-def detectAndMeasure(exposure, detConfig, measConfig):
```
r-owen: Please use a normal doc string instead of a comment, and preferably document the arguments (what classes are detConfig and measConfig instances of?)

tests/psfSelectTest.py
```diff
@@ -272,16 +265,23 @@ def setUp(self):
...
+        del self.detTask
+        del self.measTask
+        del self.schema
         del self.detector
         del self.flatDetector
         del self.starSelector
         del self.psfDeterminer
 
+    # quick and dirty detection (note: we already subtracted background)
+    def detectAndMeasure(self, exposure):
```
r-owen: Please replace comment with doc string

tests/testMeasureApCorr.py
```diff
@@ -0,0 +1,204 @@
...
+        """ Check that valid sources inside the bounding box that are used have their flags set to True"""
+        inputFilterFlagKey = self.schema.find(self.meas_apCorr_task.config.starSelector.field).key
+        self.meas_apCorr_task.run(catalog=self.sourceCat, exposure=self.exposure)
+        self.assertTrue(self.sourceCat[inputFilterFlagKey].all())
+
+    def testApertureMeasOnes(self):
+        """ Check that sources with aperture fluxes exactly the same as their catalog fluxes
+            returns an aperture correction map of 1s"""
+        apFluxName = self.apname + "_flux"
+        sourceCat = apCorrTestSourceCatalog(schema=self.schema, name=self.name, apname=self.apname,
```
r-owen: Given the frequent need to call this, I'm not convinced making a default catalog is an improvement. Instead how about making apCorrTestSourceCatalog a method so that it can have sensible defaults for most of its arguments (in fact many of its argument could be eliminated). Then calling it would typically require only overriding the one or two arguments that actually matter. I think this would make it much easier to understand what is going on.

tests/testPsfDetermination.py
```diff
@@ -219,7 +219,7 @@ def tearDown(self):
         del self.catalog
 
     @staticmethod
-    def setupDeterminer(exposure, nEigenComponents=3, starSelectorAlg="secondMoment"):
+    def setupDeterminer(exposure, schema, nEigenComponents=3, starSelectorAlg="secondMoment"):
```
r-owen: Please consider making this a normal method so that the schema argument can be eliminated or can default to self.catalog.schema (presumably by defaulting to None and if None then use self.catalog.schema).

python/lsst/meas/algorithms/measureApCorr.py
TallJimbo: Done.

tests/psfSelectTest.py
TallJimbo: Done.

tests/testMeasureApCorr.py
TallJimbo: Good idea.  Fixed.

tests/testPsfDetermination.py
TallJimbo: Done.  I just had to do something very similar to the tests in meas_extensions_psfex, so this was pretty straightforward.

PR #30, opened 2016-05-28
Allow butlers in subclasses

r-owen:
Reference catalog loaders will frequently need butlers.
This allows subclasses to optionally take a butler.


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -171,6 +171,10 @@ class LoadReferenceObjectsTask(pipeBase.Task):
     ConfigClass = LoadReferenceObjectsConfig
     _DefaultName = "LoadReferenceObjects"
 
+    def __init__(self, butler=None, *args, **kwargs):
+        pipeBase.Task.__init__(self, *args, **kwargs)
+        self.butler = butler
```
r-owen: Please add a doc string and describe the butler argument. Perhaps explain briefly why it is present.

PR #34, opened 2016-06-01
DM-5877: Use Afterburners to clean up aperture correction logic

PaulPrice:


PaulPrice:
Please explain in the first commit what has changed in meas_base that necessitates this patch.

PR #35, opened 2016-06-07
Tickets/dm 6077

r-owen:


python/lsst/meas/algorithms/psfDeterminer.py
```diff
@@ -19,25 +19,56 @@
...
+        default = 25,
+    )
+    kernelSizeMax = pexConfig.Field(
+        doc = "Maximum radius of the kernel",
+        dtype = int,
+        default = 45,
+    )
+
+
+class BasePsfDeterminerTask(pipeBase.Task):
```
djreiss: I hate to ask, but could you change the commit message to have the new class name (BasePsfDeterminerTask) instead of the old one? I only ask because it confused me for a bit. Thanks.

r-owen: Great suggestion. I did that.

python/lsst/meas/algorithms/psfDeterminer.py
```diff
@@ -19,25 +19,56 @@
...
-    '''A registry of PSF determiner factories
+import lsst.pipe.base as pipeBase
+import lsst.pex.config as pexConfig
 
-    A PSF determiner factory makes a class with the following API:
+__all__ = ["BasePsfDeterminerConfig", "BasePsfDeterminerTask", "psfDeterminerRegistry"]
 
-    def __init__(self, config, schema=None):
+class BasePsfDeterminerConfig(pexConfig.Config):
+    kernelSize = pexConfig.Field(
```
djreiss: I think I understand why you put these three settings in the base config, but is there some way to document/explain it? For example, I could see a justification for also including "borderWidth" in the base, although I realize that it's only not there because it's not used by PsfexPsfDeterminer. Is that correct?

r-owen: I only put in the few settings that were clearly needed by both existing PSF determiners and looked likely to be needed by all others, since we prefer to avoid fields that are ignored. I was disappointed that the PCA psf determiner has nothing like `badMaskBits`, since that would be an obvious shared config field. I'm not sure if `spatialOrder` or cell size would be universal. We can easily move more fields into this common config later when we are confident they are shared.

r-owen: In any case, I'll add some documentation.

PR #37, opened 2016-06-13
PR for review of DM-4160

pschella:


python/lsst/meas/algorithms/utils.py
```diff
@@ -226,9 +226,8 @@ def showPsfCandidates(exposure, psfCellSet, psf=None, frame=None, normalize=True
 
                 # residuals using spatial model
                 try:
-                    chi2 = algorithmsLib.subtractPsf(psf, im, xc, yc)
-                except:
-                    chi2 = numpy.nan
+                    algorithmsLib.subtractPsf(psf, im, xc, yc)
+                except Exception:
```
pschella: Note that by changing `except:` to `except Exception:` you changed the nature of the exceptions being caught. For instance you are now no longer catching `BaseException` and some system events. Which probably is fine, but you may want to be sure. Probably the best thing here would be to find out exactly what exception `subtractPsf` is expected to throw and catch only those. Now you get no indication if something unexpected goes wrong (but this pre-existed this ticket).

jdswinbank:
Has been merged.

PR #38, opened 2016-06-29
Added initial astrometry source selector, base class and tests.

parejkoj:
This is the first non-PSF source selector, with a new base class for other such things. It was lifted from measOptimisticB for use by jointcal.

### DIFF
--- a/python/lsst/meas/algorithms/astrometrySourceSelector.py
+++ b/python/lsst/meas/algorithms/astrometrySourceSelector.py
@@ -1,0 +1,137 @@
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+
+from lsst.afw import table
+import lsst.pex.config as pexConfig
+from .sourceSelector import BaseSourceSelectorConfig, BaseSourceSelectorTask, sourceSelectorRegistry
+from lsst.pipe.base import Struct
+
+
+class AstrometrySourceSelectorConfig(BaseSourceSelectorConfig):
+    sourceFluxType = pexConfig.Field(
+        doc = "Type of source flux; typically one of Ap or Psf",
+        dtype = str,
+        default = "Ap",
+    )
+    minSnr = pexConfig.Field(
+        dtype = float,
+        doc = "Minimum allowed signal-to-noise ratio for sources used for matching "
+              "(in the flux specified by sourceFluxType); <= 0 for no limit",
+        default = 10,
+    )
+
+
+class AstrometrySourceSelectorTask(BaseSourceSelectorTask):
+    """
+    !Select sources that are useful for astrometry.
+
+    Good astrometry sources have high signal/noise, are non-blended, and
+    did not have certain "bad" flags set during source extraction. They need not
+    be PSF sources, just have reliable centroids.
+    """
+    ConfigClass = AstrometrySourceSelectorConfig
+
+    def __init__(self, *args, **kwargs):
+        BaseSourceSelectorTask.__init__(self, *args, **kwargs)
+
+    def selectSources(self, sourceCat, matches=None):
+        """
+        !Return a catalog of sources: a subset of sourceCat.
+
+        @param[in] sourceCat  catalog of sources that may be sources
+                                (an lsst.afw.table.SourceCatalog)
+
+        @return a pipeBase.Struct containing:
+        - sourceCat  a catalog of sources
+        """
+        self._getSchemaKeys(sourceCat.schema)
+
+        result = table.SourceCatalog(sourceCat.table)
+        for source in sourceCat:
+            if self._isGood(source) and not self._isBad(source):
+                result.append(source)
+        return Struct(sourceCat=result)
+
+    def _getSchemaKeys(self, schema):
+        """Extract and save the necessary keys from schema with asKey."""
+        self.parentKey = schema["parent"].asKey()
+        self.nChildKey = schema["deblend_nChild"].asKey()
+
+        self.edgeKey = schema["base_PixelFlags_flag_edge"].asKey()
+        self.interpolatedCenterKey = schema["base_PixelFlags_flag_interpolatedCenter"].asKey()
+        self.saturatedKey = schema["base_PixelFlags_flag_saturated"].asKey()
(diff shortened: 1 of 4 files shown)


python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -0,0 +1,121 @@
...
+            and not source.get(self.fluxFlagKey) \
+            and self._goodSN(source)
+
+    def _isGood(self, source):
+        """
+        Return True if source is usable for matching and likely has a good centroid.
+
+        The additional tests for a good centroid, beyond isUsable, are:
+        - not interpolated in the center (this includes saturated sources,
+            so we don't test separately for that)
```
isullivan: This comment is confusing, since it looks like you do test separately for saturated sources below.

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -0,0 +1,121 @@
+# See COPYRIGHT file at the top of the source tree.
```
isullivan: Can we do this now? It's nice and short, but I don't see a COPYRIGHT file anywhere in the meas_algorithms repo.

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -0,0 +1,121 @@
...
+
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+
+from lsst.afw import table
```
isullivan: I thought the convention was to import these modules as:

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -0,0 +1,121 @@
...
+        for source in sourceCat:
+            if self._isGood(source) and not self._isBad(source):
+                result.append(source)
+        return Struct(sourceCat=result)
+
+    def _getSchemaKeys(self, schema):
+        """Extract and save the necessary keys from schema with asKey."""
+        self.parentKey = schema["parent"].asKey()
+        self.centroidKey = table.Point2DKey(schema["slot_Centroid"])
+        self.centroidFlagKey = schema["slot_Centroid_flag"].asKey()
```
isullivan: It doesn't look like centroidFlagKey is used anywhere else. Same with fluxField below. Consider removing them.

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -0,0 +1,121 @@
...
+    def _isMultiple(self, source):
+        """Return True if source is likely multiple sources."""
+        if (source.get(self.parentKey) != 0) or (source.get(self.nChildKey) != 0):
+            return True
+        footprint = source.getFootprint()
+        return footprint is not None and len(footprint.getPeaks()) > 1
+
+    def _hasCentroid(self, source):
+        """Return True if the source has a valid centroid"""
+        centroid = source.get(self.centroidKey)
```
isullivan: Use source.getCentroid() here. Then consider removing centroidKey above.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -0,0 +1,79 @@
...
+        # NOTE: example implementation, returning all sources that have no bad flags set.
+        result = table.SourceCatalog(sourceCat.table)
+        for source in sourceCat:
+            if not self._isBad(source):
+                result.append(source)
+        return pipeBase.Struct(sourceCat=result)
+
+    def _isBad(self, source):
+        """Return True if any of config.badFlags are set for this source."""
+        return reduce(lambda x, y: np.logical_or(x, source.get(y)), self.config.badFlags, False)
```
isullivan: It would take a few more lines, but why not write:
for field in self.config.badFlags:
    if source.get(field) is True:
        return True
else:
    return False

tests/testAstrometrySourceSelector.py
```diff
@@ -0,0 +1,121 @@
...
+
+    def testSelectSources_no_SN_cut(self):
+        self.sourceSelector.config.minSnr = 0
+        add_good_source(self.src, 1)
+        self.src['slot_ApFlux_flux'][0] = 0
+        result = self.sourceSelector.selectSources(self.src)
+        self.assertIn(self.src[0]['id'], result.sourceCat['id'])
+
+
+# for MemoryTestCase
```
isullivan: For consistency with most other tests, could you write the final section as:
#-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-

def suite():
    """Returns a suite containing all the test cases in this module."""
    utilsTests.init()

    suites = []
    suites += unittest.makeSuite(TestAstrometrySourceSelector)
    suites += unittest.makeSuite(utilsTests.MemoryTestCase)
    return unittest.TestSuite(suites)

def run(exit = False):
    """Run the utilsTests"""
    utilsTests.run(suite(), exit)

if __name__ == "__main__":
    run(True)

python/lsst/meas/algorithms/sourceSelector.py
PaulPrice: return any(source.get(flag) for flag in self.config.badFlags)

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: I'm taking this from jointcal, where I was told to do it that way (THE FUTURE!). I'll check with SQuaRE.

tests/testAstrometrySourceSelector.py
parejkoj: This is the new test boilerplate, described in https://sqr-012.lsst.io/. TestSuite is dead, long live lsst.utils.tests.init()!

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: Apparently we never settled on a namespace standard in RFC-34: https://jira.lsstcorp.org/browse/RFC-34. I guess I'll go with the LSST "standard" for now.

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: Done.

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: Oh, good catch. Done.

python/lsst/meas/algorithms/sourceSelector.py
parejkoj: A full `for` loop is significantly slower than `reduce()`. Paul's `any()` is good.

I took this from a different starSelector that ran it on the whole sourceCat at once. But that fails if the catalog is non-contiguous. I was going to switch it to that (reduce over the whole catalog is very fast) but my test confirmed that failure, so I'll go with `any()`.

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: Comment was in the original code I stole this from (measOptimisticB). I added the separate saturated check, since I'm using fake data. I think I'd rather leave it in, to catch corner cases (plus, I don't know anything about how interpolatedCenter and saturated interact beyond that one comment).

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: Word from SQuaRE is that we can't quite do the switch yet on existing projects. @jonathansick is planning a full switch for the whole codebase VerySoonNow™.

PR #39, opened 2016-07-04
Added 100x faster vectorized version

parejkoj:
Would be ~500x faster if we didn't have to do the footprint loop.

The if branch between vector and non-vectorized versions (and associated split in code logic) is unpleasant, but we'll have to live with it until we finalize the discussion here: https://community.lsst.org/t/memory-contiguity-and-python-execution-speed/884/12

### DIFF
--- a/python/lsst/meas/algorithms/astrometrySourceSelector.py
+++ b/python/lsst/meas/algorithms/astrometrySourceSelector.py
@@ -60,6 +60,10 @@ class AstrometrySourceSelectorTask(BaseSourceSelectorTask):
         """
         !Return a catalog of sources: a subset of sourceCat.
 
+        If sourceCat is cotiguous in memory, will use vectorized tests for ~100x
+        execution speed advantage over non-contiguous catalogs. This would be
+        even faster if we didn't have to check footprints for multiple peaks.
+
         @param[in] sourceCat  catalog of sources that may be sources
                                 (an lsst.afw.table.SourceCatalog)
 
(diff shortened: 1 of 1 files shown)


python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -68,16 +72,24 @@ def selectSources(self, sourceCat, matches=None):
...
+                if self._isGood(source) and not self._isBad(source):
+                    result.append(source)
         return Struct(sourceCat=result)
 
     def _getSchemaKeys(self, schema):
         """Extract and save the necessary keys from schema with asKey."""
         self.parentKey = schema["parent"].asKey()
         self.nChildKey = schema["deblend_nChild"].asKey()
+        self.centroidXKey = schema["slot_Centroid_x"].asKey()
+        self.centroidYKey = schema["slot_Centroid_y"].asKey()
```
isullivan: I thought the field was "slot_Centroid" and it returned a tuple of the (y, x) centroid location. Does this also work?

parejkoj: That works for individual sources, but I need the x/y centroids for the vectorized np.isfinite() test: there isn't an equivalent SourceCatalog thing to get a vector of tuples.

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -68,16 +72,24 @@ def selectSources(self, sourceCat, matches=None):
...
-                result.append(source)
+        if sourceCat.isContiguous():
+            bad = reduce(lambda x, y: np.logical_or(x, sourceCat.get(y)), self.config.badFlags, False)
+            good = self._isGood_vector(sourceCat)
+            result = sourceCat[good & ~bad]
+        else:
+            result = table.SourceCatalog(sourceCat.table)
+            for i, source in enumerate(sourceCat):
+                if self._isGood(source) and not self._isBad(source):
+                    result.append(source)
```
isullivan: I don't like the code duplication of having vectorized and non-vectorized versions of every function, but I suppose that can't be helped.

python/lsst/meas/algorithms/astrometrySourceSelector.py
TallJimbo: Small correction: when used as a single key `slot_Centroid` will return `afw.geom.Point2D`, which is indexable as `(x,y)`, not `(y,x)`.  The usage here is completely valid too, of course.

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: I agree, but until we have a generic solution, I'd rather not break non-contiguous functionality.

PR #44, opened 2016-08-17
tickets/DM-7248

AstroVPK:


tests/testExecutables.py
```diff
@@ -0,0 +1,12 @@
+import unittest
+import lsst.utils.tests
+
+
+class UtilsBinaryTester(lsst.utils.tests.ExecutablesTestCase):
+    pass
+
+EXECUTABLES = ['testPsfAttributes', 'testSizeMagnitudeStarSelector', 'testImagePsf', 'testPsfCaching', 'testWarpedPsf']
```
r-owen: Is this necessary? I thought it found all binaries by default.

AstroVPK: I didn't know it did that! It is great if it does that. I can check and revert to the default if that's true...

tests/testAstrometrySourceSelector.py
```diff
@@ -17,15 +18,15 @@
...
+# see <http://www.lsstcorp.org/LegalNotices/>.
 #
 from __future__ import division, absolute_import, print_function
 
 import unittest
 import numpy as np
 
 import lsst.utils
-from lsst.meas.base.tests import TestDataset
+from lsst.meas.base.tests import TestDataset as Dataset
```
r-owen: Renaming symbols on import is confusing. I understand why you can't import the bare name using its original name, but I'd suggest using the package name as a prefix instead.

isullivan: so just use lsst.meas.base.tests.TestDataset ?

.gitignore
```diff
@@ -24,8 +24,8 @@ python/lsst/meas/algorithms/version.py
 tests/.tests
 tests/data/psf*.*
 tests/photometry
-tests/psfAttributes
```
laurenam: Looks like tests/photometry can be removed.

tests/testAstrometrySourceSelector.py
```diff
@@ -17,17 +18,17 @@
 #
 # You should have received a copy of the LSST License Statement and
 # the GNU General Public License along with this program.  If not,
-# see <https://www.lsstcorp.org/LegalNotices/>.
+# see <http://www.lsstcorp.org/LegalNotices/>.
 #
```
laurenam: Looks like you are not using he current copyright template: https://github.com/lsst/templates/blob/master/CopyrightHeader.py

tests/testCoaddApCorrMap.py
```diff
@@ -20,26 +20,19 @@
...
-import lsst.utils.tests as utilsTests
 import lsst.afw.geom as afwGeom
 import lsst.afw.math as afwMath
 import lsst.afw.table as afwTable
 import lsst.afw.image as afwImage
 import lsst.afw.coord as afwCoord
-import lsst.meas.algorithms as measAlg
 from lsst.afw.geom.polygon import Polygon
+import lsst.meas.algorithms as measAlg
+import lsst.utils
```
laurenam: Should there also be a .tests appended here?

tests/testCoaddApCorrMap.py
```diff
@@ -112,7 +105,7 @@ def test(self):
         # Only the second record will be valid for this point
         self.assertApCorrMapValid(apCorrMap, pointListValid)
 
-        filename = "tests/coaddApCorrMap.fits"
+        filename = os.path.join(lsst.utils.getPackageDir("meas_algorithms"), "tests", "coaddApCorrMap.fits")
         exposure = afwImage.ExposureF(1, 1)
```
laurenam: Use __file__ as per @timj's instructions.

tests/testCoaddApCorrMap.py
```diff
@@ -125,34 +118,29 @@ def assertApCorrMap(self, apCorrMap, pointList):
         for i, point in enumerate(pointList):
             weights = [i+1, i+2]
             values = [i+1, i+2]
```
laurenam: Here and throughout there are many cases where spaces need to be added around the + and - operators, or removed around the * and / operators.

tests/testCoaddBoundedField.py
```diff
@@ -20,22 +20,23 @@
...
 import lsst.afw.geom
 import lsst.afw.image
 import lsst.afw.coord
-import lsst.meas.algorithms
 from lsst.afw.geom.polygon import Polygon
+import lsst.meas.algorithms
+import lsst.pex.exceptions
+import lsst.utils.tests
+
+np.random.seed(50)
```
laurenam: Should this be moved into setUp()?

tests/testCoaddApCorrMap.py
isullivan: I think this was because of the lsst.utils.getPackageDir... call later on, but if we use __file__ instead then this could be lsst.utils.tests

tests/testHtmIndex.py
```diff
@@ -30,21 +32,20 @@
 
 import numpy as np
 
-import lsst.utils
 import lsst.afw.table as afwTable
 import lsst.afw.geom as afwGeom
 import lsst.afw.coord as afwCoord
 import lsst.daf.persistence as dafPersist
 from lsst.meas.algorithms import IngestIndexedReferenceTask, LoadIndexedReferenceObjectsTask, \
     LoadIndexedReferenceObjectsConfig, getRefFluxField
```
laurenam: Use braces for line wrapping?
https://developer.lsst.io/coding/python_style_guide.html?highlight=print%20statements#python-s-implied-continuation-inside-parens-brackets-and-braces-should-be-used-for-wrapped-lines

tests/testExecutables.py
timj: It will auto detect. If you give a list then it will report if the test executable can't be found (maybe it wasn't built). If you don't give a list it will only test what it can find.

PR #45, opened 2016-08-25
DM-7289: Port to Python 3

timj:


python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -242,7 +243,10 @@ def _set_extra(self, record, row, key_map):
         @param[in] key_map  Map of catalog keys to use in filling the record
         """
         for extra_col in self.config.extra_col_names:
-            record.set(key_map[extra_col], row[extra_col])
+            value = row[extra_col]
+            if isinstance(value, np.str_):
```
timj: I think this check needs some justification in a comment.

python/lsst/meas/algorithms/pcaPsfDeterminer.py
```diff
@@ -234,11 +239,11 @@ def determinePsf(self, exposure, psfCandidateList, metadata=None, flagKey=None):
...
-            self.log.log(-1, \
-                "WARNING: NOT scaling kernelSize by stellar quadrupole moment " +
-                "because config.kernelSize=%s >= 15; using config.kernelSize as as the width, instead" \
-                % (self.config.kernelSize,)
-            )
+            self.log.log(-1,
+                         "WARNING: NOT scaling kernelSize by stellar quadrupole moment " +
+                         "because config.kernelSize=%s >= 15; using config.kernelSize as as the width, instead"
+                         % (self.config.kernelSize,)
+                         )
```
timj: I thought we wanted the closing paren to be aligned with the block not the open paren.

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -114,5 +114,15 @@ def run(self, filename):
         arr = np.genfromtxt(filename, dtype=None, skip_header=self.config.header_lines,
                             delimiter=self.config.delimiter,
                             names=names)
+        # This is to explicitly convert the bytes type into unicode for any column that is read in as bytes
+        # string
```
timj: pain and suffering by the looks of it.

tests/testPsfDetermination.py
```diff
@@ -466,7 +468,7 @@ def testRejectBlends(self):
 
         with self.assertRaises(RuntimeError) as cm:
             psfDeterminer.determinePsf(self.exposure, candidates, metadata)
-        self.assertEqual(cm.exception.message, "All PSF candidates removed as blends")
+        self.assertEqual(cm.exception.args[0], "All PSF candidates removed as blends")
```
timj: `str(cm.exception)` might be better. `.message` is now what happens when you stringify the exception.

tests/testReadFitsCatalog.py
```diff
@@ -22,7 +22,7 @@
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
 from __future__ import absolute_import, division, print_function
-from itertools import izip
+
```
timj: replace with `from builtins import zip`

tests/testGaussianPsfFactory.py
```diff
@@ -50,14 +50,14 @@ def testApply(self):
                             naiveSize += 1
                     else:
                         naiveSize = fixedSize
-                    for minSize in (None, max(1, naiveSize-1), naiveSize, naiveSize + 2):
+                    for minSize in (max(1, naiveSize-1), naiveSize, naiveSize + 2):
```
timj: Why were these changes done to remove the `None`?

AstroVPK: It looks like the None iterations (i.e. either of minsize or maxsize are None) don't do anything. Why are they there in the first place?

timj: That's what is worrying me. Why did the code have it originally and how sure are we it's all irrelevant.

tests/testGaussianPsfFactory.py
```diff
@@ -50,14 +50,14 @@ def testApply(self):
...
                         if minSize <= 0:
                             continue
                         factory.minSize = minSize
-                        for maxSize in (None, naiveSize-1, naiveSize):
+                        for maxSize in (naiveSize-1, naiveSize):
                             if maxSize <= 0:
                                 continue
-                            if None not in (minSize, maxSize) and maxSize < minSize:
```
timj: Which seems to be treating `None` as saying "we do not have a minimum yet"

AstroVPK: That seems to be the case, but since the remainder of the loop is skipped over with the continue, it looks like it's safe to remove the None iteration. Apparently, the code is inherited and the origin of the None iteration is lost.

tests/testGaussianPsfFactory.py
natelust: I have traced back the history of this bit of code as far as I could in git, and it lead me to Russell, who indicated that he was actually just editing it from somewhere else when it got put into the oldest file I found it in. I think it was just some poor coding, or at one point it had a use, that was lost though subsequent edits over the years. the None conditions never lead anywhere other than a continue statement, which continues the loop to the next (non None) value, so they are not needed to test this bit of code.

tests/testGaussianPsfFactory.py
natelust: see my comment above

tests/testGaussianPsfFactory.py
timj: I agree.

PR #46, opened 2016-08-29
DM-7294: Remove esutils hack

timj:
Updating to the latest version of esutils allows us to remove this
old work around


jdswinbank:
This work was done on DM-8232.

PR #51, opened 2016-10-03
make aperture correction failures optionally non-fatal

natelust:
Some measurement algorithms may always fail to produce a result on
certain data (e.g., seeing is too good or too bad to make the desired
measurement). The failure to measure an aperture correction for such
a measurement algorithm should not cause an exception. We now allow
certain pre-specified measurement algorithms (through the new config
parameter 'allowFailure') to fail and produce a warning rather than
an exception.

### DIFF
--- a/python/lsst/meas/algorithms/measureApCorr.py
+++ b/python/lsst/meas/algorithms/measureApCorr.py
@@ -100,6 +100,11 @@ class MeasureApCorrConfig(lsst.pex.config.Config):
         dtype=float,
         default=3.0,
     )
+    allowFailure = lsst.pex.config.ListField(
+        doc="Allow these measurement algorithms to fail without an exception",
+        dtype=str,
+        default=[],
+    )
 
     def validate(self):
         lsst.pex.config.Config.validate(self)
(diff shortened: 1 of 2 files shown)


python/lsst/meas/algorithms/measureApCorr.py
```diff
@@ -235,12 +240,14 @@ def run(self, exposure, catalog):
...
-                apCorrMap[fluxSigmaName] = ChebyshevBoundedField(bbox, numpy.zeros((1, 1), dtype=float))
-                continue
+                if name in self.config.allowFailure:
+                    self.log.warn("Unable to measure aperture correction for '%s': "
+                                  "only %d sources, but require at least %d." %
+                                  (name, len(subset2), self.config.minDegreesOfFreedom+1))
+                    continue
+                raise RuntimeError("Unable to measure aperture correction for required algorithm '%s': "
+                                   "only %d sources, but require at least %d." %
+                                   (name, len(subset2), self.config.minDegreesOfFreedom+1))
```
natelust: Why is it ok to drop the apCorrMap ChebyshevBoundingField stuff? I understand that the new stuff if failure is allowed may not need it, but shouldn't the behavior stay the same for the other cases?

PaulPrice: It was dead code --- it never got hit because the exception is raised first.
Moving forward, I think it's better to have no aperture correction than a fake/wrong one because in the latter case you can have a false sense of security.

PR #52, opened 2016-11-01
meas_algorithms/DM-8147: Add backwards compatibility to CoaddPsf input.

TallJimbo:
This adds support for reading very old Exposures with CoaddPsfs written before 7/10/2013.  It will not in general preserve the PSF's "average position" exactly, which is used when no position is provided to the PSF, because the information needed to reconstruct that has already been lost.

### DIFF
--- a/src/CoaddPsf.cc
+++ b/src/CoaddPsf.cc
@@ -354,8 +354,14 @@ public:
 
     virtual PTR(tbl::io::Persistable)
     read(InputArchive const & archive, CatalogVector const & catalogs) const {
-        CoaddPsfPersistenceHelper const & keys1 = CoaddPsfPersistenceHelper::get();
+        if (catalogs.size() == 1u) {
+            // Old CoaddPsfs were saved in only one catalog, because we didn't
+            // save the warping parameters and average position, and we could
+            // save the coadd Wcs in a special final record.
+            return readV0(archive, catalogs);
+        }
         LSST_ARCHIVE_ASSERT(catalogs.size() == 2u);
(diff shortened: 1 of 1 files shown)


src/CoaddPsf.cc
```diff
@@ -354,8 +354,10 @@ class CoaddPsf::Factory : public tbl::io::PersistableFactory {
 
     virtual PTR(tbl::io::Persistable)
     read(InputArchive const & archive, CatalogVector const & catalogs) const {
+        if (catalogs.size() == 1u) {
+            return readV0(archive, catalogs);
+        }
         CoaddPsfPersistenceHelper const & keys1 = CoaddPsfPersistenceHelper::get();
-        LSST_ARCHIVE_ASSERT(catalogs.size() == 2u);
```
ktlim: Can't you keep this now that you have the check above?

TallJimbo: Good point.  Will do.

src/CoaddPsf.cc
```diff
@@ -354,8 +354,11 @@ class CoaddPsf::Factory : public tbl::io::PersistableFactory {
 
     virtual PTR(tbl::io::Persistable)
     read(InputArchive const & archive, CatalogVector const & catalogs) const {
-        CoaddPsfPersistenceHelper const & keys1 = CoaddPsfPersistenceHelper::get();
+        if (catalogs.size() == 1u) {
```
r-owen: Please explain this test (why does catalogs.size() tell us the catalog version) in a code comment.

PR #53, opened 2016-11-22
DM-8088: When making PSF-matched coadds, warp first then PSF-match

jmeyers314:


jmeyers314:
Looks good @yalsayyad .  No comments on this one.

PR #54, opened 2016-12-08
tickets/DM-8175

yalsayyad:
This ticket contains computeBBox implementations for WarpedPsf, CoaddPsf, and KernelPsf (trivial).

This ticket also includes commit to reduce some of the code duplication in testCoaddPsf,  and a change to make consistent the x/y padding to the PSF image before warping it. The original author of that line (81 in WarpedPsf.cc)  agrees with the change.

### DIFF
--- a/src/CoaddPsf.cc
+++ b/src/CoaddPsf.cc
@@ -234,10 +234,10 @@ afw::geom::Box2I CoaddPsf::doComputeBBox(
     }
 
     afw::geom::Box2I ret;
-    for (afw::table::ExposureCatalog::const_iterator i = subcat.begin(); i != subcat.end(); ++i) {
+    for (auto const & exposureRecord : subcat) {
         PTR(afw::geom::XYTransform) xytransform(
-            new afw::image::XYTransformFromWcsPair(_coaddWcs, i->getWcs()));
-        WarpedPsf warpedPsf = WarpedPsf(i->getPsf(), xytransform, _warpingControl);
+            new afw::image::XYTransformFromWcsPair(_coaddWcs, exposureRecord.getWcs()));
+        WarpedPsf warpedPsf = WarpedPsf(exposureRecord.getPsf(), xytransform, _warpingControl);
         afw::geom::Box2I componentBBox = warpedPsf.computeBBox(ccdXY, color);
         ret.include(componentBBox);
     }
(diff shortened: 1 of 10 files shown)


src/WarpedPsf.cc
```diff
@@ -91,36 +93,56 @@ PTR(afw::detection::Psf::Image) warpAffine(
...
+    afw::math::warpImage(*ret, *im_padded, xyTransform, wc, 0.0);
+    return ret;
+}
+
+
+afw::geom::Box2I computeBBoxFromTransform
+    (afw::geom::Box2I const bbox,
+     afw::geom::AffineTransform const &t
+) {
+    static const double maxTransformCoeff = 200.0;
```
jmeyers314: Since you're checking against `maxTransformCoeff` here, I think you can probably delete the check in `warpAffine`.  Should probably bring the comment block down from there to here, though.

src/CoaddPsf.cc
```diff
@@ -221,6 +221,30 @@ void addToImage(
...
+    afw::table::ExposureCatalog subcat = _catalog.subsetContaining(ccdXY, *_coaddWcs, true);
+    if (subcat.empty()) {
+        throw LSST_EXCEPT(
+            pex::exceptions::InvalidParameterError,
+            (boost::format("Cannot compute BBox at point %s; no input images at that point.")
+             % ccdXY).str());
+    }
+
+    afw::geom::Box2I overallBBox;
+    for (afw::table::ExposureCatalog::const_iterator i = subcat.begin(); i != subcat.end(); ++i) {
```
TallJimbo: I think you'd just want `for (auto i : subcat) {`.

yalsayyad: without the &&, it tries to copy the elements into i:

```
src/CoaddPsf.cc:269:15: error: call to implicitly-deleted copy constructor of 'lsst::afw::table::ExposureRecord'
    for (auto exposureRecord : subcat) {
```

I'll change all the loops over the catalog to  c++11 style, in a separate commit.

TallJimbo: Oh, I'm an idiot.  Of course; I was thinking that `i` would be an iterator type, not the record type.  However, I think the appropriate invocation is then `auto const & record`; we want the element type to resolve to an lvalue const reference `ExposureRecord const &`, not an rvalue reference `ExposureRecord &&`.

jmeyers314: This is very interesting.  I was trying to guess the best choice range declaration from the examples [here](http://en.cppreference.com/w/cpp/language/range-for).  `auto const &` certainly sounds like a better option though, now that I know it exists!

TallJimbo: Yeah, type deduction for `auto` is pretty weird.  I remember thinking at one point that it all made sense, now that I understood it, but even though I now remember at least most of the rules I no longer remember why it had to be that way.

src/WarpedPsf.cc
```diff
@@ -69,14 +73,12 @@ PTR(afw::detection::Psf::Image) warpAffine(
...
 ) {
-    static const int dst_padding = 0;
-
     afw::geom::AffineXYTransform xyTransform(t);
 
     afw::math::SeparableKernel const& kernel = *wc.getWarpingKernel();
     afw::geom::Point2I const& center = kernel.getCtr();
     int const xPad = std::max(center.getX(), kernel.getWidth() - center.getX());
-    int const yPad = std::min(center.getY(), kernel.getHeight() - center.getY());
+    int const yPad = std::max(center.getY(), kernel.getHeight() - center.getY());
```
yalsayyad: I've written a unit test  for this and will move the discussion to the Jira ticket, because I want to attach pictures: https://jira.lsstcorp.org/browse/DM-8175

PR #55, opened 2016-12-16
Cleanup matplotlib imports Tickets/dm 8676

parejkoj:
Pulled matplotlib imports into the functions that use them and added logging. Now you can import meas_algorithms and still set the backend yourself!

### DIFF
--- a/python/lsst/meas/algorithms/utils.py
+++ b/python/lsst/meas/algorithms/utils.py
@@ -46,10 +46,6 @@ from lsst.afw.image.utils import CalibNoThrow
 
 keptPlots = False                       # Have we arranged to keep spatial plots open?
 
-#
-# This should be provided by the mapper.  The details are camera-specific and
-#
-
 
 def splitId(oid, asDict=True):
 


python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -430,21 +428,35 @@ def selectStars(self, exposure, sourceCat, matches=None):
         centers, clusterId = _kcenters(width, nCluster=4, useMedian=True,
                                        widthStdAllowed=self.config.widthStdAllowed)
 
-        if display and plotMagSize and pyplot:
-            fig = plot(mag, width, centers, clusterId,
-                       magType=self.config.sourceFluxField.split(".")[-1].title(),
-                       marker="+", markersize=3, markeredgewidth=None, ltype=':', clear=True)
-        else:
-            fig = None
-
```
r-owen: Why did you remove this code? I'm sure we can live without it, but I am curious. I think it would work to leave it in and call self._plot. Is it because of all that event handling code that was not called previously?

python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -430,21 +428,35 @@ def selectStars(self, exposure, sourceCat, matches=None):
...
+        goodSources = [s for g, s in zip(good, sourceCat) if g]
+        for isStellar, source in zip(stellar, goodSources):
+            if isStellar:
+                starCat.append(source)
+
+        return Struct(
+            starCat=starCat,
+        )
+
+    def _plot(self, display, displayExposure, exposure, plotMagSize, mag, width,
```
r-owen: Making this a separate method is probably wise (though I'm sorry all that event handling is needed; it seems far too details for diagnostic plots, and like it belongs in the display library, not here). However, I find the design confusing in that:
- The main interface for plotting is to call to the private `_plot` method, which then calls a public `plot` function. It feels backwards to have to call private code which calls public code; usually things go the other way -- a public function will call a private function for some implementation. Some renaming would help.
- Please add a doc string to the `_plot` method. It's not clear to me what some of the arguments are.

python/lsst/meas/algorithms/objectSizeStarSelector.py
parejkoj: That didn't get removed, just moved to `self._plot()`. The silly `fig=None` was removed, though.

python/lsst/meas/algorithms/objectSizeStarSelector.py
parejkoj: Hmm... in the interest of not mucking with this mess any more than I have to, I'm going to get rid of the private _plot() method I made (which may have changed how the plots were created), and just keep the moved matplotlib import.

PR #56, opened 2016-12-17
Tickets/DM-8645: Implement new matcherSourceSelector object for use in matchOptimisticB code

morriscb:
Implemented new matcherSourceSelctor for use in meas_astrom with matchOptimisticB.

### DIFF
--- a/python/lsst/meas/algorithms/matcherSourceSelector.py
+++ b/python/lsst/meas/algorithms/matcherSourceSelector.py
@@ -1,0 +1,170 @@
+#
+# LSST Data Management System
+# Copyright 2008-2015 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+from __future__ import absolute_import, division, print_function
+
+import numpy as np
+
+from lsst.afw import table
+import lsst.pex.config as pexConfig
+from .sourceSelector import BaseSourceSelectorConfig, BaseSourceSelectorTask, sourceSelectorRegistry
+from lsst.pipe.base import Struct
+from functools import reduce
+
+
+class MatcherSourceSelectorConfig(BaseSourceSelectorConfig):
+    sourceFluxType = pexConfig.Field(
+        doc="Type of source flux; typically one of Ap or Psf",
+        dtype=str,
+        default="Ap",
+    )
+    minSnr = pexConfig.Field(
+        dtype=float,
+        doc="Minimum allowed signal-to-noise ratio for sources used for matching "
+        "(in the flux specified by sourceFluxType); <= 0 for no limit",
+        default=40,
+    )
+
+
+class MatcherSourceSelectorTask(BaseSourceSelectorTask):
+    """
+    !Select sources that are useful for matching.
+
+    Good matching sources have high signal/noise, are non-blended. They need not
+    be PSF sources, just have reliable centroids.
+    """
+    ConfigClass = MatcherSourceSelectorConfig
+
+    def __init__(self, *args, **kwargs):
+        BaseSourceSelectorTask.__init__(self, *args, **kwargs)
+
+    def selectSources(self, sourceCat, matches=None):
+        """
+        !Return a catalog of sources: a subset of sourceCat.
+
+        If sourceCat is cotiguous in memory, will use vectorized tests for ~100x
+        execution speed advantage over non-contiguous catalogs. This would be
+        even faster if we didn't have to check footprints for multiple peaks.
+
+        @param[in] sourceCat  catalog of sources that may be sources
+                                (an lsst.afw.table.SourceCatalog)
+
+        @return a pipeBase.Struct containing:
+        - sourceCat  a catalog of sources
+        """
+        self._getSchemaKeys(sourceCat.schema)
+
+        if sourceCat.isContiguous():
+            good = self._isUsable_vector(sourceCat)
+            result = sourceCat[good]
+        else:
+            result = table.SourceCatalog(sourceCat.table)
+            for i, source in enumerate(sourceCat):
+                if self._isUsable(source):
+                    result.append(source)
+        return Struct(sourceCat=result)
+
+    def _getSchemaKeys(self, schema):
+        """Extract and save the necessary keys from schema with asKey."""
+        self.parentKey = schema["parent"].asKey()
+        self.nChildKey = schema["deblend_nChild"].asKey()
+        self.centroidXKey = schema["slot_Centroid_x"].asKey()
+        self.centroidYKey = schema["slot_Centroid_y"].asKey()
+        self.centroidFlagKey = schema["slot_Centroid_flag"].asKey()
+
+        fluxPrefix = "slot_%sFlux_" % (self.config.sourceFluxType,)
+        self.fluxField = fluxPrefix + "flux"
+        self.fluxKey = schema[fluxPrefix + "flux"].asKey()
+        self.fluxFlagKey = schema[fluxPrefix + "flag"].asKey()
+        self.fluxSigmaKey = schema[fluxPrefix + "fluxSigma"].asKey()
+
+    def _isMultiple_vector(self, sourceCat):
+        """Return True for each source that is likely multiple sources."""
+        test = (sourceCat.get(self.parentKey) != 0) | (sourceCat.get(self.nChildKey) != 0)
+        # have to currently manage footprints on a source-by-source basis.
+        for i, cat in enumerate(sourceCat):
+            footprint = cat.getFootprint()
+            test[i] |= (footprint is not None) and (len(footprint.getPeaks()) > 1)
+        return test
+
+    def _isMultiple(self, source):
+        """Return True if source is likely multiple sources."""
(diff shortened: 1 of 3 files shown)


parejkoj:
Where's the unittest? It doesn't appear to have come along with these changes.

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -0,0 +1,171 @@
...
+        "(in the flux specified by sourceFluxType); <= 0 for no limit",
+        default=40,
+    )
+
+
+class MatcherSourceSelectorTask(BaseSourceSelectorTask):
+    """
+    !Select sources that are useful for astrometry.
+
+    Good astrometry sources have high signal/noise, are non-blended, and
```
parejkoj: You need to change this docstring, since it's still verbatim the one from astrometrySourceSelector. Might be worth being more explicit about the differences between this and astrometrySS (in both of their docstrings). Also please take out all the preceding `!` and reformat things as numpydoc, instead of doxygen.

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -0,0 +1,171 @@
...
+    be PSF sources, just have reliable centroids.
+    """
+    ConfigClass = MatcherSourceSelectorConfig
+
+    def __init__(self, *args, **kwargs):
+        BaseSourceSelectorTask.__init__(self, *args, **kwargs)
+
+    def selectSources(self, sourceCat, matches=None):
+        """
+        !Return a catalog of sources: a subset of sourceCat.
```
parejkoj: Please reformat this as numpydoc.

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -0,0 +1,171 @@
...
+        @param[in] sourceCat  catalog of sources that may be sources
+                                (an lsst.afw.table.SourceCatalog)
+
+        @return a pipeBase.Struct containing:
+        - sourceCat  a catalog of sources
+        """
+        self._getSchemaKeys(sourceCat.schema)
+
+        if sourceCat.isContiguous():
+            good = self._isUsable_vector(sourceCat)
```
parejkoj: I think you still want to remove things with the "bad" flag here. Do you really want cosmics, interpolated pixels and stuff on the edge of a chip?

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: This was was as is and is straight from the astrometrySourceSelector. Not sure how to make it a numpydoc and if we would need to change the base class as well.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: I do want interpolated pixels and edge of chip things. Which specific bad flag tests for cosmics? I can add that if we want. I'm mostly just following the tests verbatim from the old SourceInfo class in matchOptimisticB.py which does not test for those things.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: As stated in the other comment I'm not aware of the new docstring standard. This is copied verbatim from astrometrySourceSelector. I can re-do it if absolutely necessary. I have already edited the doc strings and will push the changes shortly.

morriscb:
Added unittest for matcherSourceSelector to the ticket after it was erroneously not added. Have not yet modified the doc strings as these are the same as currently implemented in the base source selector and astrometrySourceSelector. Perhaps we should open a new ticket to modify all of these?

Also I have not yet added in isBad flags as none of these flags were used in the original SourceInfo for matchOptimisticB which matcherSourceSelector is based on. Can add these in if completely necessary.

python/lsst/meas/algorithms/matcherSourceSelector.py
parejkoj: Ok, I guess we'll hold off on reformatting as numpydoc (I did astrometrySS before that RFC had been fully settled). New python code is allowed (an I believe recommended) to use numpydoc instead of doxygen.

python/lsst/meas/algorithms/matcherSourceSelector.py
parejkoj: The base class I think already defines all the ones you want in the `badFlags` config parameter. Cosmics are crCenter, I think.

I really do think you should consider what the "correct" source selection criteria are, instead of just reproducing SourceInfo: there's really no guarantee it was fully thought out.

tests/testMatcherSourceSelector.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def setUp(self):
+        schema = lsst.meas.base.tests.TestDataset.makeMinimalSchema()
+        schema.addField("slot_ApFlux_flux", type=float)
+        schema.addField("slot_ApFlux_fluxSigma", type=float)
+        for flag in badFlags:
+            schema.addField(flag, type="Flag")
+
+        self.src = afwTable.SourceCatalog(schema)
+        self.sourceSelector = sourceSelector.sourceSelectorRegistry['matcher']()
```
parejkoj: I think here you can just construct the matcherSourceSelector directly (unless I'm missing something). On the other hand, this does test that you correctly put it in the registry, so...

tests/testMatcherSourceSelector.py
```diff
@@ -0,0 +1,147 @@
...
+        for i in range(5):
+            add_good_source(self.src, i)
+        result = self.sourceSelector.selectSources(self.src)
+        # TODO: assertEqual doesn't work correctly on source catalogs.
+        # self.assertEqual(result.sourceCat, self.src)
+        for x in self.src['id']:
+            self.assertIn(x, result.sourceCat['id'])
+
+    def testSelectSources_bad(self):
+        for i, flag in enumerate(badFlags):
```
parejkoj: Doesn't matcherSourceSelector ignore badFlags? So how does this test pass?

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: Isn't this more of a new ticket then? As in implementing the source selector framework first and then testing a new selection criteria later? Wouldn't that make it easier to track the changes to the code? Or am I over thinking this?

tests/testMatcherSourceSelector.py
morriscb: Some of the badFlags are redundant with tests preformed in the isGood test. This is true of the astrometrySelector as well. This is why the is badFlag test still passes.

tests/testMatcherSourceSelector.py
morriscb: I think we can leave it as is if that a necessary test, that is checking to see that it is properly registered. Again, if we change this we will should also modify the tests for astrometrySourceSelector as it is doing a similar thing.

tests/testMatcherSourceSelector.py
```diff
@@ -0,0 +1,147 @@
...
+import unittest
+import numpy as np
+
+import lsst.afw.table as afwTable
+from lsst.meas.algorithms import sourceSelector
+import lsst.meas.base.tests
+import lsst.utils.tests
+
+
+badFlags = [
```
morriscb: Hey here's the current text I've written. Let me know if this is what you want and I'll commit it.

"
# badFlags are flags that, if the bad flag is set the objecct should be
# imediately thrown out. This is as opposed to combinations of flags
# signifying good or bad data. Such flags should then be a part of the
# isGood test
"

python/lsst/meas/algorithms/matcherSourceSelector.py
parejkoj: That's a fair point. File the ticket(s) about figuring out what "correct" source selection for matcherSS would mean, and link them from 8645.

tests/testMatcherSourceSelector.py
parejkoj: Such bold, much wow...

That's probably fine? Though those bad flags aren't used by matcherSS right now, right?

tests/testMatcherSourceSelector.py
parejkoj: All fair points.

tests/testMatcherSourceSelector.py
parejkoj: Its not redundancy that I'm worried about: if the flag is set and the test expects a failure, how can this test pass?

morriscb: Because it will still fail isGood which it is awesome testing.

tests/testMatcherSourceSelector.py
morriscb: That is correct. Again, this is so that it is identical to the SourceInfo test.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: Okay, will do.

tests/testMatcherSourceSelector.py
parejkoj: Since we talked about it: you can delete the _bad test, since you're not using bad flags at all. The two slot flags are really part of testing the negation of isGood (I realized now that testAstrometrySS does this in an unfortunate way). Once you figure out whether matcherSS needs bad flags, you can add a test back in.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: This issue has been filed as a ticket here: https://jira.lsstcorp.org/browse/DM-8987

tests/testMatcherSourceSelector.py
morriscb: I forget what we decided, was I going to put this test in to matcherSourceSelector or into BaseSourceSelector?

tests/testMatcherSourceSelector.py
morriscb: Also I believe we wanted to remove this test. Is this the case?

tests/testMatcherSourceSelector.py
parejkoj: This block is being removed from testMatcherSS, and there is no "testBaseSS" (can't be: its abstract). So, the above text could go into testAstrometrySS.

tests/testMatcherSourceSelector.py
parejkoj: Yes: see my comment just above.

PR #57, opened 2016-12-22
Review for DM-8714: Add unittest for base_SdssShape_psf measurement

jdswinbank:
This test is being added here in meas_algorithms rather than meas_base
(where SdssShape lives) due to the need to apply a spatially varying
psf model such that the psf is different at each position in the image.
This varying psf model is built with meas_algorithms' PcaPsf (which is
not accessible from meas_base).

### DIFF
--- a/tests/testSdssShapePsf.py
+++ b/tests/testSdssShapePsf.py
@@ -1,0 +1,151 @@
+#!/usr/bin/env python
+#
+# LSST Data Management System
+# Copyright 2008-2016 AURA/LSST.
+#
+# This product includes software developed by the
+# LSST Project (http://www.lsst.org/).
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


tests/testSdssShapePsf.py
```diff
@@ -0,0 +1,128 @@
...
+        self.assertFalse(result.getFlag(measBase.SdssShapeAlgorithm.PSF_SHAPE_BAD))
+
+    def testMeasureGoodPsf(self):
+        """Test that we measure shapes and record the PSF shape correctly"""
+        # Apply a varying PSF to the image to make sure we are measuring the psf at the source position
+        varyingPsf = self._computeVaryingPsf()
+        exposure, catalog = self._runMeasurementTask(psf=varyingPsf)
+        key = measBase.SdssShapeResultKey(catalog.schema["base_SdssShape"])
+        for record in catalog:
+            psfTruth = exposure.getPsf().computeShape(afwGeom.Point2D(record.getX(), record.getY()))
```
jdswinbank: I have absolute faith that the PSF really is varying exactly as you say it is.

But.. what if it wasn't? If we screwed up and `varyingPsf` was actually constant across the image, then this test would pass even if the original bug was still in SdssShape.

I grant it's unlikely, but could we add an extra assertion here to establish that `psfTruth` really does vary by record?

laurenam: 👍

PR #60, opened 2017-01-20
Make loader agnostic to ingest name

SimonKrughoff:


tests/testHtmIndex.py
```diff
@@ -239,6 +242,16 @@ def testIngest(self):
         cent = make_coord(*tupl)
         cat = loader.loadSkyCircle(cent, self.search_radius, filterName='a')
         self.assertTrue(len(cat) > 0)
+        # test that a catalog can be loaded even with a name not used for ingestion
+        butler = dafPersist.Butler(self.test_repo_path)
+        config = LoadIndexedReferenceObjectsConfig()
+        config.ref_dataset_name = "diff_ref_name"
```
PaulPrice: This is a magic string used in multiple places.  Can you consolidate them into a single variable?

tests/testHtmIndex.py
```diff
@@ -239,6 +242,16 @@ def testIngest(self):
...
+        loader = LoadIndexedReferenceObjectsTask(butler=butler, config=config)
+        # This location is known to have objects
+        tupl = (93.0, -90.0)
+        cent = make_coord(*tupl)
```
PaulPrice: Is there a reason not to strip out the `tupl` variable and just to `make_coord(93.0, -90.0)`?

tests/testHtmIndex.py
SimonKrughoff: Good point.  Done.

tests/testHtmIndex.py
SimonKrughoff: Another good point.  I also consolidated so I don't define `cent` twice.

timj: Saves me from moaning that `tupl` is a terrible name for a variable 😉

PR #63, opened 2017-02-04
Review for DM-8030: Identify and correct differing background model between py2 and py3

jdswinbank:
Restores an integer division which had become a float division in the course
of the Python 3 port.

### DIFF
--- a/python/lsst/meas/algorithms/pcaPsfDeterminer.py
+++ b/python/lsst/meas/algorithms/pcaPsfDeterminer.py
@@ -40,6 +40,30 @@ from . import utils as maUtils
 
 __all__ = ["PcaPsfDeterminerConfig", "PcaPsfDeterminerTask"]
 
+def numCandidatesToReject(numBadCandidates, numIter, totalIter):
+    """Return the number of PSF candidates to be rejected.
+
+    The number of candidates being rejected on each iteration gradually
+    increases, so that on the Nth of M iterations we reject N/M of the bad
+    candidates.
+
+    Parameters
+    ----------
+    numBadCandidates : int
+        Number of bad candidates under consideration.
+
+    numIter : int
+        The number of the current PSF iteration.
+
+    totalIter : int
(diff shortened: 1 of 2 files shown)


tests/testPsfDetermination.py
```diff
@@ -472,6 +473,19 @@ def testRejectBlends(self):
...
+class PsfCandidateTestCase(lsst.utils.tests.TestCase):
+    def testNumToReject(self):
+        """Reject the correct number of PSF candidates on each iteration"""
+        # Numerical values correspond to the problem case identified in
+        # DM-8030.
+
+        nCand = 5
+        nIter = 3
+
+        for it, value in ((0, 1), (1, 3), (2, 5)):
```
mrawls: The variables in this test would be a little easier to follow if they used the same names as those in your nice new numCandidatesToReject function (i.e., numBadCandidates, numIter, totalIter).

python/lsst/meas/algorithms/pcaPsfDeterminer.py
```diff
@@ -369,7 +393,8 @@ def determinePsf(self, exposure, psfCandidateList, metadata=None, flagKey=None):
                         badCandidates.append(cand)
 
             badCandidates.sort(key=lambda x: x.getChi2(), reverse=True)
-            numBad = int(len(badCandidates) * (iter + 1) / self.config.nIterForPsf + 0.5)
+            numBad = numCandidatesToReject(len(badCandidates), iter,
```
timj: Can we change the variable name `iter` so it doesn't clash with a python function name? It's a bit confusing at first glance. I thought for a moment that you were passing the function reference into the routine.

jdswinbank: It's icky, but I was reluctant to change it since it's not really related to this ticket. I don't suppose it'll do any harm, so maybe I should just go ahead.

PR #70, opened 2017-03-02
DM-9188: Cleanup pybind11 code in core meas and ip packages

pschella:


python/lsst/meas/algorithms/__init__.py
```diff
@@ -23,7 +23,26 @@
 """lsst.meas.algorithms
 """
 from __future__ import absolute_import
-from .algorithmsLib import *
+
+import lsst.afw.image
+import lsst.afw.math
+from lsst.afw.image import DefectBase
```
TallJimbo: Do we actually refer to `lsst.meas.algorithms.DefectBase`?  That might be a mistake we want to fix.

python/lsst/meas/algorithms/cr.cc
```diff
@@ -35,14 +35,14 @@ namespace algorithms {
 
 namespace {
 template <typename PixelT>
-void declareFindCosmicRays(py::module& mod) {
+static void declareFindCosmicRays(py::module& mod) {
```
TallJimbo: Should have a newline between namespace open and the template declaration.  Would like another one after the function, and ideally a comment on the namespace close.

python/lsst/meas/algorithms/gaussianPsfFactory.py
```diff
@@ -23,7 +23,7 @@
 import math
 
 from lsst.pex.config import Config, Field, ConfigurableField
-from .algorithmsLib import SingleGaussianPsf, DoubleGaussianPsf
+from . import SingleGaussianPsf, DoubleGaussianPsf
```
TallJimbo: More direct import preferred.

python/lsst/meas/algorithms/interp.cc
```diff
@@ -34,32 +34,32 @@ namespace algorithms {
 
 namespace {
 template <typename PixelT>
-void declareInterpolateOverDefects(py::module& mod) {
+static void declareInterpolateOverDefects(py::module& mod) {
```
TallJimbo: Add newlines as in previous comment.
\

python/lsst/meas/algorithms/interp.cc
```diff
@@ -34,32 +34,32 @@ namespace algorithms {
...
+            .value("NEAR_LEFT", Defect::DefectPosition::NEAR_LEFT)
+            .value("WIDE_LEFT", Defect::DefectPosition::WIDE_LEFT)
+            .value("MIDDLE", Defect::DefectPosition::MIDDLE)
+            .value("WIDE_NEAR_LEFT", Defect::DefectPosition::WIDE_NEAR_LEFT)
+            .value("WIDE", Defect::DefectPosition::WIDE)
+            .value("WIDE_NEAR_RIGHT", Defect::DefectPosition::WIDE_NEAR_RIGHT)
+            .value("NEAR_RIGHT", Defect::DefectPosition::NEAR_RIGHT)
+            .value("WIDE_RIGHT", Defect::DefectPosition::WIDE_RIGHT)
+            .value("RIGHT", Defect::DefectPosition::RIGHT)
+            .export_values();
```
TallJimbo: Is it `clang-format` that wants this extra indentation for chained method calls?

python/lsst/meas/algorithms/makeCoaddApCorrMap.py
```diff
@@ -1,5 +1,5 @@
 from lsst.afw.image import ApCorrMap
-from .algorithmsLib import CoaddBoundedField, CoaddBoundedFieldElement
+from . import CoaddBoundedField, CoaddBoundedFieldElement
```
TallJimbo: More direct imports, please.

python/lsst/meas/algorithms/pcaPsfDeterminer.py
```diff
@@ -35,7 +35,7 @@
 import lsst.afw.display.ds9 as ds9
 import lsst.afw.math as afwMath
 from .psfDeterminer import BasePsfDeterminerTask, psfDeterminerRegistry
-from . import algorithmsLib
+from . import PsfCandidateF, createKernelFromPsfCandidates, countPsfCandidates, fitSpatialKernelFromPsfCandidates, PcaPsf, fitKernelParamsToImage
```
TallJimbo: More direct imports.

python/lsst/meas/algorithms/psfCandidate/__init__.py
```diff
@@ -0,0 +1,4 @@
+from __future__ import absolute_import
```
TallJimbo: I've probably missed a few instances of this before, but we should probably be adding license statements to all these `__init__.py` files.

python/lsst/meas/algorithms/psfCandidate/psfCandidateContinued.py
```diff
@@ -0,0 +1,15 @@
+from __future__ import absolute_import
+
+from lsst.utils import continueClass
+from .psfCandidate import PsfCandidateF
+
+__all__ = []  # import only for side effects
```
TallJimbo: I'm sure I've missed instances of this, too, but as @r-owen recently pointed out on the meas_modelfit review, we should be putting `__all__` above the imports (but below `from __future__`) according to PEP8.

python/lsst/meas/algorithms/secondMomentStarSelector.py
```diff
@@ -36,7 +36,7 @@
 import lsst.afw.image as afwImage
 import lsst.afw.math as afwMath
 import lsst.afw.geom as afwGeom
-from . import algorithmsLib
+from . import makePsfCandidate, DoubleGaussianPsf
```
TallJimbo: Direct imports.

python/lsst/meas/algorithms/starSelector.py
```diff
@@ -29,7 +29,7 @@
 import lsst.afw.math as afwMath
 import lsst.pex.config as pexConfig
 import lsst.pipe.base as pipeBase
-from . import algorithmsLib
+from . import makePsfCandidate
 from future.utils import with_metaclass
 
 __all__ = ["BaseStarSelectorConfig", "BaseStarSelectorTask", "starSelectorRegistry"]
```
TallJimbo: Move `__all__` up.

python/lsst/meas/algorithms/gaussianPsfFactory.py
r-owen: Can you please clarify what you are asking for? This looks like a standard import to me.

TallJimbo: Sorry, I started using shorthand on this point at some point, but that's not helpful to someone who didn't read my comments in the same order I made them.

When you import a non-module symbol directly from the current package, you're assuming the `__init__.py` for the package imports the module that defines those symbols and lifts them all to the package namespace *before* the current module is imported by that `__init__.py` file.  The alternative is
```
from .singleGaussianPsf import SingleGaussianPsf
from .doubleGaussianPsf import DoubleGaussianPsf
``
So while doing it @pschella's way protects against this module having to know what subpackage/module defines the symbols it wants, my way does not depend on the order modules are imported in `__init__.py`.  After seeing how common this is, I'm no longer completely convinced mine is better (as per my JIRA comment).

r-owen: Thanks for the clarification. For the record: I like your way better as I think it better guards against circular imports.

python/lsst/meas/algorithms/__init__.py
pschella: We use it as a base. Now added `py::module::import("lsst.afw.image");` instead.

pschella: Or should we use `py::module::import("lsst.afw.image._defect");`? Do we have policy for imports to use lowest level here? Of course this will have to be changed to `defect` once the `afw` cleanup is merged.

python/lsst/meas/algorithms/cr.cc
pschella: Done in all files.

python/lsst/meas/algorithms/gaussianPsfFactory.py
pschella: done

python/lsst/meas/algorithms/__init__.py
TallJimbo: I would use `py::module::import("lsst.afw.image")`.  I think we should import at sibling level only within the same package.

python/lsst/meas/algorithms/psfCandidate/__init__.py
pschella: Only on `__init__` files? That would be nice :)

r-owen: But...really? They don't contain *any* interesting code. I'd personally skip it if they are not already present.

PR #71, opened 2017-03-24
Tickets/dm 9750: Clean up and stackify matchPessimisticB.

morriscb:
DM-9750: Clean up and stackify matchPessimisticB.

### DIFF
--- a/python/lsst/meas/algorithms/matcherSourceSelector.py
+++ b/python/lsst/meas/algorithms/matcherSourceSelector.py
@@ -1,8 +1,8 @@
-# 
+#
 # LSST Data Management System
 #
 # Copyright 2008-2017  AURA/LSST.
-# 
+#
 # This product includes software developed by the
 # LSST Project (http://www.lsst.org/).
 #
@@ -10,14 +10,14 @@
 # it under the terms of the GNU General Public License as published by
 # the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
-# 
+#
 # This program is distributed in the hope that it will be useful,
 # but WITHOUT ANY WARRANTY; without even the implied warranty of
 # MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 # GNU General Public License for more details.
-# 
-# You should have received a copy of the LSST License Statement and 
-# the GNU General Public License along with this program.  If not, 
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
 from __future__ import absolute_import, division, print_function
@@ -26,7 +26,9 @@ import numpy as np
 
 from lsst.afw import table
 import lsst.pex.config as pexConfig
-from .sourceSelector import BaseSourceSelectorConfig, BaseSourceSelectorTask, sourceSelectorRegistry
+from .sourceSelector import BaseSourceSelectorConfig
+from .sourceSelector import BaseSourceSelectorTask
+from .sourceSelector import sourceSelectorRegistry
 from lsst.pipe.base import Struct
 from functools import reduce
 
@@ -39,8 +41,9 @@ class MatcherSourceSelectorConfig(BaseSourceSelectorConfig):
     )
     minSnr = pexConfig.Field(
         dtype=float,
-        doc="Minimum allowed signal-to-noise ratio for sources used for matching "
-        "(in the flux specified by sourceFluxType); <= 0 for no limit",
+        doc="Minimum allowed signal-to-noise ratio for sources used for "
+            "matching (in the flux specified by sourceFluxType); <= 0 for no "
+            "limit.",
         default=40,
     )
 
@@ -49,8 +52,8 @@ class MatcherSourceSelectorTask(BaseSourceSelectorTask):
     """
     !Select sources that are useful for matching.
 
-    Good matching sources have high signal/noise, are non-blended. They need not
-    be PSF sources, just have reliable centroids.
+    Good matching sources have high signal/noise, are non-blended. They need
+    not be PSF sources, just have reliable centroids.
     """
     ConfigClass = MatcherSourceSelectorConfig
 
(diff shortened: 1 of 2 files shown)


PaulPrice:
Please [phrase the commit summary in the imperative sense](https://developer.lsst.io/processes/workflow.html#writing-commit-summary-lines).

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -96,6 +96,21 @@ def _getSchemaKeys(self, schema):
...
 
+        self.edgeKey = schema["base_PixelFlags_flag_edge"].asKey()
+        self.interpolatedCenterKey = schema["base_PixelFlags_flag_interpolatedCenter"].asKey()
+        self.saturatedKey = schema["base_PixelFlags_flag_saturated"].asKey()
+
+    def _isGood_vector(self, sourceCat):
+        test = ~sourceCat.get(self.saturatedKey) & \
+               ~sourceCat.get(self.interpolatedCenterKey) & \
+               ~sourceCat.get(self.edgeKey)
+        return test
```
r-owen: Trivial: consider returning the value without saving it to `test` first, like `_isGood`

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -162,4 +177,38 @@ def _isUsable(self, source):
...
+        - have a valid flux (of the type specified in this object's constructor)
+        - have adequate signal-to-noise
+        """
+        return self._hasCentroid_vector(sourceCat) \
+            & self._isParent_vector(sourceCat) \
+            & self._goodSN_vector(sourceCat) \
+            & ~sourceCat.get(self.fluxFlagKey) \
+            & self._isGood_vector(sourceCat)
+
+    def _isUsable(self, source):
```
r-owen: Error: this is a second copy of this method (the first is at line 164)

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -162,4 +177,38 @@ def _isUsable(self, source):
             and not source.get(self.fluxFlagKey) \
             and self._goodSN(source)
 
+    def _isUsable_vector(self, sourceCat):
```
r-owen: Error: this is a second copy of the method!

Similarly _isUsable appears in two places.

The two implementations seem to differ slightly and I don't know which is correct.

Please always always lint your Python with flake8 before saving (or, better yet, while editing). It will show errors like this. Aside from these two duplicated methods the remaining violations are minor, though still worth fixing.

morriscb: Don't know why this didn't crash the code in the first place. The duplicates have been deleted. I had remembered to run pep8 on the meas_astrom part of this ticket but forgot to do it here before the pull request. I have run a linter on the code as of the most recent commit.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: Duplicates deleted.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: Changed it. The function now returns the array rather than saving a temporary variable.

morriscb:
New version of matcherSourceSelector up that includes @r-owen 's comments.

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -61,9 +64,8 @@ def selectSources(self, sourceCat, matches=None):
         """
         !Return a catalog of sources: a subset of sourceCat.
 
-        If sourceCat is cotiguous in memory, will use vectorized tests for ~100x
-        execution speed advantage over non-contiguous catalogs. This would be
-        even faster if we didn't have to check footprints for multiple peaks.
+        If sourceCat is cotiguous in memory, will use vectorized tests for
```
r-owen: "cotiguous" -> "contiguous"

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -119,16 +135,20 @@ def _hasCentroid(self, source):
         return np.all(np.isfinite(centroid)) and not source.getCentroidFlag()
 
     def _goodSN_vector(self, sourceCat):
-        """Return True for each source that has Signal/Noise > config.minSnr."""
+        """Return True for each source that has Signal/Noise > config.minSnr.
+        """
         if self.config.minSnr <= 0:
             return True
```
r-owen: This is a pre-existing issue, but please update the docs to say that it returns a scalar True if config.minSnr <= 0.

I'm guessing that works because numpy can perform mathematical operations on vectors with scalars. If there's any doubt then it's better to return [True]*len(sourceCat). But as long as this branch is tested it probably suffices to document what's going on.

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -26,7 +26,9 @@
 
 from lsst.afw import table
 import lsst.pex.config as pexConfig
-from .sourceSelector import BaseSourceSelectorConfig, BaseSourceSelectorTask, sourceSelectorRegistry
+from .sourceSelector import BaseSourceSelectorConfig
+from .sourceSelector import BaseSourceSelectorTask
+from .sourceSelector import sourceSelectorRegistry
```
r-owen: This seems far more verbose than the original, which can use a line continuation character if need be

python/lsst/meas/algorithms/matcherSourceSelector.py
r-owen: Unfortunately Python does not detect duplicate methods; it silently picks one.

Thanks for running a linter, though I hope it was flake8, not pep8. The former warns of important things like undefined variables and duplicate methods. The latter only warns about pep8 violations, which are much less interesting. (Then there's autopep8 that reformats code in an attempt to make it meet pep8 standards; it's not as smart as clang-format and can't fix everything, so I recommend running flake8 and manually inspecting the results afterwards).

morriscb:
Will push this pull request later as part of another ticket.

PR #72, opened 2017-03-29
Tickets/dm 9828

yalsayyad:


python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -272,19 +282,23 @@ def fitBackground(self, maskedImage, nx=0, ny=0, algorithm=None):
...
         @param[in] algorithm  name of interpolation algorithm; if None use self.config.algorithm
 
         @return fit background as an lsst.afw.math.Background
 
         @throw RuntimeError if lsst.afw.math.makeBackground returns None,
             which is apparently one way it indicates failure
         """
+
+        binSizeX = self.config.binSizeX if self.config.binSizeX else self.config.binSize
+        binSizeY = self.config.binSizeY if self.config.binSizeY else self.config.binSize
```
PaulPrice: Add explicit `== 0`?  Makes it clearer and corresponds more exactly with your config docs.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -327,29 +341,30 @@ def fitBackground(self, maskedImage, nx=0, ny=0, algorithm=None):
...
+                    raise ValueError("Too few points in grid (%d, %d) for order (%d) and binSize (%d, %d)" %
+                                     (nx, ny, order, binSizeX, binSizeY))
                 elif self.config.undersampleStyle == "REDUCE_INTERP_ORDER":
                     if order < 1:
                         raise ValueError("Cannot reduce approxOrder below 0.  " +
                                          "Try using undersampleStyle = \"INCREASE_NXNYSAMPLE\" instead?")
                     order = min(nx, ny) - 1
                     self.log.warn("Reducing approxOrder to %d" % order)
                 elif self.config.undersampleStyle == "INCREASE_NXNYSAMPLE":
+                    # Reduce bin size to the largest acceptable square bins
```
PaulPrice: Do they have to be square?  The `order` is (unfortunately) identical for both dimensions, but I don't think there's a reason the bin size has to be identical.

yalsayyad: They don't.  A goal with this ticket was not to change behavior of existing config. As I mentioned on the ticket comments, its not obvious what behavior is more intuitive and expected here: to reduce the bin size only in the dimension with too few data points or reduce the bin size in both dimensions. Given that everyone up until this point has specified square bins, I thought it would be least surprising if the bins were reduced in both dimensions as they were before this ticket.

python/lsst/meas/algorithms/subtractBackground.py
yalsayyad: :+1:

PR #74, opened 2017-04-13
Tickets/dm 8108

natelust:


src/CR.cc
```diff
@@ -282,7 +281,10 @@ void checkSpanForCRs(detection::Footprint *extras, // Extra spans get added to t
             }
             loc.image() = corr;
 
-            extras->addSpan(y + imageY0, x + imageX0, x + imageX0);
+            std::vector<geom::Span> spanList(extras->getSpans()->begin(),
+                                              extras->getSpans()->end());
+            spanList.push_back(geom::Span(y + imageY0, x + imageX0, x +imageX0));
+            extras->setSpans(std::make_shared<geom::SpanSet>(std::move(spanList)));
```
natelust: We spoke about this in person, but just to note here for posterity, this behavior is intended such that a spanSet is an immutable object to protect any other code that has a shared pointer to a  SpanSet from predictable behavior if someone modifies the SpanSet member of a Footprint. I will make a note in the documentation on the addSpan member about this, and encourage if multiple spans are to be added it would be better to construct a vector of spans, and construct a new SpanSet once, and set it back with setSpans

src/CR.cc
```diff
@@ -293,18 +295,14 @@ void checkSpanForCRs(detection::Footprint *extras, // Extra spans get added to t
...
         _sum(0.0) {}
 
-    // method called for each pixel by apply()
-    void operator()(typename ImageT::xy_locator loc, // locator pointing at the pixel
-                    int, int
-                   ) {
-        _sum += *loc - _bkgd;
+    void operator()(geom::Point2I const & point, ImageT const & val){
+        _sum += val - _bkgd;
     }
```
natelust: anything that makes use of the applyFunctor method must have a call signature of a Point2I as its first argument. All other arguments are optional and depend only on how applyFunctor is called.

PR #76, opened 2017-04-21
DM-10271: Only use tempLocalBackground to detect Peaks.

TallJimbo:
We now detect Footprints with the local background still in the image, and only detect Peaks with it subtracted.

### DIFF
--- a/python/lsst/meas/algorithms/detection.py
+++ b/python/lsst/meas/algorithms/detection.py
@@ -103,6 +103,12 @@ class SourceDetectionConfig(pexConfig.Config):
         doc="Do temporary interpolated background subtraction before footprint detection?",
         default=True,
     )
+    nPeaksMaxSimple = pexConfig.Field(
+        dtype=int,
+        doc=("The maximum number of peaks in a Footprint before trying to "
+             "replace its peaks using the temporary local background"),
+        default=1,
+    )
 
     def setDefaults(self):
         self.tempLocalBackground.binSize = 64
@@ -326,8 +332,13 @@ into your debug.py file and run measAlgTasks.py with the \c --debug flag.
             del mask
 
         if self.config.doTempLocalBackground:
-            tempBgRes = self.tempLocalBackground.run(exposure)
-            tempLocalBkgdImage = tempBgRes.background.getImage()
+            # Estimate the background, but add it back in instead of leaving
+            # it subtracted (for now); we'll want to smooth before we
+            # subtract it.
+            tempBg = self.tempLocalBackground.fitBackground(
+                exposure.getMaskedImage()
+            )
+            tempLocalBkgdImage = tempBg.getImageF()
 
         if sigma is None:
             psf = exposure.getPsf()
@@ -366,10 +377,37 @@ into your debug.py file and run measAlgTasks.py with the \c --debug flag.
 
         fpSets = pipeBase.Struct(positive=None, negative=None)
 
+        # Detect the Footprints (peaks may be replaced if doTempLocalBackground)
         if self.config.thresholdPolarity != "negative":
-            fpSets.positive = self.thresholdImage(middle, "positive")
+            threshold = self.makeThreshold(middle, "positive")
+            fpSets.positive = afwDet.FootprintSet(
+                middle,
+                threshold,
+                "DETECTED",
+                self.config.minPixels
+            )
         if self.config.reEstimateBackground or self.config.thresholdPolarity != "positive":
-            fpSets.negative = self.thresholdImage(middle, "negative")
+            threshold = self.makeThreshold(middle, "negative")
+            fpSets.negative = afwDet.FootprintSet(
+                middle,
+                threshold,
+                "DETECTED_NEGATIVE",
+                self.config.minPixels
+            )
+
+        if self.config.doTempLocalBackground:
+            # Subtract the local background from the smoothed image. Since we
+            # never use the smoothed again we don't need to worry about adding
+            # it back in.
+            tempLocalBkgdImage = tempLocalBkgdImage.Factory(tempLocalBkgdImage,
+                                                            middle.getBBox())
+            middle -= tempLocalBkgdImage
+            thresholdPos = self.makeThreshold(middle, "positive")
+            thresholdNeg = self.makeThreshold(middle, "negative")
+            if self.config.thresholdPolarity != "negative":
+                self.updatePeaks(fpSets.positive, middle, thresholdPos)
+            if self.config.thresholdPolarity != "positive":
+                self.updatePeaks(fpSets.negative, middle, thresholdNeg)
 
         for polarity, maskName in (("positive", "DETECTED"), ("negative", "DETECTED_NEGATIVE")):
             fpSet = getattr(fpSets, polarity)
@@ -390,9 +428,6 @@ into your debug.py file and run measAlgTasks.py with the \c --debug flag.
         if self.config.thresholdPolarity != "negative":
             self.log.info("Detected %d positive sources to %g sigma.",
                           fpSets.numPos, self.config.thresholdValue*self.config.includeThresholdMultiplier)
-
-        if self.config.doTempLocalBackground:
-            maskedImage += tempLocalBkgdImage
 
         fpSets.background = None
         if self.config.reEstimateBackground:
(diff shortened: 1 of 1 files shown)


python/lsst/meas/algorithms/detection.py
```diff
@@ -366,11 +370,36 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True):
...
+            middle += tempLocalBkgdImage
+
+        # Detect the Footprints (peaks may be replaced if doTempLocalBackground)
         if self.config.thresholdPolarity != "negative":
             fpSets.positive = self.thresholdImage(middle, "positive")
         if self.config.reEstimateBackground or self.config.thresholdPolarity != "positive":
             fpSets.negative = self.thresholdImage(middle, "negative")
 
+        if self.config.doTempLocalBackground:
+            # Replace the Peaks in the Footprints from those detected with the
```
laurenam: from --> with ?

python/lsst/meas/algorithms/detection.py
```diff
@@ -366,11 +370,36 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True):
...
+            # temporary local background.
```
laurenam: temporary --> temporarily subtracted ?

python/lsst/meas/algorithms/detection.py
```diff
@@ -489,6 +515,38 @@ def setEdgeBits(maskedImage, goodBBox, edgeBitmask):
...
+    @pipeBase.timeMethod
+    def replacePeaks(self, targetFpList, goodPeakFpList):
+        """Replace all of the peaks in one FootprintSet with the overlapping ones from another.
+
+        When this would remove all of the Peaks from a Footprint, we instead remove all but the
+        brightest.
+
+        Parameters
+        ----------
+        targetFpSet : `list`
```
PaulPrice: `list` of what?

Ditto below.

python/lsst/meas/algorithms/detection.py
```diff
@@ -489,6 +515,38 @@ def setEdgeBits(maskedImage, goodBBox, edgeBitmask):
...
+            newPeaks = afwDet.PeakCatalog(targetFp.getPeaks().getTable())
+            for goodPeakFp in goodPeakFpList:
+                if not targetBBox.overlaps(goodPeakFp.getBBox()):
+                    continue
+                for candidatePeak in goodPeakFp.getPeaks():
+                    if targetFp.contains(candidatePeak.getI()):
+                        newPeaks.append(candidatePeak)
+            targetPeaks = targetFp.getPeaks()
+            if len(newPeaks) == 0:
+                del targetPeaks[1:]
```
PaulPrice: Does this deletion leave memory non-contiguous and therefore make things more troublesome later?

python/lsst/meas/algorithms/detection.py
```diff
@@ -489,6 +515,38 @@ def setEdgeBits(maskedImage, goodBBox, edgeBitmask):
...
+        ----------
+        targetFpSet : `list`
+            Sequence containing the Footprints whose Peaks we want to replace.
+        goodPeakFpSet : `list`
+            Sequence containing the Footprints whose Peaks we want to use instead.
+        """
+        for targetFp in targetFpList:
+            targetBBox = targetFp.getBBox()
+            newPeaks = afwDet.PeakCatalog(targetFp.getPeaks().getTable())
+            for goodPeakFp in goodPeakFpList:
```
PaulPrice: This double `for` is going to chew up the runtime.  Can we do it a bit more cleverly?  Maybe remove items from the `goodPeakFpList` when they get copied over?  Can you order the footprints so that you can cease iterating (no need to even check the bbox overlap) once you get past a certain point?

python/lsst/meas/algorithms/detection.py
```diff
@@ -489,6 +515,38 @@ def setEdgeBits(maskedImage, goodBBox, edgeBitmask):
...
+                if not targetBBox.overlaps(goodPeakFp.getBBox()):
+                    continue
+                for candidatePeak in goodPeakFp.getPeaks():
+                    if targetFp.contains(candidatePeak.getI()):
+                        newPeaks.append(candidatePeak)
```
PaulPrice: This is going to break memory contiguity, right?  Can you throw in a `reserve` statement in order to preserve contiguity (maybe put all the matching peaks in a list, then `reserve` and `append` all of them when you have the complete list)?
Ditto on the `extend` later.

python/lsst/meas/algorithms/detection.py
```diff
@@ -326,8 +326,12 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True):
             del mask
 
         if self.config.doTempLocalBackground:
+            # Estimate the background, but add it back in instead of leaving
+            # it subtracted (for now); we'll want to smooth before we
+            # subtract it.
             tempBgRes = self.tempLocalBackground.run(exposure)
             tempLocalBkgdImage = tempBgRes.background.getImage()
+            maskedImage += tempLocalBkgdImage
```
TallJimbo: Aha!  I didn't realize it existed; will do.

python/lsst/meas/algorithms/detection.py
```diff
@@ -366,11 +370,36 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True):
 
         fpSets = pipeBase.Struct(positive=None, negative=None)
 
+        if self.config.doTempLocalBackground:
+            # Subtract the local background from the smoothed image.
```
TallJimbo: We should definitely estimate it before subtracting (otherwise the smoothing messes up the sigma-clipping), and we do.  From there, I don't think it matters when we subtract it, since even this background should have no power on scales as small as the PSF.

PaulPrice: One could imagine using a pretty small-scale background, so I'm not sure your assumption holds.

python/lsst/meas/algorithms/detection.py
TallJimbo: I don't think I can remove things from `goodPeakFpList`; not only is deleting in a loop tricky, it's possible one of those footprints could overlap multiple footprints in the other list.

But sorting beforehand (maybe by ymax?) is a good idea.  I'll have a go at implementing that.

python/lsst/meas/algorithms/detection.py
TallJimbo: It will break memory contiguity, but I believe in practice we'll copy over the vast majority of the new peaks in the same order they were allocated and remove the vast majority of the old peaks, so in the absence of profiling I don't think there's a strong case to be made on deep vs. shallow copy here.

python/lsst/meas/algorithms/detection.py
TallJimbo: But if that's true, then
```
add_background(smooth(subtract_background(image))) != smooth(image)
```
but think we still have:
```
add_background(subtract_background(smooth(image))) == smooth(image)
```
so I think then we *definitely* need to smooth before subtracting.

python/lsst/meas/algorithms/detection.py
TallJimbo: It does make memory noncontiguous, but I don't think that matters.

python/lsst/meas/algorithms/detection.py
```diff
@@ -439,31 +474,79 @@ def plotPeaks(fps, ctype):
...
+        image : `afw.image.MaskedImage`
+            Image to measure noise statistics from if needed.
+        thresholdParity: `str`
+            One of "positive" or "negative", to set the kind of fluctuations
+            the Threshold will detect.
         """
         parity = False if thresholdParity == "negative" else True
-        threshold = afwDet.createThreshold(self.config.thresholdValue, self.config.thresholdType, parity)
+        threshold = afwDet.createThreshold(self.config.thresholdValue,
+                                           self.config.thresholdType, parity)
```
PaulPrice: A bunch of line length changes (autopep8?) have snuck in here.

python/lsst/meas/algorithms/detection.py
```diff
@@ -103,6 +103,12 @@ class SourceDetectionConfig(pexConfig.Config):
         doc="Do temporary interpolated background subtraction before footprint detection?",
         default=True,
     )
+    nPeaksMaxSimple = pexConfig.Field(
+        dtype=int,
+        doc=("The maximum number of peaks in a Footprint before try to replace"
```
PaulPrice: trying

python/lsst/meas/algorithms/detection.py
```diff
@@ -439,31 +474,79 @@ def plotPeaks(fps, ctype):
...
+            if len(oldPeaks) <= self.config.nPeaksMaxSimple:
+                continue
+            # We detect a new FootprintSet within each non-simple Footprint's
+            # bbox to avoid a big O(N^2) comparison between the two sets of
+            # Footprints.
+            sub = image.Factory(image, footprint.getBBox())
+            fpSetForPeaks = afwDet.FootprintSet(
+                sub,
+                threshold,
+                "", # don't set a mask plane
```
PaulPrice: Two spaces before `#`.

python/lsst/meas/algorithms/detection.py
```diff
@@ -439,31 +474,79 @@ def plotPeaks(fps, ctype):
...
+        Footprint, the brightest original Peak in that Footprint is kept.
+        """
+        for footprint in fpSet.getFootprints():
+            oldPeaks = footprint.getPeaks()
+            if len(oldPeaks) <= self.config.nPeaksMaxSimple:
+                continue
+            # We detect a new FootprintSet within each non-simple Footprint's
+            # bbox to avoid a big O(N^2) comparison between the two sets of
+            # Footprints.
+            sub = image.Factory(image, footprint.getBBox())
```
PaulPrice: Just checking: the peaks found in the subimage have the xy0 added, right?

python/lsst/meas/algorithms/detection.py
TallJimbo: I thought I'd only made these sorts of changes in code I touched, but git is more clever than I expected about seeing changes instead of blocks of new and old code; all of these are in a new function, but it's pretty similar to an old one I removed.

python/lsst/meas/algorithms/detection.py
TallJimbo: Yes.  We've actually always relied on that, because we run detection on a trimmed image (representing the area we could smooth) and then assume those Footprints and Peaks are valid for the untrimmedone.

PR #77, opened 2017-04-27
DM-9192 add == and toString for new BoundedField API

parejkoj:
These were added to the Boundedfield API as part of the work on PhotoCalib.

### DIFF
--- a/src/CoaddBoundedField.cc
+++ b/src/CoaddBoundedField.cc
@@ -239,6 +239,15 @@ PTR(afw::math::BoundedField) CoaddBoundedField::operator*(double const scale) co
     throw LSST_EXCEPT(pex::exceptions::NotFoundError, "Scaling of CoaddBoundedField is not implemented");
 }
 
+bool CoaddBoundedField::operator==(BoundedField const& rhs) const {
+    auto rhsCasted = dynamic_cast<CoaddBoundedField const *>(&rhs);
+    if (!rhsCasted) return false;
+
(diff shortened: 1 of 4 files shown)


src/CoaddBoundedField.cc
```diff
@@ -239,6 +239,18 @@ PTR(afw::math::BoundedField) CoaddBoundedField::operator*(double const scale) co
...
 
+bool CoaddBoundedField::operator==(BoundedField const& rhs) const {
+    try {
+        auto& rhsCasted = dynamic_cast<CoaddBoundedField const&>(rhs);
+        return (getBBox() == rhsCasted.getBBox()) && (_default == rhsCasted._default) &&
+               ((*_coaddWcs) == (*rhsCasted._coaddWcs)) && (_elements == rhsCasted._elements);
+    }
+    catch (std::bad_cast&) {
+        return false;
+    }
```
TallJimbo: I'd recommend converting the try block to:
```
auto rhsCasted = dynamic_cast<CoaddBoundedfield const *>(&rhs);
if (!rhsCasted) {
    return false;
}
...
```
Some (most?) compilers make checking for `nullptr` *much* faster than throwing and catching an exception, so when you don't actually gain anything from exceptions (e.g. unwinding the stack gracefully) it's better to use the pointer form of `dynamic_cast`.

parejkoj: Thanks for the suggestion. I've implemented this here, and done the same for ChebyshevBoundedField.

PR #85, opened 2017-07-11
Enable Psfs to return resized copy of themselves

yalsayyad:


include/lsst/meas/algorithms/CoaddPsf.h
```diff
@@ -71,6 +71,8 @@ class CoaddPsf : public afw::table::io::PersistableFacade<CoaddPsf>, public Imag
     /// Polymorphic deep copy.  Usually unnecessary, as Psfs are immutable.
     virtual PTR(afw::detection::Psf) clone() const;
 
+    /// Return a clone with specified kernel dimensions
+    virtual PTR(afw::detection::Psf) resized(int width, int height) const;
```
pschella: We don't use `PTR` anymore.

pschella: Should be `override` instead of `virtual`, here and below (but see comment on `afw` PR about separate commits).

src/PcaPsf.cc
```diff
@@ -66,6 +66,12 @@ PTR(afw::detection::Psf) PcaPsf::clone() const {
     return std::make_shared<PcaPsf>(*this);
 }
 
+PTR(afw::detection::Psf) PcaPsf::resized(int width, int height) const {
+    PTR(afw::math::LinearCombinationKernel)
+    kern = std::static_pointer_cast<afw::math::LinearCombinationKernel>(getKernel()->resized(width, height));
```
pschella: Doesn't `resized` return the derived type (e.g. `LinearCombinationKernel`)?

tests/testGaussianPsf.py
```diff
@@ -189,6 +189,12 @@ def testKernelPsf(self):
         self.assertAlmostEqual(stats.getValue(afwMath.MAX), 0.0, places=16)
         self.assertAlmostEqual(stats.getValue(afwMath.MIN), 0.0, places=16)
 
+        for pad in [0, 2, -2]:
+            resizedKPsf = kPsf.resized(ksize + pad, ksize + pad)
+            self.assertEqual(resizedKPsf.computeBBox().getDimensions(),
+                             afwGeom.Extent2I(ksize + pad, ksize + pad))
+            self.assertEqual(resizedKPsf.getKernel().getKernelParameters(),
+                             kPsf.getKernel().getKernelParameters())
```
pschella: But not test the values?

tests/testSdssShapePsf.py
```diff
@@ -138,6 +138,19 @@ def testMeasureGoodPsf(self):
             psfResult = key.getPsfShape(record)
             self._checkPsfShape(result, psfResult, psfTruth)
 
+    def testResizedPcaPsf(self):
+        """Test that  PcaPsf can resize itself.
+
+        This test resides here because PcaPsfs do not have their own test module"""
```
pschella: Why not make one?

tests/testSdssShapePsf.py
```diff
@@ -138,6 +138,19 @@ def testMeasureGoodPsf(self):
...
+        psf = self._computeVaryingPsf()
+        dim = psf.computeBBox().getDimensions()
+        for pad in [0, 2, -2]:
+            resizedPsf = psf.resized(dim.getX() + pad, dim.getY() + pad)
+            self.assertEqual(resizedPsf.computeBBox().getDimensions(),
+                             afwGeom.Extent2I(dim.getX() + pad, dim.getY() + pad))
+            self.assertEqual(resizedPsf.getKernel().getKernelParameters(),
+                             psf.getKernel().getKernelParameters())
```
pschella: What about values?

src/PcaPsf.cc
yalsayyad: it returns a  `shared_ptr<lsst::afw::math::Kernel>`, but I did think these kinds of casts were automatic...
```
 src/PcaPsf.cc:49:9: note: candidate constructor not viable: no known conversion from 'shared_ptr<lsst::afw::math::Kernel>' to
      'shared_ptr<afw::math::LinearCombinationKernel>' for 1st argument
```

PR #86, opened 2017-07-21
DM-11338: CoaddPsf: add control class

PaulPrice:
This allows simple configuration of CoaddPsf from python.

### DIFF
--- a/include/lsst/meas/algorithms/CoaddPsf.h
+++ b/include/lsst/meas/algorithms/CoaddPsf.h
@@ -26,6 +26,7 @@
 
 #include <memory>
 #include "lsst/base.h"
+#include "lsst/pex/config.h"
 #include "lsst/meas/algorithms/ImagePsf.h"
 #include "lsst/afw/image/Wcs.h"
 #include "lsst/afw/table/Exposure.h"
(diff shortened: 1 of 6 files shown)


include/lsst/meas/algorithms/CoaddPsf.h
```diff
@@ -36,6 +37,18 @@
...
 namespace lsst { namespace meas { namespace algorithms {
 
+class CoaddPsfControl {
+public:
+
+    LSST_CONTROL_FIELD(warpingKernelName, std::string,
+                       "Name of warping kernel; choices: lanczos3,lanczos4,lanczos5,bilinear,nearest");
+    LSST_CONTROL_FIELD(cacheSize, int, "Warping kernel cache size");
+
+    explicit CoaddPsfControl(std::string _warpingKernelName="lanczos3", int _cacheSize=10000) :
```
TallJimbo: I don't believe we have an established convention for this, but in cases like this I tend to use a trailing underscore instead of a leading one.  Not a big deal either way.

python/lsst/meas/algorithms/__init__.py
```diff
@@ -70,3 +70,7 @@
 for name in dict(globals()):
     if name.endswith("_swigregister"):
         del globals()[name]
+
+from lsst.pex.config import makeConfigClass
+makeConfigClass(CoaddPsfControl, "CoaddPsfConfig")
+del makeConfigClass
```
TallJimbo: While it's a bigger change, I think it'd be more appropriate to move this to `coaddPsf/coaddPsfContinued.py` (and move `coaddPsf.cc` to `coaddPsf/coaddPsf.cc` and add an `__init__.py` there).

And while it's unrelated, would you mind zapping the Swig cleanup code just above this block?

PR #88, opened 2017-07-28
DM-11401: testHtmIndex: use lower HTM depth for faster test

PaulPrice:
This test used to take 372 sec because it used an HTM depth of 8 (mean pixel
area of 0.08 deg^2) and a 3 deg search radius. Reducing the HTM depth to 4
(mean pixel area of 20 deg^2, roughly matching the 28 deg^2 search area;
big enough to chunk the data reasonably, small enough to force the code
to check neighbours) allows the test to run in 27 sec.

Changing the HTM depth required regenerating a comparison file; added a hook
to make this easier in the future. The new comparison file has multiple
sources in it, so robustified the test to work regardless of the number of
sources.


tests/testHtmIndex.py
```diff
@@ -176,15 +179,20 @@ def testSanity(self):
...
+            os.unlink(self.test_cat_path)
+            ref_cat.writeFits(self.test_cat_path)
+            return
         ex1 = ref_cat.extract('*')
         ex2 = self.test_cat.extract('*')
-        # compare sets as the order may be different
-        self.assertDictEqual(ex1, ex2)
+        self.assertEqual(set(ex1.keys()), set(ex2.keys()))
+        for kk in ex1:
+            self.assertTrue(np.all(ex1[kk] == ex2[kk]))
```
r-owen: What is the benefit to this change? The original worked, though I'd be inclined to change `assertDictEqual` to `assertEqual` (the latter should call the former, both in Python 2.7 and Python 3). Does your new code give better error messages?

If you do want to use numpy to test each key individually, then I suggest using `numpy.testing.assert_equal`, as it's shorter and more explicit.

tests/testHtmIndex.py
```diff
@@ -218,7 +226,7 @@ def testIngest(self):
         default_config.mag_column_list = ['a', 'b']
         default_config.mag_err_column_map = {'a': 'a_err', 'b': 'b_err'}
         default_config.dataset_config.ref_dataset_name = 'myrefcat'
-        default_config.dataset_config.indexer.active.depth = 10
+        default_config.dataset_config.indexer.active.depth = self.depth - 1
```
r-owen: I'm puzzled by this change and by the old code. The old code used self.depth = 8 (any idea why?) which was not self.depth - 1. 

Please add a code comment explaining why this depth is chosen.

tests/testHtmIndex.py
```diff
@@ -176,15 +179,20 @@ def testSanity(self):
...
-        pix_id = 671901
+        pix_id = 2222
         dataset_name = IngestIndexedReferenceTask.ConfigClass().dataset_config.ref_dataset_name
         data_id = self.indexer.make_data_id(pix_id, dataset_name)
         self.assertTrue(self.test_butler.datasetExists('ref_cat', data_id))
         ref_cat = self.test_butler.get('ref_cat', data_id)
+        if REGENERATE_COMPARISON:
+            os.unlink(self.test_cat_path)
+            ref_cat.writeFits(self.test_cat_path)
+            return
```
r-owen: Please make it clear that `REGENERATE_COMPARISON` is true and the test will not be run. My suggestion is to call `self.fail` instead of returning, so it is impossible to accidentally leave this set.

PR #89, opened 2017-08-14
DM-5922: Rework camera geometry to use the replacement for XYTransform

r-owen:


python/lsst/meas/algorithms/gaussianPsfFactory.py
```diff
@@ -53,8 +53,8 @@ class GaussianPsfFactory(Config):
         check=isPositive,
     )
     sizeFactor = Field(
-        doc="Kernel size as a factor of fwhm (dimensionless); "
-        + "size = sizeFactor * fwhm; ignored if size is not None",
+        doc = "Kernel size as a factor of fwhm (dimensionless); " +
```
timj: I didn't think the spaces should be there for keyword args?

timj: Also, I don't think the space needs to be there at all does it?

r-owen: I believe spaces around = for keyword arguments are:
- encouraged for keyword arguments that are on their own line
- discouraged for other cases

TallJimbo: I checked, and @r-owen is correct according to the style guide.  This makes me sad; I don't think whitespace around `=` should be affected at all by how we split lines.  But I also don't have the energy to RFC a change.

python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -368,10 +368,9 @@ def selectStars(self, exposure, sourceCat, matches=None):
         dumpData = lsstDebug.Info(__name__).dumpData                   # dump data to pickle file?
 
         detector = exposure.getDetector()
-        pixToTanXYTransform = None
+        pixToTanPix = None
         if detector is not None:
-            tanSys = detector.makeCameraSys(TAN_PIXELS)
-            pixToTanXYTransform = detector.getTransformMap().get(tanSys)
+            pixToTanPix = detector.getTransform(TAN_PIXELS)
```
TallJimbo: Now that I'm seeing this interface in action in real code, I'm a bit bothered by how difficult it is to see what's going on.  This case is saved by the helpful `pixToTanPix` variable name, but I think `getTransform` should be replaced by a function that really says what's going on:
 - what other coordinate system is involved ("PIXELS")?
 - are we going from the given cordinate system ("TAN_PIXELS") to the implicit one ("PIXELS"), or the reverse?

To keep the meaning of the method the same, I'd recommend calling it something like `getTransformFromPixelsTo(target)`.  It's a lot more verbose, but a lot more readable.

Of course, this is all pre-existing, but it seems like a good time to fix it.

r-owen: I agree. Depending on how widely used this is, I'll either go back to `detector.getTransformMap().getTransform(fromSys, toSys)` or add a convenience method `detector.getTransform`

r-owen:
I fixed the primary complaint (which was well justified) and am much happier with the code now.

PR #94, opened 2017-10-24
DM-12207: Add color selection for photocal

PaulPrice:


python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
...
+
+class ColorLimit(pexConfig.Config):
+    """Select sources using a color limit
+
+    This object can be used as a `lsst.pex.config.Config` for configuring
+    the limit, and then the `apply` method can be used to test sources in
+    a catalog against the configured limit.
+
+    We refer to 'primary' and 'secondary' flux measurements; these are the
+    two components of the color, which is 'mag(primary) - mag(secondary)'.
```
parejkoj: I think the doc should say something like `instFluxToMag(cat[primary]) - instFluxToMag(cat[secondary])`, to make it explicit what the conversion is.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
 sourceSelectorRegistry = pexConfig.makeRegistry(
     doc="A registry of source selectors (subclasses of BaseSourceSelectorTask)",
 )
+
+
+class ColorLimit(pexConfig.Config):
```
parejkoj: I think these configs and the new SourceSelector below probably belong in their own `.py` file.

Also, I feel like the names of the new configs should be e.g. `ColorLimitConfig`?

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
...
+
+        Parameters
+        ----------
+        catalog : `lsst.afw.table.SourceCatalog`
+            Source to which to apply limits.
+
+        Returns
+        -------
+        selected : `numpy.ndarray`
+            Boolean array indicating whether each source is selected.
```
parejkoj: note: `True = selected` or somesuch (here and for each one below).

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
 sourceSelectorRegistry = pexConfig.makeRegistry(
     doc="A registry of source selectors (subclasses of BaseSourceSelectorTask)",
 )
+
+
+class ColorLimit(pexConfig.Config):
+    """Select sources using a color limit
+
+    This object can be used as a `lsst.pex.config.Config` for configuring
+    the limit, and then the `apply` method can be used to test sources in
```
parejkoj: "...can be used to identify sources in the catalog that match the configured limit.", or something like that. (and for each one below)

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
...
+            selected &= flux < self.maximum
+        return selected
+
+
+class MagnitudeLimit(pexConfig.Config):
+    """Select sources using a magnitude limit
+
+    Note that this assumes that a zero-point has already been applied and
+    the fluxes are in AB fluxes. It is therefore principally intended for
+    reference catalogs rather than catalogs extracted from science images.
```
parejkoj: Maybe make it explicit how flux (in Jansky!) is converted to magnitude in the docs, so the refcat vs. sourceCat distinction is clearer?

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
...
+    """
+    good = pexConfig.ListField(dtype=str, default=[],
+                               doc="List of source flag fields that must be set for a source to be used.")
+    bad = pexConfig.ListField(dtype=str, default=[],
+                              doc="List of source flag fields that must NOT be set for a source to be used.")
+
+    def apply(self, catalog):
+        """Apply the flag requirements to a catalog
+
+        Returns whether the source is selected.
```
parejkoj: "Return a boolean array specifying whether each source is selected." (and below)

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
...
+    flags = pexConfig.ConfigField(dtype=RequireFlags, doc="Flags to require")
+    unresolved = pexConfig.ConfigField(dtype=RequireUnresolved, doc="Star/galaxy separation to apply")
+
+    def setDefaults(self):
+        pexConfig.Config.setDefaults(self)
+        self.flags.bad = ["base_PixelFlags_flag_edge", "base_PixelFlags_flag_interpolated",
+                          "base_PixelFlags_flag_saturated"]
+
+class ScienceSourceSelectorTask(BaseSourceSelectorTask):
+    """Science source selector
```
parejkoj: What does "science" mean in this context? e.g. I'd worry about using a strict magnitude or star/galaxy cut for defining "science worthy" sources. I'm not sure I have a better name for it, though.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -97,3 +103,301 @@ def _isBad(self, source):
...
+        if self.config.doFluxLimit:
+            selected &= self.config.fluxLimit.apply(catalog)
+        if self.config.doFlags:
+            selected &= self.config.flags.apply(catalog)
+        if self.config.doUnresolved:
+            selected &= self.config.unresolved.apply(catalog)
+
+        self.log.info("Selected %d/%d sources", selected.sum(), len(catalog))
+
+        result = type(catalog)(catalog.table)
```
parejkoj: What does this line do?

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+import lsst.meas.algorithms
+import lsst.utils.tests
+
+from lsst.pipe.base import Struct
+
+
+class ReserveSourcesTaskTest(lsst.utils.tests.TestCase):
+    """TestCase for the ReserveSourcesTask"""
+    def setUp(self):
+        self.num = 100  # Number of sources
```
parejkoj: You should also set `self.longMessage` in setUp, to enable more clear output for failed tests.

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+        """Test source reservation with a Catalog
+
+        We test multiple reservation fractions.
+        """
+        for fraction in (0.0, 0.1, 0.2, 0.5):
+            data = self.construct("this_table_is", fraction)
+            self.check(data.catalog, data.task, data.key, fraction)
+
+    def testSources(self):
+        """Test source reservation with a list of sources"""
```
parejkoj: Do we really need to support lists?

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+        # Second run with different seed
+        for ss in sources:
+            ss.set(key, False)
+        results2 = task.run(sources, expId=2)
+        flagged2 = np.array([ss.get(key) for ss in sources])
+        self.assertEqual(flagged1.sum(), numExpect)
+        np.testing.assert_array_equal(results2.reserved, flagged2)
+        np.testing.assert_array_equal(results2.use, np.logical_not(flagged2))
+        if numExpect > 0:
+            self.assertFalse(np.all(flagged1 == flagged2))  # Pretty unlikely they match given different seeds
```
parejkoj: You can put your comments in the assert failure message (that includes `fraction`). (also below)

Though, in this particular case, I'm a bit uncomfortable about "pretty unlikely" in a unittest; that seems to be screaming out for a failure on a new compiler/OS.

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+        catalog.reserve(self.num)
+        for _ in range(self.num):
+            catalog.addNew()
+        return Struct(catalog=catalog, task=task, key=key)
+
+    def check(self, sources, task, key, fraction):
+        """Check that source reservation is working
+
+        We test that the source reservation works, that it depends
+        on the RNG seed and that thigns behave as we expect when there's
```
parejkoj: "things"

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+            Key to the flag column.
+        fraction : `float`
+            Fraction of sources to reserve.
+        """
+        numExpect = fraction*len(sources)
+
+        # No prior
+        results1 = task.run(sources, expId=1)
+        flagged1 = np.array([ss.get(key) for ss in sources])
+        self.assertEqual(flagged1.sum(), numExpect)
```
parejkoj: For this and the other asserts, define a message like `"fraction=%s, key=%s"%(fraction,key)`, to simplify debugging (`longMessage=True` above will print this in addition to the usual message),

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+        catalog : `lsst.afw.table.Catalog` or `list` of `lsst.afw.table.Record`
+            List of sources.
+        task : `lsst.meas.algorithms.ReserveSourcesTask`
+            Task to do the reservations.
+        key : `lsst.afw.table.Key`
+            Key to the flag column.
+        fraction : `float`
+            Fraction of sources to reserve.
+        """
+        numExpect = fraction*len(sources)
```
parejkoj: Is there any rounding here to worry about? Should this be cast to `int`?

python/lsst/meas/algorithms/reserveSourcesTask.py
```diff
@@ -0,0 +1,183 @@
...
+            that are flagged `True` in this array.
+        expId : `int`
+            Exposure identifier; used for seeding the random number generator.
+
+        Return struct contents
+        ----------------------
+        reserved : `numpy.ndarray` of type `bool`
+            Sources to be reserved are flagged `True` in this array.
+        use : `numpy.ndarray` of type `bool`
+            Sources the user should use in analysis.
```
parejkoj: The distinction between `reserved` and `use` is not that clear here. What "analysis"?

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+
+        # No prior
+        results1 = task.run(sources, expId=1)
+        flagged1 = np.array([ss.get(key) for ss in sources])
+        self.assertEqual(flagged1.sum(), numExpect)
+        np.testing.assert_array_equal(results1.reserved, flagged1)
+        np.testing.assert_array_equal(results1.use, np.logical_not(flagged1))
+
+        # Second run with different seed
+        for ss in sources:
```
parejkoj: Note that you're clearing the selected flag (I think?).

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+        self.assertEqual(results3.use.sum(), (1.0 - fraction)*prior.sum())
+        self.assertEqual(results3.reserved.sum(), fraction*prior.sum())
+        np.testing.assert_array_equal(results3.use, prior & ~flagged3)  # Actual definition of 'use'
+
+    def testCatalog(self):
+        """Test source reservation with a Catalog
+
+        We test multiple reservation fractions.
+        """
+        for fraction in (0.0, 0.1, 0.2, 0.5):
```
parejkoj: Why not `for fraction in np.linspace(0,1,25)` to catch a variety of different fractions?

tests/test_reserveSourcesTask.py
```diff
@@ -0,0 +1,154 @@
...
+            ss.set(key, False)
+        prior = np.arange(0, self.num, 1, dtype=int) % 2 == 0
+        results3 = task.run(sources, prior, expId=1)
+        flagged3 = np.array([ss.get(key) for ss in sources])
+        self.assertEqual(flagged3.sum(), fraction*prior.sum())
+        if numExpect > 0:
+            self.assertFalse(np.all(flagged1 == flagged3))  # Flagging should change, despite same seed
+        np.testing.assert_array_equal(results3.reserved, flagged3)
+        self.assertEqual((results3.use & flagged3).sum(), 0)  # No sources should be both reserved and used
+        self.assertEqual(results3.use.sum(), (1.0 - fraction)*prior.sum())
```
parejkoj: rounding error possible here?

tests/test_sourceSelector.py
```diff
@@ -0,0 +1,227 @@
...
+
+import lsst.afw.table
+import lsst.meas.algorithms
+import lsst.utils.tests
+
+from lsst.meas.algorithms import ScienceSourceSelectorTask, ReferenceSourceSelectorTask, ColorLimit
+
+
+
+class SourceSelectorTester(object):
```
parejkoj: Two lines before class.

Also should have a docstring to note that this is not a test, but a test baseclass.

tests/test_sourceSelector.py
```diff
@@ -0,0 +1,227 @@
...
+        # Works with no maximum set?
+        self.config.colorLimits["test"].maximum = None
+        self.config.colorLimits["test"].minimum = minimum
+        self.check((color > minimum).tolist())
+
+        # Multiple limits
+        self.config.colorLimits = {"test": ColorLimit(primary="flux", secondary="other_flux", minimum=minimum),
+                                   "other": ColorLimit(primary="flux", secondary="other_flux",
+                                                       maximum=maximum)}
+        self.check(((color > minimum) & (color < maximum)).tolist())
```
parejkoj: Do your minimum and maximum result in non-mutually-exclusive ranges here? I think so, but it might be worth checking.

python/lsst/meas/algorithms/reserveSourcesTask.py
```diff
@@ -0,0 +1,183 @@
...
+        full = np.zeros(len(prior), dtype=bool)
+        full[prior] = selection
+        return full
+
+    def markSources(self, sources, selection):
+        """Mark sources in a list or catalog
+
+        This requires iterating through the list and setting the flag in
+        each source individually. Even if the `sources` is a `Catalog`
+        with contiguous records, it's not possible to set a boolean column
```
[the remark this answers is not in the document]
PaulPrice: Thanks for the ticket; included.

Matches come in `list`s rather than `Catalog`s, so one could imagine passing in a `[mm.first for mm in matches]`.

python/lsst/meas/algorithms/sourceSelector.py
PaulPrice: Maybe they belong in their own file, but I don't know what to call that file except `sourceSelector.py`!

I've a mild preference for not automatically adding `Config` to the end of the names: it's implemented as a `Config` (which is useful for configuring them) but their principal purpose is as a limit. If you have strong feelings, let me know and we'll hash it out.

python/lsst/meas/algorithms/sourceSelector.py
PaulPrice: Added mention of Jansky. I think questions about the specific implementation are best answered by the (simple) code.

python/lsst/meas/algorithms/sourceSelector.py
PaulPrice: That's what the "Returns" section already says, so I think that's redundant.

python/lsst/meas/algorithms/sourceSelector.py
PaulPrice: Added an explanation of "science":

>    By "science" sources, we mean sources that are on images that we
>    are processing, as opposed to sources from reference catalogs.

python/lsst/meas/algorithms/sourceSelector.py
PaulPrice: `type(catalog)` gives the catalog class, then we're calling it: it's a constructor.

Added a comment to these lines.

tests/test_reserveSourcesTask.py
PaulPrice: Oh, if all I have to do is the above then that's easy.

tests/test_reserveSourcesTask.py
PaulPrice: I don't know how else to test that different seeds produce different selections. I don't think it's as big a problem as you think, as the seeds are fixed --- we're not allowing it to run *completely* randomly. However, there is a small, non-zero probability that two seeds produce the same selection. For the worst case, (selecting 10 out of 100), I believe that probability is one in 17,310,309,456,440, i.e., "pretty unlikely".

tests/test_reserveSourcesTask.py
PaulPrice: The choice of fractions is so that the multiplications produce whole numbers, and I thought doing 3 non-zero fractions was sufficient.

tests/test_reserveSourcesTask.py
PaulPrice: Yes, because matches are lists.

python/lsst/meas/algorithms/sourceSelector.py
parejkoj: In that case, I wonder if the we should instead move the existing `BaseSourceSelector` into its own eponymous file then?

I don't have a strong feeling about appending `Config`, but the names as they are just feel short somehow.

tests/test_reserveSourcesTask.py
parejkoj: But I worry that you aren't testing the non-int multiplication case, which the code should be able to handle, no?

PR #95, opened 2017-11-23
DM-12798: Add source selection by signal-to-noise ratio

PaulPrice:


python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -299,26 +355,25 @@ def apply(self, catalog):
...
     fluxLimit = pexConfig.ConfigField(dtype=FluxLimit, doc="Flux limit to apply")
     flags = pexConfig.ConfigField(dtype=RequireFlags, doc="Flags to require")
     unresolved = pexConfig.ConfigField(dtype=RequireUnresolved, doc="Star/galaxy separation to apply")
+    signalToNoise = pexConfig.ConfigField(dtype=SignalToNoiseLimit, doc="Signal-to-noise limit to apply")
 
     def setDefaults(self):
         pexConfig.Config.setDefaults(self)
         self.flags.bad = ["base_PixelFlags_flag_edge", "base_PixelFlags_flag_interpolated",
                           "base_PixelFlags_flag_saturated"]
+        self.signalToNoise.fluxField = "base_PsfFlux_flux"
```
laurenam: Do you also need to default the errField (since having them associated is not assumed?)

tests/test_sourceSelector.py
```diff
@@ -82,6 +83,23 @@ def testFlags(self):
...
+        low.set("other_fluxSigma", 1.0)
+        good = self.catalog.addNew()
+        good.set("other_flux", 1.0)
+        good.set("other_fluxSigma", 0.1)
+        high = self.catalog.addNew()
+        high.set("other_flux", 1.0)
+        high.set("other_fluxSigma", 0.001)
+        self.config.doSignalToNoise = True
+        self.config.signalToNoise.fluxField = "other_flux"
+        self.config.signalToNoise.errField = "other_fluxSigma"
```
laurenam: I think @parejkoj is trying to homogenize the use of err vs. Sigma in the stack (RFC-333).

parejkoj: Sadly, none of that has been implemented yet, so `fluxSigma` will remain.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -225,10 +250,44 @@ def apply(self, catalog):
...
+
+        Parameters
+        ----------
+        catalog : `lsst.afw.table.SourceCatalog`
+            Source to which to apply limits.
+
+        Returns
+        -------
+        selected : `numpy.ndarray`
+            Boolean array indicating whether each source is selected
```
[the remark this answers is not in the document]
PaulPrice: The flags can be handled by the `RequireFlags` selection, so no need to include them here.

PR #96, opened 2017-12-12
DM-12943: Add utilities for detection when noise is correlated.

TallJimbo:


src/CorrelatedNoiseDetection.cc
```diff
@@ -0,0 +1,192 @@
...
+    int const height = mi.getHeight();
+    // iterate over pixels in the MaskedImage, skipping any that meet our bad mask criteria
+    for (int y1 = 0; y1 < height; ++y1) {
+        int const y2a = std::max(0, y1 - radius);
+        int const y2b = std::min(height, y1 + radius + 1);
+        for (int x1 = 0; x1 < width; ++x1) {
+            if ((*mi.getMask())(x1, y1) & badBitMask) {
+                continue;
+            }
+            float z = (*mi.getImage())(x1, y1) / (*mi.getVariance())(x1, y1);
```
rearmstr: Don't you need to subtract the mean here and below from the image values?

src/CorrelatedNoiseDetection.cc
```diff
@@ -0,0 +1,192 @@
...
+            int const x2b = std::min(height, x1 + radius + 1);
+            for (int y2 = y2a; y2 < y2b; ++y2) {
+                auto miIter = mi.row_begin(y2) + x2a;
+                auto outIter = result.row_begin(radius + y2 - y1) + radius + x2a - x1;
+                auto countIter = count.row_begin(radius + y2 - y1) + radius + x2a - x1;
+                for (int x2 = x2a; x2 < x2b; ++x2, ++miIter, ++outIter, ++countIter) {
+                    if (miIter.mask() & badBitMask) {
+                        continue;
+                    }
+                    *outIter += z*miIter.image();
```
rearmstr: Why are you dividing the variance value from only the first pixel?  Shouldn't be something like ```sqrt(var(x1,y1))*sqrt(var(x2,y2))```?

tests/test_correlatedNoiseDetection.py
```diff
@@ -0,0 +1,103 @@
...
+try:
+    type(display)
+except NameError:
+    display = False
+
+
+class CorrelatedNoiseDetectionTestCase(lsst.utils.tests.TestCase):
+    """Test utilities for dealing with correlated noise when running detection."""
+
+    def testMeasureCorrelatationKernel(self):
```
rearmstr: Your test here doesn't use correlated values.  I'm wondering if this is a useful test?  I could imagine getting the correlation values correct for this simple case, but missing a bug that would get the correlation wrong when there actually is correlation.

src/CorrelatedNoiseDetection.cc
TallJimbo: My thinking was that because this should only be running noise pixels, we could assume the mean was zero.  I suppose I should at least add a comment noting that we should check that assumption if we ever use this code in a pipeline setting.

src/CorrelatedNoiseDetection.cc
TallJimbo: I think you're right - I was originally thinking of this as part of the stationarity assumption, but I don't need to assume the variance is stationary to assume that the correlations are.

tests/test_correlatedNoiseDetection.py
TallJimbo: Yup, this was just laziness, because I didn't want to work out how to add correlated stationary noise with NumPy.  If you happen to have a recipe for that on hand, I'd love to have it.

src/CorrelatedNoiseDetection.cc
rearmstr: I found that there were many regions where this was not the case in the HSC wide data.

rearmstr: Of course, they may just get filtered out when taking the median over a tract.

TallJimbo: Just started thinking about fixing this, and I realized that I don't understand what you're proposing - the "mean" we need to subtract is a correction to the background estimate, right?  How do we estimate that?

tests/test_correlatedNoiseDetection.py
TallJimbo: I gave this a shot, and unfortunately the naive way to do it with NumPy (constructing a full covariance matrix) is totally impractical - if we want enough pixels in the noise image to be able to estimate the kernel, the covariance matrix is too big to even allocate.  I could imagine working around that with sparse matrices, and I'm sure there's a way to do this with an FFT, but at this point it's not looking like it's worth the effort.

src/CorrelatedNoiseDetection.cc
rearmstr: Yes, depending on how large the image is the background estimate may be off.  Can't you just compute the mean of the image and subtract that off?

tests/test_correlatedNoiseDetection.py
rearmstr: Yes, FFT is the way to go.  I could potentially give this a shot today.  It is conceptually simple, but it always takes me a while to get all the factors correct.

TallJimbo: That'd be great.  I'm not in any hurry to merge this, so please don't feel like you need to get it done today.

Would it make sense to just make that a separate ticket (maybe to add more reusable functionality to `afw::math::Random`), and block the merge of this ticket on that one?

src/CorrelatedNoiseDetection.cc
TallJimbo: Yup, if you think just subtracting off the image mean would be an improvement (I was imagining something more local), that's definitely something I can do.

tests/test_correlatedNoiseDetection.py
rearmstr: Yeah, that sounds like a good idea.

TallJimbo: https://jira.lsstcorp.org/browse/DM-13044

PR #97, opened 2017-12-12
DM-10765: Replace existing WCS classes with SkyWcs

r-owen:


r-owen:
Incorrect pull request, as @laurenam pointed out.

PR #98, opened 2017-12-19
DM-13055 Reject NaN centroid sigmas

parejkoj:


python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -122,12 +124,16 @@ def _hasCentroid_vector(self, sourceCat):
...
+            & np.isfinite(sourceCat.get(self.centroidYSigmaKey)) \
             & ~sourceCat.get(self.centroidFlagKey)
 
     def _hasCentroid(self, source):
         """Return True if the source has a valid centroid"""
         centroid = source.getCentroid()
-        return np.all(np.isfinite(centroid)) and not source.getCentroidFlag()
+        centroidErr = source.getCentroidErr()
+        return np.all(np.isfinite(centroid)) and np.all(np.isfinite(centroidErr)) \
+            and not source.getCentroidFlag()
```
TallJimbo: `np.all(np.isfinite(centroid))` should always be unnecessary.  If you really want to check, you could `assert` on it - it should be finite even when `source.getCentroidFlag()` is true.

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -122,12 +124,16 @@ def _hasCentroid_vector(self, sourceCat):
         """Return True for each source that has a valid centroid"""
         return np.isfinite(sourceCat.get(self.centroidXKey)) \
             & np.isfinite(sourceCat.get(self.centroidYKey)) \
+            & np.isfinite(sourceCat.get(self.centroidXSigmaKey)) \
+            & np.isfinite(sourceCat.get(self.centroidYSigmaKey)) \
             & ~sourceCat.get(self.centroidFlagKey)
```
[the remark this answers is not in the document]
parejkoj: I've never really had an issue with the binary `&` for `np` expressions, but I can convert to using `np.logical_and`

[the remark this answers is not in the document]
TallJimbo: I actually don't know of anyone having trouble either, so maybe I'm just paranoid.  Just thought I'd bring it up.

parejkoj:
> My only big-picture critique is that these SourceSelectors seem quite rigid, and I think the proximate cause is that not enough of these tests are configurable - perhaps this new test should be, at least, so code that doesn't care about centroid uncertainty will see no change in behavior?

`astrometrySourceSelector` is rigid by design: its job is to select sources that are useful for astrometric measurements, which has a pretty clear definition that includes "reliable centroids". I don't see how "reliable centroid" doesn't include "reliable centroid error".

python/lsst/meas/algorithms/astrometrySourceSelector.py
parejkoj: Ok, I guess I can turn the centroid finiteness checks into asserts, but I'm going to argue in an RFC shortly that the errors should also be finite for unflagged sources.

python/lsst/meas/algorithms/astrometrySourceSelector.py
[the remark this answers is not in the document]
parejkoj: I just tried some tests, and it looks like `np.logical_and.reduce()` and `np.bitwise_and.reduce()` is significantly slower than chained `&`:

```
...
In [27]: aa=np.random.random(1000000) > .5

In [28]: %timeit np.logical_and.reduce((xx,yy,zz,aa))
669 µs ± 47 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)

In [29]: %timeit xx & yy & zz & aa
182 µs ± 1.82 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

In [35]: %timeit np.bitwise_and.reduce((xx,yy,zz,aa))
651 µs ± 9.46 µs per loop (mean ± std. dev. of 7 runs, 1000 loops each)
```

So, I'd rather go with `&`, even though your concern isn't crazy.

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -120,14 +122,22 @@ def _isMultiple(self, source):
...
         """Return True for each source that has a valid centroid"""
-        return np.isfinite(sourceCat.get(self.centroidXKey)) \
-            & np.isfinite(sourceCat.get(self.centroidYKey)) \
+        def centroid_nonfinite_but_not_flagged():
+            """Return True for sources with non-flagged, non-finite centroids."""
+            return ((~np.isfinite(sourceCat.get(self.centroidXKey)) |
+                    ~np.isfinite(sourceCat.get(self.centroidYKey))) &
+                    ~sourceCat.get(self.centroidFlagKey))
+        assert ~centroid_nonfinite_but_not_flagged().any(), \
+            "Centroids not finite for %d unflagged sources." % (centroid_nonfinite_but_not_flagged().sum())
```
TallJimbo: You can actually just do
```
assert np.isfinite(sourceCat.get(self.centroidXKey)).all() and np.isfinite(sourceCat.get(self.centroidYKey)).all()
```
In other words, centroid values should never be NaN, *even when the flag is set*.  And that's why I wanted to make that an assert: if a centroid algorithm disobeys that rule, it's a serious bug.

TallJimbo: Also, I'm obligated to remind you that snake case isn't allowed :(

python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -120,14 +122,22 @@ def _isMultiple(self, source):
...
+        return np.isfinite(sourceCat.get(self.centroidXSigmaKey)) \
+            & np.isfinite(sourceCat.get(self.centroidYSigmaKey)) \
             & ~sourceCat.get(self.centroidFlagKey)
 
     def _hasCentroid(self, source):
         """Return True if the source has a valid centroid"""
-        centroid = source.getCentroid()
-        return np.all(np.isfinite(centroid)) and not source.getCentroidFlag()
+        assert np.all(np.isfinite(source.getCentroid())) and not source.getCentroidFlag(), \
+            'Centroid not finite for unflagged source: %s'%source
```
TallJimbo: Same as above: no need to check for centroid flag in this assert.

tests/test_astrometrySourceSelector.py
```diff
@@ -96,10 +99,27 @@ def testSelectSources_bad(self):
...
+        self.src[0].set('slot_Centroid_y', np.nan)
+        with self.assertRaises(AssertionError):
+            self.sourceSelector.selectSources(self.src)
+
+    def testSelectSources_bad_centroid_Sigma(self):
+        """Reject sources with NaN centroid_[xy]Sigma.
+        Currently, there is no guarantee that a valid centroid_flag implies
+        a finite variance."""
+        add_good_source(self.src, 1)
+        add_good_source(self.src, 2)
```
TallJimbo: More snake case.  Looks like a lot of it predates these changes; maybe file a ticket to fix that later?

PR #103, opened 2018-02-06
DM-13493: SourceSelector should accept saturated sources.

jdswinbank:
Per discussion on Slack of 2018-02-05:

> `interpolated`... now effectively means "this pixel was interpolated in one
> of the inputs and that's okay because we think the interpolation was good".

(Jim Bosch, in #subaru-hsc)

We should therefore not reject sources flagged as such in the source selector.


parejkoj:
Looks good!

PaulPrice:
The commit message is misleading --- should be "Do not reject **interpolated** sources..."

PR #105, opened 2018-02-22
DM-13617: detection: fix restoration of tempWideBackground

PaulPrice:
We weren't restoring the original image because we didn't have a copy
of it.

### DIFF
--- a/python/lsst/meas/algorithms/detection.py
+++ b/python/lsst/meas/algorithms/detection.py
@@ -899,7 +899,7 @@ into your debug.py file and run measAlgTasks.py with the \c --debug flag.
         doTempWideBackground = self.config.doTempWideBackground
         if doTempWideBackground:
             self.log.info("Applying temporary wide background subtraction")
-            original = exposure.maskedImage.image.array[:]
+            original = exposure.maskedImage.image.array[:].copy()
             self.tempWideBackground.run(exposure).background
             # Remove NO_DATA regions (e.g., edge of the field-of-view); these can cause detections after
             # subtraction because of extrapolation of the background model into areas with no constraints.
--- a/tests/test_detection.py
+++ b/tests/test_detection.py
@@ -24,9 +24,11 @@
 from __future__ import absolute_import, division, print_function
 from builtins import range
 import unittest
+import numpy as np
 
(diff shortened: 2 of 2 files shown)


python/lsst/meas/algorithms/detection.py
```diff
@@ -899,7 +899,7 @@ def tempWideBackgroundContext(self, exposure):
         doTempWideBackground = self.config.doTempWideBackground
         if doTempWideBackground:
             self.log.info("Applying temporary wide background subtraction")
-            original = exposure.maskedImage.image.array[:]
+            original = exposure.maskedImage.image.array[:].copy()
```
TallJimbo: Could simplify this to
```
original = exposure.image.array.copy()
```

tests/test_detection.py
```diff
@@ -98,6 +100,36 @@ def makeCoordList(self, bbox, numX, numY, minCounts, maxCounts, sigma):
...
         return coordList
 
+    def testTempBackgrounds(self):
+        """Test that the temporary backgrounds we remove are properly restored"""
+        bbox = afwGeom.Box2I(afwGeom.Point2I(12345, 67890), afwGeom.Extent2I(128, 127))
+        original = afwImage.ExposureF(bbox)
+        rng = np.random.RandomState(123)
+        original.maskedImage.image.array[:] = rng.normal(size=original.maskedImage.image.array.shape)
+        original.maskedImage.mask.set(0)
+        original.maskedImage.variance.set(1.0)
```
TallJimbo: Could similarly make all of these `original.(mask|image|variance)...`.

tests/test_detection.py
```diff
@@ -98,6 +100,36 @@ def makeCoordList(self, bbox, numX, numY, minCounts, maxCounts, sigma):
...
+
+        def checkExposure(original, doTempLocalBackground, doTempWideBackground):
+            config = SourceDetectionTask.ConfigClass()
+            config.reEstimateBackground = False
+            config.thresholdType = "pixel_stdev"
+            config.doTempLocalBackground = False
+            config.doTempWideBackground = False
```
TallJimbo: Shouldn't these be set to the function arguments?

PR #111, opened 2018-03-31
DM-8828: Support proper motions in reference catalogs

PaulPrice:


python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
```diff
@@ -136,3 +140,50 @@ def _trim_to_circle(self, catalog_shard, ctrCoord, radius):
             if record.getCoord().separation(ctrCoord) < radius:
                 temp_cat.append(record)
         return temp_cat
+
+
+class LoadPanstarrsObjectsTask(LoadIndexedReferenceObjectsTask):
+    """Reference catalog loader for ps1_pv3_3pi_20170110 catalog
```
parejkoj: Period at the end.

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
```diff
@@ -136,3 +140,50 @@ def _trim_to_circle(self, catalog_shard, ctrCoord, radius):
...
+
+    This catalog is in the "LSST format" suitable for use with
+    the `LoadIndexedReferenceObjectsTask`, but has a slightly
+    different format for the proper motion columns than was
+    eventually adopted for the default of the LSST format:
+
+    - The proper motions are in units of arcsec/yr instead of mas/yr.
+    - The epoch is in seconds since the Unix epoch instead of MJD.
```
parejkoj: I'm concerned about the creation of a ReferenceObjects class designed for exactly one catalog. I worry that we'll end up with similar `LoadGaiaDR1FirstCutTask`, etc. We should probably at least call it `LoadPanstarrs20170110ObjectsTask` or something, to make it clear that it's for "one very specific PS catalog".

Could we instead write something to modify those catalogs to match the above format instead?

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
```diff
@@ -136,3 +140,50 @@ def _trim_to_circle(self, catalog_shard, ctrCoord, radius):
...
+        -------
+        struct : `lsst.pipe.base.Struct`
+            Struct of proper motion scales, containing:
+
+            - ``values`` : values of the proper motion scale (may be
+                either a scalar `float` or an `numpy.ndarray`)
+            - ``units`` : units of proper motion (`lsst.afw.geom.Angle`)
+        """
+        epochUnix = DateTime(float(epoch), DateTime.TAI).nsecs()/1.0e9  # Sec since Unix epoch
+        timeDiff = (epochUnix - catalog["epoch"])/(3600*24*365.25)  # Time difference, years
```
parejkoj: This feels like an good use case for `astropy.time`. Or do we have an appropriate time model in DateTime? `365.25` isn't technically precise, is it?

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -105,6 +106,11 @@ class LoadReferenceObjectsConfig(pexConfig.Config):
         itemtype=str,
         default={},
     )
+    requireProperMotion = pexConfig.Field(
+        doc="Require that the proper motion correction be successful?",
```
parejkoj: What does "be successful" mean in this context? The effect of this config parameter is not clear to me.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -174,6 +180,14 @@ class LoadReferenceObjectsTask(with_metaclass(abc.ABCMeta, pipeBase.Task)):
     ConfigClass = LoadReferenceObjectsConfig
     _DefaultName = "LoadReferenceObjects"
 
+    _properMotionColumns = pipeBase.Struct(
+        ra = "pmRa",
+        dec = "pmDec",
+        raErr = "pmRaErr",
+        decErr = "pmDecErr",
+        epoch = "epoch",
+    )
```
parejkoj: Why define this Struct if we've got a standard schema name for these?

tests/test_htmIndex.py
```diff
@@ -308,6 +340,29 @@ def testDefaultFilterAndFilterMap(self):
...
+
+        # One year difference --> coordinate difference should be self.proper_motion in RA only
+        loader.applyProperMotions(references, self.mjd + 365.25)
+        for orig, ref in zip(original, references):
+            self.assertFloatsAlmostEqual(orig.getCoord().bearingTo(ref.getCoord()).asDegrees(),
+                                         180.0, rtol=1.0e-9)
+            self.assertFloatsAlmostEqual(orig.getCoord().separation(ref.getCoord()).asArcseconds(),
+                                         self.proper_motion, rtol=1.0e-12)
+
+        # XXXX check errors are modified
```
parejkoj: This comment suggests something was left unfinished?

tests/test_htmIndex.py
```diff
@@ -53,14 +54,15 @@ def make_coord(ra, dec):
 
 
 class HtmIndexTestCase(lsst.utils.tests.TestCase):
-
-    @staticmethod
-    def make_sky_catalog(out_path, size=1000):
+    @classmethod
```
parejkoj: I don't see why this needs to be a classmethod. It looks like all of the setup happens in `setUpClass`, which is worrying because what happens if any of those values are changed by a test? I think the `setUpClass` should become just `setUp` and then this can just be a regular method.

tests/test_htmIndex.py
```diff
@@ -70,26 +72,35 @@ def make_sky_catalog(out_path, size=1000):
         is_variable = np.random.randint(2, size=size)
         extra_col1 = np.random.normal(size=size)
         extra_col2 = np.random.normal(1000., 100., size=size)
+        pm_ra = -np.ones(size)*cls.proper_motion  # arcsec/year
+        pm_dec = np.zeros(size)  # arcsec/year
+        pm_ra_err = np.ones(size)*0.001  # arcsec/year
+        pm_dec_err = np.ones(size)*0.001  # arcsec/year
```
parejkoj: More "I wish we could use astropy.Quantity", but I don't know that there's a feasible way to do so.

tests/test_htmIndex.py
```diff
@@ -198,6 +219,14 @@ def testIngest(self):
...
         # test with multiple files and correct config
         default_config.mag_err_column_map = {'a': 'a_err', 'b': 'b_err'}
+        default_config.ra_err_name = "ra_err"
+        default_config.dec_err_name = "dec_err"
+        default_config.pm_ra_name = "pm_ra"
+        default_config.pm_dec_name = "pm_dec"
+        default_config.epoch_name = "unixtime"
+        default_config.pm_ra_err_name = "pm_ra_err"
+        default_config.pm_dec_err_name = "pm_dec_err"
+        default_config.pm_scale = 1.0e-3  # arcsec/yr --> mas/yr
```
parejkoj: There's a `config` that gets created and thrown away in `setUpClass`. Why not make it an instance variable and use it here?

tests/test_htmIndex.py
```diff
@@ -308,6 +340,29 @@ def testDefaultFilterAndFilterMap(self):
...
+
+        # Zero epoch change --> no proper motion correction (except minor numerical effects)
+        loader.applyProperMotions(references, self.mjd)
+        self.assertFloatsAlmostEqual(references["coord_ra"], original["coord_ra"], rtol=1.0e-14)
+        self.assertFloatsAlmostEqual(references["coord_dec"], original["coord_dec"], rtol=1.0e-14)
+        self.assertFloatsEqual(references["coord_ra_err"], original["coord_ra_err"])
+        self.assertFloatsEqual(references["coord_dec_err"], original["coord_dec_err"])
+
+        # One year difference --> coordinate difference should be self.proper_motion in RA only
+        loader.applyProperMotions(references, self.mjd + 365.25)
```
parejkoj: Another 365.25 hard coded here...

tests/test_htmIndex.py
```diff
@@ -308,6 +340,29 @@ def testDefaultFilterAndFilterMap(self):
...
+        loader = LoadIndexedReferenceObjectsTask(butler=self.test_butler)
+        references = loader.loadSkyCircle(center, self.search_radius, filterName='a').refCat
+        original = references.copy(True)
+
+        # Zero epoch change --> no proper motion correction (except minor numerical effects)
+        loader.applyProperMotions(references, self.mjd)
+        self.assertFloatsAlmostEqual(references["coord_ra"], original["coord_ra"], rtol=1.0e-14)
+        self.assertFloatsAlmostEqual(references["coord_dec"], original["coord_dec"], rtol=1.0e-14)
+        self.assertFloatsEqual(references["coord_ra_err"], original["coord_ra_err"])
+        self.assertFloatsEqual(references["coord_dec_err"], original["coord_dec_err"])
```
parejkoj: Good set of tests here (no correction means no change)!

tests/test_htmIndex.py
```diff
@@ -70,26 +72,35 @@ def make_sky_catalog(out_path, size=1000):
         is_variable = np.random.randint(2, size=size)
         extra_col1 = np.random.normal(size=size)
         extra_col2 = np.random.normal(1000., 100., size=size)
+        pm_ra = -np.ones(size)*cls.proper_motion  # arcsec/year
+        pm_dec = np.zeros(size)  # arcsec/year
```
parejkoj: Worth a comment here to make it clear that the intention is for the proper motion to be explicitly in ra only.

tests/test_htmIndex.py
```diff
@@ -308,6 +340,29 @@ def testDefaultFilterAndFilterMap(self):
...
+        self.assertFloatsEqual(references["coord_ra_err"], original["coord_ra_err"])
+        self.assertFloatsEqual(references["coord_dec_err"], original["coord_dec_err"])
+
+        # One year difference --> coordinate difference should be self.proper_motion in RA only
+        loader.applyProperMotions(references, self.mjd + 365.25)
+        for orig, ref in zip(original, references):
+            self.assertFloatsAlmostEqual(orig.getCoord().bearingTo(ref.getCoord()).asDegrees(),
+                                         180.0, rtol=1.0e-9)
+            self.assertFloatsAlmostEqual(orig.getCoord().separation(ref.getCoord()).asArcseconds(),
+                                         self.proper_motion, rtol=1.0e-12)
```
parejkoj: I'm somewhat surprised the tolerance has to be that high for both direction and distance.

I was about to say you should use `assertSpherePointListAlmostEqual`, until I thought about the nature of the test (move X along RA only). So, maybe a comment about why the above test method wouldn't be ideal?

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
wmwv: 3600 sec/hour *24 hours/day * 365.25 days is indeed defined to be what astronomers mean when they say "year" (where the sec is an SI second).  And thus if an astronomer quotes to you a proper motion/year, this is the year they were supposed to have used.

https://www.iau.org/publications/proceedings_rules/units/

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -434,7 +457,91 @@ def joinMatchListWithCatalog(self, matchCat, sourceCat):
...
+        the units of proper motion.
+
+        Parameters
+        ----------
+        catalog : `lsst.afw.table.SimpleCatalog`
+            Reference catalog. By default, this contains a column
+            ``epoch``, which is the mean epoch of the source as a
+            Modified Julian Date (MJD) in TAI.
+        epoch : `float`
+            Epoch to which to move objects (TAI MJD).
```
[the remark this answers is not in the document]
wmwv: JD and in turn MJD are defined in times of TAI.  One is supposed to explicitly note if one is using an JD or MJD calculated based on a different system.

https://www.iers.org/IERS/EN/Science/Recommendations/resolutionB1.html

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
wmwv: I'm fine with having a `LoadPanstarrsObjectsTask`, but it seems like this is a good opportunity to write the accepted LSST one as well (although that can be a separate ticket).

In this future ticket, one might imagine making `LoadPanstarrsObjectsTask` and `LoadLSSTObjectsTask` (naming?) inherit from something slightly more general.

PaulPrice: Please, everyone read the docstring, which explains the motivation behind this class: it only exists to provide backward compatibility.

`LoadLsstObjectsTask` won't be necessary because we will create catalogs with the correct schema.

wmwv: @PaulPrice Ah, yes.  I understand.  Thank you for the clarification.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -238,6 +314,32 @@ def _set_mags(self, record, row, key_map):
...
+        """
+        if self.config.pm_ra_name is None:  # IngestIndexedReferenceConfig.validate ensures all or none
+            return
+        record.set(key_map["pmRa"], row[self.config.pm_ra_name]*self.config.pm_scale)
+        record.set(key_map["pmDec"], row[self.config.pm_dec_name]*self.config.pm_scale)
+        record.set(key_map["epoch"], self.epoch_poly(row[self.config.epoch_name]))
+        if self.config.pm_ra_err_name is None or self.config.pm_dec_err_name is None:
+            return
+        record.set(key_map["pmRaErr"], row[self.config.pm_ra_err_name]*self.config.pm_scale)
+        record.set(key_map["pmDecErr"], row[self.config.pm_dec_err_name]*self.config.pm_scale)
```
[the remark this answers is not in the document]
PaulPrice: I plan to use whatever has been approved in the RFC. If you want something different, then you need to RFC it.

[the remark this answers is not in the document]
r-owen: Ouch. In my opinion having an inconsistent naming convention for fields in the schema will be a disaster. I'd much rather hold off merging this and deal with it as a new RFC if necessary. Is `coord_ra_err` mandated in an RFC?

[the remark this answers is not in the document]
r-owen: RFC 271 doesn't say anything about the names for coord errors, so I'm picking `coordRaErr` and `coordDecErr` to match the proper motion error names.

The down side to any of these name choices is that we cannot save them as pair type (unlike `coord` which is a `Coord` field that internally has two scalar fiels `coord_ra` and `coord_dec`). However, the only obvious pair type is `Point` and the scalar fields would have to have suffixes `_x` and `_y` which is hardly ideal for RA and Dec.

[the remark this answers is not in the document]
TallJimbo: `afw::table::CovarianceMatrixKey` is intended to support storing any kind of error, including diagonal covariances.  It uses the old conventions now, but if you use it, it'll be updated to RFC-271 automatically when `CovarianceMatrixKey` itself is updated.

The names should be `coord_raErr`/`coord_decErr`, and that's what `CovarianceMatrixKey` will produce.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -131,8 +177,15 @@ def validate(self):
         if not (self.ra_name and self.dec_name and self.mag_column_list):
             raise ValueError("ra_name and dec_name and at least one entry in mag_column_list must be" +
                              " supplied.")
+        if (self.pm_ra_err_name is not None) != (self.pm_dec_err_name is not None):  # XOR
+            raise ValueError("Only one of ra_err_name and dec_err_name has been specified")
         if len(self.mag_err_column_map) > 0 and not len(self.mag_column_list) == len(self.mag_err_column_map):
             raise ValueError("If magnitude errors are provided, all magnitudes must have an error column")
+        if (sum(1 for col in ("pm_ra_name", "pm_dec_name", "epoch_name") if getattr(self, col) is not None)
+            not in (0, 3)):
```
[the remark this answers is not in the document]
r-owen: Agreed. However, we need this functionality for the pm error fields as well, so I'll make a small function to handle it based on your suggested logic above.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -120,6 +130,42 @@ class IngestIndexedReferenceConfig(pexConfig.Config):
...
+    )
+    epoch_name = pexConfig.Field(
+        dtype=str,
+        doc="Name of epoch column",
+        optional=True,
+    )
+    epoch_poly = pexConfig.ListField(
+        dtype=float,
+        doc="Coefficients of ordinary polynomial to convert epoch values to MJD TAI, "
+            "from zeroth to highest order. To convert from Unix time, use [40587.0, 1.0/86400].",
```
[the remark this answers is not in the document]
r-owen: For the record: I am changing it as per my suggestion.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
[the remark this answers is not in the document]
TallJimbo: (note edits in my last post - I was confusing RFC-271 with RFC-333)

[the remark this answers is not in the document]
r-owen: @TallJimbo thank you for the suggestion and for re-opening RFC-271 to match RFC-333! I will proceed as if your modification to RFC-271 is accepted, since I will be surprised if it proves controversial.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -435,7 +573,67 @@ def joinMatchListWithCatalog(self, matchCat, sourceCat):
...
+            self.log.warn("Proper motion correction not available from catalog")
+            return
+        if not catalog.isContiguous():
+            raise RuntimeError("Catalog must be contiguous")
+        self.log.debug("Correcting reference catalog for proper motion to %f", epoch)
+        timeDiffsYears = (epoch - catalog["epoch"])/365.25  # Time difference, years
+        coordKey = catalog.table.getCoordKey()
+        # Compute the offset of each object due to proper motion
+        # as components of the arc of a great circle along RA and Dec
+        offsetsRaRad = catalog["pm_ra"]*timeDiffsYears
```
SimonKrughoff: I'd slightly prefer that this be done in a utility library so others can use it if needed.  Ideally it would delegate to a third party package to do the calculations, but if we don't have that, so be it.

tests/test_htmIndex.py
```diff
@@ -165,47 +174,159 @@ def testAgainstPersisted(self):
...
+
+    def testValidateRaDecMag(self):
+        config = self.makeConfig()
+        config.validate()
+
+        for name in ("ra_name", "dec_name", "mag_column_list"):
+            with self.subTest(name=name):
+                config = self.makeConfig()
+                setattr(config, name, None)
+                with self.assertRaises(ValueError):
```
SimonKrughoff: I'm afraid I just don't understand the test here.  Why would it raise?

tests/test_htmIndex.py
```diff
@@ -232,16 +356,18 @@ def testIngest(self):
         config = LoadIndexedReferenceObjectsConfig()
         config.ref_dataset_name = "myrefcat"
         loader = LoadIndexedReferenceObjectsTask(butler=butler, config=config)
-        cat = loader.loadSkyCircle(cent, self.search_radius, filterName='a')
+        cat = loader.loadSkyCircle(cent, self.search_radius, filterName='a').refCat
```
SimonKrughoff: Good catch.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -141,6 +235,14 @@ class IngestIndexedReferenceTask(pipeBase.CmdLineTask):
...
     shards.  In this case each shard contains the entries from the catalog in a single
     HTM trixel
+
+    For producing catalogs this task makes the following assumptions
+    about the input catalogs:
+    - RA, Dec, RA error and Dec error are all in decimal degrees.
+    - Epoch is available in a column, in a format supported by astropy.time.Time.
+    - Off-diagonal covariance terms are not available, e.g.
+        covariance between RA and Dec, or between PM RA and PM Dec.
+        Thus this is not an appropriate task to ingest Gaia catalogs.
```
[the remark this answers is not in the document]
r-owen: I called out Gaia because it is the only catalog I know of that has covariance terms. I'll add an "for example" somewhere in the comment.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -32,6 +35,9 @@
 from lsst.afw.image import fluxFromABMag, fluxErrFromABMagErr
 from .indexerRegistry import IndexerRegistry
 from .readTextCatalogTask import ReadTextCatalogTask
+from .loadReferenceObjects import LoadReferenceObjectsTask
+
+_RAD_PER_DEG = math.pi / 180
```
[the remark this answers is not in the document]
r-owen: Two reasons:
- I prefer to avoid numpy for scalars, for fear of inefficiencies and because it tends to confuse readers (who may think it is operating on array-like data)
- I personally dislike `math.radians` and `np.radians` because I feel their meaning is opaque.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -169,6 +271,7 @@ def __init__(self, *args, **kwargs):
         self.indexer = IndexerRegistry[self.config.dataset_config.indexer.name](
             self.config.dataset_config.indexer.active)
         self.makeSubtask('file_reader')
+        self.radPerMilliarcsec = math.pi/(180*3600*1000)
```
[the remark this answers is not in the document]
r-owen: That's a great point. I will define a new constant next to `_RAD_PER_DEG`

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -163,6 +171,34 @@ class LoadReferenceObjectsTask(pipeBase.Task, metaclass=abc.ABCMeta):
...
+    - pm_raErr: uncertainty in `pm_ra` (rad/year)
+    - pm_decErr: uncertainty in `pm_dec` (rad/year)
+    - pm_ra_dec_Cov: covariance between pm_ra and pm_dec (rad2/year2)
+    - pm_flag: set if proper motion, error or covariance is bad
+
+    - parallax: parallax (rad)
+    - parallaxErr: uncertainty in `parallax` (rad)
+    - parallax_flag: set if parallax value or parallaxErr is bad
+
+    - coord_ra_pm_ra_Cov: covariance between coord_ra and pm_ra (rad2/year)
```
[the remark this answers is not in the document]
r-owen: I said that `IngestIndexReferenceTask` does not support covariances. Reference catalogs can contain them, and if they are present, then they should have the names documented here.

tests/test_htmIndex.py
r-owen: Better yet I'll put the PM at an angle with `cls.proper_motion_amt` and `cls.proper_motion_dir` and make them instances of `lsst.geom.Angle`

tests/test_htmIndex.py
r-owen: The three config items being tested all must have non-None values, so if any value is None then validation should fail.

r-owen:
@parejkoj in response to your main comments:
- Field names: I agree they needed work. For the record we had an RFC and agreed on new field names and handling of invalid data.
- I agree about hard-coded constants and use of astropy quantities. The code still has a few hard-coded constants for angular conversion, but now uses astropy for date and time.
- I agree about `LoadPanstarrsObjectsTask` and was able to get rid of it. The cost is that we will have to generate new PanSTARRs reference catalogs, but there are benefits to this, including fixing a few fields that were all zeros and adding parallax.

python/lsst/meas/algorithms/loadReferenceObjects.py
r-owen: I agree. It means that the data needed to apply proper motion corrections is present. I'll try to improve the help string.

tests/test_htmIndex.py
r-owen: I believe this tolerance was based on writing out rather few digits to to the "original catalog" that is ingested. I have overhauled the code to compare angles and use absolute tolerance. Here is the new code:
```
self.assertAnglesAlmostEqual(orig.getCoord().separation(ref.getCoord()),
                             self.proper_motion_amt, maxDiff=1.0e-6*lsst.geom.arcseconds)
self.assertAnglesAlmostEqual(orig.getCoord().bearingTo(ref.getCoord()),
                             self.proper_motion_dir, maxDiff=1.0e-4*lsst.geom.arcseconds)
```
the separation is quite a bit better than this, but I didn't want to overdo it. Direction is harder to compute accurately, but the allowed error is quite acceptable.

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
PaulPrice: Is that going to be as fast?

tests/test_htmIndex.py
PaulPrice: Shouldn't be using `under_scores` in variable names.

r-owen: Good point. I'll fix it (though I don't think I made up those names; didn't you come up with them?)

tests/test_htmIndex.py
r-owen: Much of the work happens in setupClass because it makes a butler, which is very expensive

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
r-owen: I think it will be fast enough. The time difference per object is an array operation (which is no guarantee, but it is bound to help).

tests/test_htmIndex.py
r-owen: I personally think this particular test is more readable by constructing the config each time, though I will change it to not re-use the same variable name for different configs.

tests/test_htmIndex.py
PaulPrice: I don't think so. Probably they came in with the origin of the "LSST format catalogs", which I think was back when the policy was unclear and thought to be in transition.

tests/test_htmIndex.py
parejkoj: Good case for a `mock`? But not on this ticket, probably.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -223,19 +265,26 @@ def loadPixelBox(self, bbox, wcs, filterName=None, calib=None):
...
-    def loadSkyCircle(self, ctrCoord, radius, filterName=None):
+    def loadSkyCircle(self, ctrCoord, radius, filterName=None, epoch=None):
         """!Load reference objects that overlap a circular sky region
 
         @param[in] ctrCoord  ICRS center of search region (an lsst.geom.SpherePoint)
         @param[in] radius  radius of search region (an lsst.geom.Angle)
         @param[in] filterName  name of filter, or None for the default filter;
             used for flux values in case we have flux limits (which are not yet implemented)
+        @param[in] epoch  Epoch for proper motion and parallax correction
+                          (an astropy.time.Time), or None
```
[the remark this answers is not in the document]
parejkoj: Scratch that: you can use docstring inheritance now, so you can delete the docstrings from the derived class's methods (see discussion on #dm-docs).

python/lsst/meas/algorithms/loadReferenceObjects.py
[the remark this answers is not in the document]
SimonKrughoff: ok

PR #114, opened 2018-04-16
DM-14119: dynamicDetection: make robust against extended emission

PaulPrice:
by applying an explicit background subtraction using the
base_LocalBackground measurement algorithm. Otherwise, extended
low-level background (e.g. around bright stars) can contaminate
our measurements.


python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -120,16 +120,20 @@ def calculateThreshold(self, exposure, seed, sigma=None):
...
+        good = (~catalog["base_PsfFlux_flag"] & ~catalog["base_LocalBackground_flag"] &
+                np.isfinite(fluxes) & np.isfinite(area) & np.isfinite(bg))
 
         if good.sum() < self.config.minNumSources:
             self.log.warn("Insufficient good flux measurements (%d < %d) for dynamic threshold calculation",
                           good.sum(), self.config.minNumSources)
             return Struct(multiplicative=1.0, additive=0.0)
 
-        bgMedian = np.median((fluxes/catalog["base_PsfFlux_area"])[good])
+        bgMedian = np.median((fluxes/area - bg)[good])
```
yalsayyad: What is this quantity exactly? An approx bg-subtracted surface brightness?

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -120,16 +120,20 @@ def calculateThreshold(self, exposure, seed, sigma=None):
...
         if good.sum() < self.config.minNumSources:
             self.log.warn("Insufficient good flux measurements (%d < %d) for dynamic threshold calculation",
                           good.sum(), self.config.minNumSources)
             return Struct(multiplicative=1.0, additive=0.0)
 
-        bgMedian = np.median((fluxes/catalog["base_PsfFlux_area"])[good])
+        bgMedian = np.median((fluxes/area)[good])
 
-        lq, uq = np.percentile(fluxes[good], [25.0, 75.0])
+        lq, uq = np.percentile((fluxes - bg*area)[good], [25.0, 75.0])
```
yalsayyad: How does this change the multiplicative term?

yalsayyad: Oh I get it, you don't want variations in background to increase the estimate of the std-dev?

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -236,7 +236,19 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
...
-            bgLevel = self.calculateThreshold(exposure, seed, sigma=sigma).additive
+            #
+            # But we want to keep any large-scale background (e.g., scattered light from bright stars)
+            # from being selected for sky objects in the calculation, so do another detection pass without
+            # either the local or wide temporary background subtraction; the DETECTED pixels will mark
+            # the area to ignore.
+            originalMask = maskedImage.mask.array.copy()
+            try:
+                convolveResults = self.convolveImage(maskedImage, psf, doSmooth=doSmooth)
+                prelim = self.applyThreshold(middle, maskedImage.getBBox(), self.config.prelimThresholdFactor)
```
yalsayyad: Did you mean `convolveResults.middle`? Where did this middle come from?

yalsayyad: Do you need to reconvolve?

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -236,7 +236,19 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
...
+                self.finalizeFootprints(maskedImage.mask, prelim, sigma, self.config.prelimThresholdFactor)
```
yalsayyad: Which sigma is this?

parejkoj:
I'm worried that this change does not affect any existing tests and also doesn't have any new tests to demonstrate what it does. Given the discussion on slack#subaru-hsc, this is a non-trivial change so we really should have at least one test to prove it's doing what you intend.

RobertLuptonTheGood:
I am very sympathetic with John here; we are short of functional tests for the science code.  I am not sure, however, that this is something that can be added as part of this ticket as opposed to adding to a not-yet-existing scientific regression test.

PaulPrice:
I've re-crafted the commits, and extended the existing test in order to confirm at least some of the new functionality. I've had some difficulty implementing a test that exactly reproduces Ikeda-san's case (incorrect background tweak), but the extended test fails on master and passes with the changes here. I believe the problem is that I'm not accurately matching the complicated geometry from scattered light around the bright star in the HSC camera; but I think that is better addressed in a scientific regression test rather than the simplified unit test framework.

So, the bottom line is that the test is not quite all @parejkoj (and I) hoped for, but nevertheless it is useful and I hope it will prove sufficient.

parejkoj:
At least some test is probably better than no test at all?

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -232,7 +236,20 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
...
+            # But we want to keep any large-scale background (e.g., scattered light from bright stars)
+            # from being selected for sky objects in the calculation, so do another detection pass without
+            # either the local or wide temporary background subtraction; the DETECTED pixels will mark
+            # the area to ignore.
+            originalMask = maskedImage.mask.array.copy()
+            try:
+                self.clearMask(exposure.mask)
+                convolveResults = self.convolveImage(maskedImage, psf, doSmooth=doSmooth)
+                tweakDetResults = self.applyThreshold(convolveResults.middle, maskedImage.getBBox(), factor)
+                self.finalizeFootprints(maskedImage.mask, tweakDetResults, sigma, factor)
```
yalsayyad: Why sigma from line 199?

PR #118, opened 2018-04-27
DM-14102 makePsfCandidates is now its own Task

parejkoj:


mrawls:
I also noticed this is failing Travis, but I'm not sure whether that is meaningful or not since we're still in the process of getting Travis set up across the Stack.

timj:
There is a clash with #119  -- @r-owen enabled travis to test that PR but now travis runs for all pull requests.  #119 seems like a big change so I hope it doesn't conflict with this PR. Not sure when it will be merged.

python/lsst/meas/algorithms/makePsfCandidates.py
```diff
@@ -0,0 +1,141 @@
...
+                    didSetSize = True
+
+                im = psfCandidate.getMaskedImage().getImage()
+            except lsst.pex.exceptions.Exception as err:
+                self.log.warn("Failed to make a psfCandidate from star %d: %s", star.getId(), err)
+                continue
+
+            vmax = afwMath.makeStatistics(im, afwMath.MAX).getValue()
+            if not np.isfinite(vmax):
+                continue
```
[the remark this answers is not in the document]
parejkoj: We think it is a way of checking if the candidate cutout image doesn't contain NaN or inf. Whether this is the best approach for that, I don't know. Not knowing more, I'd rather not attempt to clarify it in place.

python/lsst/meas/algorithms/makePsfCandidates.py
```diff
@@ -0,0 +1,141 @@
...
+        psfCandidateList = []
+        didSetSize = False
+        for star in starCat:
+            try:
+                psfCandidate = makePsfCandidate(star, exposure)
+
+                # The setXXX methods are class static, but it's convenient to call them on
+                # an instance as we don't know exposures's pixel type
+                # (and hence psfCandidate's exact type)
+                if not didSetSize:
```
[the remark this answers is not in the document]
parejkoj: `psfCandidate` and it's parent class `SpatialCellImageCandidate`, inappropriately use `static` members for border, width, and height (among other things), meaning once you set them, they apply to *all* psfCandidates. We have to make those not `static` if we want to clean up this block. That's going to require an RFC, as according to @PaulPrice there is some code that assumes they are static.

python/lsst/meas/algorithms/makePsfCandidates.py
```diff
@@ -0,0 +1,141 @@
...
+        """Make a list of PSF candidates from a star catalog.
+
+        Parameters
+        ----------
+        starCat : `lsst.afw.table.SourceCatalog`
+            Catalog of stars, as returned by
+            ``lsst.meas.algorithms.starSelector.run()``.
+        exposure : `lsst.afw.image.Exposure` (optional)
+            The exposure containing the sources.
+
```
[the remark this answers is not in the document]
parejkoj: Good point. I've renamed it `starSelectedField`, since it's actually `str`, not a `bool`.

This is, however, reminding me that we don't have a unittest for makePsfCandidates...

tests/test_psfSelectTest.py
```diff
@@ -298,7 +300,8 @@ def testPsfCandidate(self):
 
         # select psf stars
         print("PSF selection")
-        psfCandidateList = self.starSelector.run(exposDist, sourceList).psfCandidates
+        stars = self.starSelector.run(exposDist, sourceList)
+        psfCandidateList = self.makePsfCandidates.run(stars.starCat, exposure=exposDist).psfCandidates
 
         # determine the PSF
         print("PSF determination")
```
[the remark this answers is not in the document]
parejkoj: No, it shouldn't. But there are so many janky things about this test, I don't want to touch it other than to get it to work with the new API.

python/lsst/meas/algorithms/makePsfCandidates.py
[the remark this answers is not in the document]
mrawls: Great idea! How about a unit test! 😸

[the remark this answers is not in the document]
parejkoj: Based on my work on the new test, I've renamed it again to `psfCandidateField`, which I think is even better. And there's a new unittest in another commit.

parejkoj:
I've pushed a cleanup commit (no review needed) based on your comments and a new test of `MakePsfCandidatestTask`. Can you please review the new test? It's not perfect (doesn't test the `max==NaN/inf` part of the code, but I couldn't figure out how to make fake sources with NaNs in them), but it at least tests two code paths, which is more than we had before.

tests/test_psfCandidate.py
```diff
@@ -37,46 +36,96 @@
     display = False
 
 
+def makeEmptyCatalog(psfCandidateField=''):
```
mrawls: Is there a reason you're using `''` as the default rather than just None?

tests/test_psfCandidate.py
```diff
@@ -37,46 +36,96 @@
...
+                ds9.dot("x", peak.getIx(), peak.getIy(), frame=1)
+
+    # There might be multiple footprints; only the one around x,y should go in the source
+    found = False
+    for fp in fpSet.getFootprints():
+        if fp.contains(afwGeom.Point2I(x, y)):
+            found = True
+            print(fp.getBBox())
+            break
+    assert found, "Unable to find central peak in footprint: faulty test"
```
mrawls: It's not immediately clear that this line is asserting `found` must be `True` and prints the "unable to find..." message if it's False. But if this is the correct/pythonic way to do so, I can live with it.

tests/test_psfCandidate.py
```diff
@@ -139,6 +188,48 @@ def testFaintNeighborMasking(self):
         self.checkCandidateMasking([(self.x+5, self.y, 0.5)], threshold=0.9, pixelThreshold=1.0)
 
 
+class MakePsfCandiatesTaskTest(lsst.utils.tests.TestCase):
```
mrawls: Probably don't want a typo in the class name ("Candidates") 😄

tests/test_psfCandidate.py
```diff
@@ -139,6 +188,48 @@ def testFaintNeighborMasking(self):
...
+    def setUp(self):
+        self.psfCandidateField = "psfCandidate"
+        self.catalog = makeEmptyCatalog(self.psfCandidateField)
+
+        # id=0 is bad because it's on the edge, so fails with a WARN: LengthError.
+        self.badIds = [1, ]
+        self.goodIds = [2, 3]
+        # x and y coordinate: keep these in sync with the above good/bad list.
+        self.xx = [0, 100, 200]
```
mrawls: xcoord and ycoord might be nicer variable names than xx and yy

mrawls:
Thanks for adding the test! It looks fine to me.

tests/test_psfCandidate.py
parejkoj: This is the syntax for the python `assert` statement. Is the purpose of the assert here clear?

Would you be happier with `assert found is True, "sometext"` instead?

tests/test_psfCandidate.py
parejkoj: Clearly I cannot type that word.

tests/test_psfCandidate.py
timj: I think it's really a question of it looking different because everywhere else in tests you see `self.assertTrue(found)` but here you see a bare assert because `self` has not been passed in to the function.

timj: If this was in a class that the other test classes inherited from (maybe with multiple inheritance) then the standard scheme could be used.

parejkoj: In this case, it's supposed to be "is my test even sane?", hence the raw assert.

parejkoj: I've added a note to clarify the purpose.

r-owen: Please don't use `foo is True`; it isn't the same as testing the boolean value of `foo` and it's not good Python unless you are sure you need exactly the `True` singleton.

parejkoj: Oh, right @r-owen I forgot about that. Anyway, I think it should be clear with the comment I added.

PR #119, opened 2018-05-01
DM-14253: Modernize python in meas_base and meas_algorithms

r-owen:


python/lsst/meas/algorithms/coaddPsf/coaddPsfContinued.py
```diff
@@ -20,12 +20,10 @@
...
-from __future__ import absolute_import
 
 __all__ = ["CoaddPsfConfig"]
 
-from lsst.utils import continueClass
 from .coaddPsf import CoaddPsfControl
 from lsst.pex.config import makeConfigClass
 
-makeConfigClass(CoaddPsfControl, "CoaddPsfConfig")
+CoaddPsfConfig = makeConfigClass(CoaddPsfControl, "CoaddPsfConfig")
```
timj: How did this work before?

r-owen: I'm not sure it did work before. I simply cannot imagine what it used to do

r-owen:
In most cases the bare exceptions were in plotting code, where it seemed that a "best effort" was intended, instead of expecting some particular exception. In cases where that was not true I tried to be more careful.

r-owen:
That said (about exceptions) it is sleazy and could be masking lots of bugs. The best I can say is this change didn't make the code any worse. It's not easily testable. I can file a ticket to have a closer look if you like, but I fear it may never get worked on.

python/lsst/meas/algorithms/utils.py
```diff
@@ -224,7 +224,7 @@ def showPsfCandidates(exposure, psfCellSet, psf=None, frame=None, normalize=True
                 # residuals using spatial model
                 try:
                     subtractPsf(psf, im, xc, yc)
-                except:
+                except Exception:
```
parejkoj: Please check if there's an obvious exception(s) raised by `subtractPSf` here.

python/lsst/meas/algorithms/utils.py
```diff
@@ -246,7 +246,7 @@ def showPsfCandidates(exposure, psfCellSet, psf=None, frame=None, normalize=True
 
                     try:
                         noSpatialKernel = psf.getKernel()
-                    except:
+                    except Exception:
```
parejkoj: Please check if there's an obvious exception(s) raised by `psf.getKernel()` here.

python/lsst/meas/algorithms/pcaPsfDeterminer.py
```diff
@@ -39,7 +39,7 @@
 from .spatialModelPsf import createKernelFromPsfCandidates, countPsfCandidates, \
     fitSpatialKernelFromPsfCandidates, fitKernelParamsToImage
 from .pcaPsf import PcaPsf
-from . import utils as maUtils
+from . import utils as mautils
```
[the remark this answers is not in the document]
r-owen: Because flake8 complained

tests/test_psfDetermination.py
```diff
@@ -183,10 +183,10 @@ def setUp(self):
                     if ix < 0 or ix >= self.mi.getWidth():
                         continue
 
-                    I = I0*psfVal(ix, iy, x + dx, y + dy, sigma1, sigma2, b)
-                    Isample = rand.poisson(I) if addNoise else I
+                    Ival = I0*psfVal(ix, iy, x + dx, y + dy, sigma1, sigma2, b)
+                    Isample = rand.poisson(Ival) if addNoise else Ival
```
parejkoj: Is `intensity` a better name here?

tests/test_psfDetermination.py
```diff
@@ -258,7 +258,7 @@ def subtractStars(self, exposure, catalog, chi_lim=-1):
             if bbox.contains(afwGeom.PointI(int(xc), int(yc))):
                 try:
                     measAlg.subtractPsf(psf, subtracted, xc, yc)
-                except:
+                except Exception:
```
parejkoj: Please check if there's an obvious exception(s) raised by subtractPSf here (as above).

python/lsst/meas/algorithms/pcaPsfDeterminer.py
```diff
@@ -488,36 +488,36 @@ def determinePsf(self, exposure, psfCandidateList, metadata=None, flagKey=None):
...
+                            mautils.showPsfCandidates(exposure, psfCellSet, psf=psf, frame=4,
                                                       normalize=normalizeResiduals,
                                                       showBadCandidates=showBadCandidates)
-                            maUtils.showPsfCandidates(exposure, psfCellSet, psf=psf, frame=5,
+                            mautils.showPsfCandidates(exposure, psfCellSet, psf=psf, frame=5,
                                                       normalize=normalizeResiduals,
                                                       showBadCandidates=showBadCandidates,
                                                       variance=True)
-                        except:
+                        except Exception:
```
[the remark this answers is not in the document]
r-owen: I don't think that is reasonable. It's 200 lines of code, with no documentation as to what exceptions it might raise. I admit that many of those lines are contained in their own try/except Exception blocks, but it's still a big job to try to figure out all the possible errors it might raise.

r-owen:
I did indeed search for "future" and "builtins" and removed all associated imports.

python/lsst/meas/algorithms/pcaPsfDeterminer.py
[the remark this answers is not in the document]
parejkoj: Complained about what? `maUtils` is follows our other camelCase renamings better.

python/lsst/meas/algorithms/utils.py
r-owen: Here is the full documentation for subtractPsf:
```
/**
 * Subtract a PSF from an image at a given position
 */
template<typename MaskedImageT>
double subtractPsf(afwDetection::Psf const& psf,      ///< the PSF to subtract
                   MaskedImageT *data,                ///< Image to subtract PSF from
                   double x,                          ///< column position
                   double y,                          ///< row position
                   double psfFlux                     ///< object's PSF flux (if not NaN)
                  )
```
I think this is a clear case where `except Exception:` is actually reasonable. This means "make a best effort to subtract the PSF and continue if it fails. Fortunately this is plotting code.

parejkoj:
Further thought about `except Exception as e: pass` blocks in plotting code: at bare minimum, add `log.warn("%s: %s"%(type(ex).__name__, e)` before the pass, so we at least have a log of what was raised.

python/lsst/meas/algorithms/utils.py
r-owen: It is C++ code, with different versions for different PSF classes. And none of it documents any exceptions that I can find.

python/lsst/meas/algorithms/pcaPsfDeterminer.py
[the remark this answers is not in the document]
r-owen: `warning       flake8: N812         lowercase 'maUtils' imported as non lowercase`

tests/test_dynamicDetection.py
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env python
 
-from __future__ import absolute_import, division, print_function
 
-from builtins import zip
```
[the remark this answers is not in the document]
r-owen: Not that I can see in my copy. I wonder why this got hidden as outdated? I have not pushed anything since May 1

python/lsst/meas/algorithms/pcaPsfDeterminer.py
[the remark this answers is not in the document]
parejkoj: Can you instead just do `from . import utils` instead of renaming it?

tests/test_psfDetermination.py
r-owen: I agree that this is a very nasty spot to ignore errors. I tried leaving out the try/except to see what got raised and nothing did, so I removed it.

r-owen:
I fixed one try/except Exception in a unit test and am declaring that enough. The plotting code is riddled with them and I think it will take some care and exercising the code to understand where exceptions should be logged or their scope reduced.

python/lsst/meas/algorithms/pcaPsfDeterminer.py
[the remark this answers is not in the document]
r-owen: OK

PR #124, opened 2018-06-05
DM-13297: CoaddBoundedField persistence not exact

r-owen:


src/CoaddBoundedField.cc
```diff
@@ -32,6 +32,28 @@
...
+ *
+ * If both pointers are set then return *a == *b
+ * else return a == b
+ */
+template <typename T>
+bool ptrEquals(T a, T b) {
+    if (a && b) {
+        return *a == *b;
+    }
+    return a == b;
```
TallJimbo: It'd be better to start with the check for pointer equality; if `a == b`, then it's guaranteed that `*a == *b`, and we can skip that potentially much more expensive test.

PR #129, opened 2018-07-05
DM-14527: Update for image API changes.

TallJimbo:


tests/test_psfCandidate.py
```diff
@@ -82,7 +82,7 @@ def createFakeSource(x, y, catalog, exposure, threshold=0.1):
     source['centroid_x'] = x
     source['centroid_y'] = y
 
-    exposure.getMaskedImage().getImage()[x, y] = 1.0
+    exposure.image[x, y, afwImage.PARENT] = 1.0
```
fred3m: Shouldn't this be `LOCAL` to replicate the old behavior?

tests/test_psfCandidate.py
```diff
@@ -139,7 +139,7 @@ def checkCandidateMasking(self, badPixels, extraPixels=[], size=25, threshold=0.
         """
         image = self.exposure.getMaskedImage().getImage()
         for x, y, f in badPixels + extraPixels:
-            image[x, y] = f
+            image[x, y, afwImage.PARENT] = f
```
fred3m: this should also be `LOCAL`

tests/test_psfCandidate.py
TallJimbo: I *think* this was one of the places where I looked carefully and then determined that they're equivalent.  But in retrospect it's probably better to just use `LOCAL` and not rely to much on the carefulness of that looking.

PR #138, opened 2018-08-24
DM-15532: Convert getTargetList to a classmethod

isullivan:


jdswinbank:
Ticket has been “won't fix”ed.

PR #141, opened 2018-10-11
DM-15767: Add previously-undeclared dependency on pex_policy.

TallJimbo:


laurenam:
This seems out of place for a ticket entitled "Remove pex_policy...", but ok!

TallJimbo:
Yeah, I had to add it because meas_algorithms still uses pex_policy (it's harder to remove here) but afw does not, so meas_algorithms can't rely on afw to bring in pex_policy anymore.

PR #143, opened 2018-11-30
DM-16338: Remove addFluxErr (permanently make it True)

parejkoj:


tests/test_loadReferenceObjects.py
```diff
@@ -105,12 +104,11 @@ def testFilterAliasMap(self):
         """Make a schema with filter aliases."""
         for defaultFilter in ("", "r", "camr"):
             for filterMap in ({}, {"camr": "r"}):
-                for addFluxErr in (False, True):
                     config = TrivialLoader.ConfigClass()
```
SimonKrughoff: If you are removing a level of iteration, don't you need to outdent the inside loop?

PR #144, opened 2019-01-03
DM-17043: Add selection on S/N in objectSizeStarSelector

laurenam:


python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -37,6 +36,11 @@
 
 
 class ObjectSizeStarSelectorConfig(BaseSourceSelectorTask.ConfigClass):
+    doFluxLimit = pexConfig.Field(
```
morriscb: Maybe a better name for these variables would be, as you say in the comment, applyFluxLimit and applySignalToNoiseLimit. It makes it slightly clearer what they are for where "do" seems to apply a sort of processing.

python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -410,13 +434,18 @@ def selectSources(self, sourceCat, matches=None, exposure=None):
         width = numpy.sqrt(0.5*(xx + yy))
         with numpy.errstate(invalid="ignore"):  # suppress NAN warnings
             bad = reduce(lambda x, y: numpy.logical_or(x, sourceCat.get(y)), self.config.badFlags, False)
-            bad = numpy.logical_or(bad, flux < self.config.fluxMin)
             bad = numpy.logical_or(bad, numpy.logical_not(numpy.isfinite(width)))
             bad = numpy.logical_or(bad, numpy.logical_not(numpy.isfinite(flux)))
+            if self.config.doFluxLimit:
```
morriscb: Perhaps it would be better to set the Min/Max values to a default of `None` and test if a value has been set rather than using the extra boolean variables. For instance

    if self.config.fluxMin is not None:
        bad = numpy.logical_or(bad, flux < self.config.fluxMin)

python/lsst/meas/algorithms/objectSizeStarSelector.py
laurenam: Just to be thorough, @PaulPrice, do you have any strong feelings about the above?

PaulPrice: I don't think it's worth the trouble of changing; surely we have much more important things to do.

laurenam: So you would leave the **do**s?

laurenam: @morriscb are you ok with me leaving the ticket essentially as is?

PaulPrice: Is there a good reason to change the names? Good enough to spend a few days on an RFC and implementation and backwards compatibility and support?

python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -49,6 +53,23 @@ class ObjectSizeStarSelectorConfig(BaseSourceSelectorTask.ConfigClass):
...
+    doSignalToNoiseLimit = pexConfig.Field(
+        doc="Apply signal-to-noise (i.e. flux/fluxErr) limit to Psf Candidate selection?",
+        dtype=bool,
+        default=False,
+    )
+    signalToNoiseMin = pexConfig.Field(
+        doc="specify the minimum signal-to-noise for good Psf Candidates",
+        dtype=float,
+        default=20.0,
+        check=lambda x: x >= 0.0,
```
morriscb: Is there are reason that this class should never work with negative flux objects? This is related to the fluxMin/Max fields. I understand that this class is mainly used for selection PSF fitting objects, but there is nothing necessarily stopping this from being used, for instance on catalogs produced by difference images.

python/lsst/meas/algorithms/objectSizeStarSelector.py
laurenam: I was fine following the convention you used and sticking to it would cause less churn.  The RFC is more about adapting the selectors to make (homogeneous) use of the base and derived tasks you defined in **sourceSelector.py** (API changing and thus RFC-required).  I also want to change the default selection criterion to be S/N instead of flux for this particular selector, but I'll probably need to run at least the HSC RC2 dataset to make sure things still look good after this change.

I'll wait to see if Chris is ok with keeping the status quo...

python/lsst/meas/algorithms/objectSizeStarSelector.py
laurenam: This is technically a preexisting condition (well, slightly derived from the preexisting fluxMin condition!).  I'm not in the market for making any such changes on this ticket, but I do think it's true that this selector is only being used in PSF modeling candidate selection (I'm not sure what the diffim selectors do...)

python/lsst/meas/algorithms/objectSizeStarSelector.py
morriscb: I feel that the extra booleans make the config fairly cluttered and are also not very descriptive as to what they are doing. I would prefer the option below instead where if the value is set and not `None`, the cuts are applied vs having the an extra boolean that needs to be set in addition to the values. The extra booleans could lead to silent failures where the one expects the selection to be applied but it is not because the boolean was not set. The likelihood of this happening is increased by one selection being default `True` and the other being default `False`. I'm not sure how necessary an RFC is for this but feel free to ask around. The option below of testing `None` would seem to work in cases where the task this selector is being called by sets it's own default for the value.

laurenam: Ok, I think I can get on board with getting rid of the booleans here an having the None/not None-ness of the limits determine whether a cut is made (thus making the naming discussion moot).

laurenam: Allrighty...have a look and see if you're ok with it now.  Jenkins is running:
https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/29231

PaulPrice: I really don't like the idea of having `None`-ness disable a selection. It is extremely convenient to be able to disable a selection on the command-line through a boolean flag. Also, having a boolean flag allows you to disable a selection while preserving its parameters.

python/lsst/meas/algorithms/objectSizeStarSelector.py
morriscb: That's fair, it may be worth ticketing this for the future and converting these checks into a `validate` method that could, for instance, check that the min value is less that the max value.

python/lsst/meas/algorithms/objectSizeStarSelector.py
morriscb: Okay, it's not quite directly your case Paul, but in general I can see how using `None` to disable the cut could be fairly awkward and require a bit more in depth parsing of the config rather than just glancing at the config and setting the boolean. The tickets is already marked as `Reviewed` so I'm happy to use Laruen's reasoning above that the other selectors use these variables to move along.

PR #145, opened 2019-01-16
tickets/DM-14819

natelust:
Add reference obj loader class

Make a reference object loader that is a class rather than a task,
so that it fits in with the data access design of gen3 middleware.

### DIFF
--- a/python/lsst/meas/algorithms/loadReferenceObjects.py
+++ b/python/lsst/meas/algorithms/loadReferenceObjects.py
@@ -21,9 +21,11 @@
 # see <https://www.lsstcorp.org/LegalNotices/>.
 #
 
-__all__ = ["getRefFluxField", "getRefFluxKeys", "LoadReferenceObjectsTask", "LoadReferenceObjectsConfig"]
+__all__ = ["getRefFluxField", "getRefFluxKeys", "LoadReferenceObjectsTask", "LoadReferenceObjectsConfig",
+           "ReferenceObjectLoader"]
 
 import abc
+import itertools
 
 import astropy.time
 import astropy.units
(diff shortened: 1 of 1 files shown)


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -32,11 +34,513 @@
...
+        md.add('SMATCHV', 1, 'SourceMatchVector version number')
+        filterName = "UNKNOWN" if filterName is None else str(filterName)
+        md.add('FILTER', filterName, 'filter name for photometric data')
+        md.add('EPOCH', "NONE" if epoch is None else epoch, 'Epoch (TAI MJD) for catalog')
+        return md
+
+    @staticmethod
+    def addFluxAliases(refCat, defaultFilter, filterReferenceMap):
+        """This function creates maps the input reference catalog into a new reference
+        catalog with added flux columns and aliases between camera and reference flux.
```
TallJimbo: Something went wrong in this sentence somewhere around "creates maps".

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -32,11 +34,513 @@
...
+                refFluxErrName = refFluxName + "Sigma"
+                camFluxErrName = camFluxName + "Sigma"
+                aliasMap.set(camFluxErrName, refFluxErrName)
+
+        return refCat
+
+    @staticmethod
+    def remapReferenceCatalogSchema(refCat, *, filterNameList=None, position=False, photometric=False):
+        """This class takes in a reference catalog and maps it to a catalog with additional
+        columns.
```
TallJimbo: This is not a class.  Also probably better to use a different verb than "maps", which is a bit unclear.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -32,11 +34,513 @@
...
+
+        Parameters
+        ----------
+        refCat : `lsst.afw.table.SourceCatalog`
+            Reference catalog to map to new catalog
```
TallJimbo: I think this should probably be `SimpleCatalog`.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -32,11 +34,513 @@
...
+
+        Returns
+        -------
+        expandedCat : `lsst.afw.table.SourceCatalog`
```
TallJimbo: `SimpleCatalog` here, too.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -756,36 +1245,131 @@ def applyProperMotions(self, catalog, epoch):
...
+        - ``pm_raErr`` : Error in ``pm_ra`` (rad/yr), optional.
+        - ``pm_dec`` : Proper motion in Declination (rad/yr,
+            North positive)
+        - ``pm_decErr`` : Error in ``pm_dec`` (rad/yr), optional.
+        - ``epoch`` : Mean epoch of object (an astropy.time.Time)
+    epoch : `astropy.time.Time` (optional)
+        Epoch to which to correct proper motion and parallax,
+        or None to not apply such corrections.
+    """
+    if ("epoch" not in catalog.schema or "pm_ra" not in catalog.schema or "pm_dec" not in catalog.schema):
```
TallJimbo: Unnecessary parens.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -32,11 +34,513 @@
...
+        self.log.info(f"Loaded {len(refCat)} reference objects")
+
+        if epoch is not None and "pm_ra" in refCat.schema:
+            # check for a catalog in a non-standard format
+            if isinstance(refCat.schema["pm_ra"].asKey(), lsst.afw.table.KeyAngle):
+                applyProperMotionsImpl(self.log, refCat, epoch)
+            else:
+                self.log.warn("Catalog pm_ra field is not an Angle; not applying proper motion")
+
+        expandedCat = self.remapReferenceCatalogSchema(refCat, position=True)
```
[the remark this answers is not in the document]
natelust: I think this is not a problem, as it is just adding position columns, I have made a few changes that might make it even safer, but it it probably worth going through this with you

parejkoj:
Does this code have no tests?

TallJimbo:
> Does this code have no tests?

Only the pre-existing ones, which admittedly do not test the new entry points (though right now the old entry points are very much the most important ones to test, as the new ones are only used in Gen3 and hence by a small number of people).

We are not yet doing automated testing for the new PipelineTasks because the infrastructure that would make that easy is being developed along with the tasks themselves, and we're not quite to the point where we can put it all together yet.

So yes, there is some technical debt creation happening here, but it's happening because we think it actually will be easier to remove later than address it up front.

parejkoj:
> Only the pre-existing ones...

I don't see how the existing tests are being executed by this code at all.

> We are not yet doing automated testing for the new PipelineTasks because the infrastructure that would make that easy is being developed along with the tasks themselves, and we're not quite to the point where we can put it all together yet.

If so, that's all the more reason to have unittests of individual pieces of functionality. We're building an API here: that should make it easy enough to write tests of the new code. I don't understand how we can claim to be producing working code without demonstrating it works via tests of each piece of functionality.

> So yes, there is some technical debt creation happening here, but it's happening because we think it actually will be easier to remove later than address it up front.

You know as well as I that writing tests after the fact always gets bumped far down the priority list, and is much harder to do, than writing them first.

TallJimbo:
> I don't see how the existing tests are being executed by this code at all.

I just meant that the old reference-loading Tasks are not being modified, and their old tests are still passing; that's not much test coverage, as the shared code is actually pretty small right now.

>> We are not yet doing automated testing for the new PipelineTasks because the infrastructure that would make that easy is being developed along with the tasks themselves, and we're not quite to the point where we can put it all together yet.

> If so, that's all the more reason to have unittests of individual pieces of functionality. We're building an API here: that should make it easy enough to write tests of the new code. I don't understand how we can claim to be producing working code without demonstrating it works via tests of each piece of functionality.

I actually disagree that we're putting together a new API right now - that would be quite premature, because right now this code is being torqued in unnatural directions by the need to simultaneously satisfy two complex systems that have quite different conceptions of how reference catalogs are managed.  We are trying to maintain an existing API while integrating a new system.

>> So yes, there is some technical debt creation happening here, but it's happening because we think it actually will be easier to remove later than address it up front.

> You know as well as I that writing tests after the fact always gets bumped far down the priority list, and is much harder to do, than writing them first.

I agree with the first part (it's all too easy to bump writing tests down the priority list later), but not the latter: I do not think TDD or any other write-tests-immediately model is a fast way to prototype designs.
I think we're exiting the prototype stage for Gen3 now, but I also don't think it's always a good use of developer time to write complex custom test harnesses (including appropriate test datasets) for intermediate pipeline components when the same coverage can be obtained by adding checks for that component to a larger-scale test that runs them in a more realistic context.  More tests are always good to have, but when I can get coverage in an integration test for 1/10 the cost of lower-level test, I think it's a worthwhile trade-off.

Where I am worried here is the degree to which we are indeed relying on *one-off* integration tests, rather automated ones, right now.  Here's where we're at on that:
 - all the test data we need is already in the ci_hsc test dataset;
 - we can't write tests in places like meas_algorithms that depend on the ci_hsc EUPS package;
 - adding new tests to the current ci_hsc git repo is painful because of its idiosyncratic SCons usage
 - once we have an end-to-end set of PipelineTasks we can trivially put together a Gen3 ci_hsc equivalent (same data, new scripts).  When we do that, we can do it right and make it easy to add standard unit tests for components like this one run them in a realistic concept without a ton of custom harness work.

This would be even nicer to have while working on the PipelineTasks, of course, and I've tried twice to get ci_hsc in the kind of shape that would allow that before getting the full set of PipelineTasks done (DM-16264, DM-16803).  I quit both times after identifying blockers that would be much easier to solve with more of the Gen3 infrastructure in place.  We absolutely need to make sure that happens eventually, and it still will be a significant amount of work to put it together.  I share your concern that some of it may ultimately fall by the wayside - the core ci_hsc test script is sufficiently important that I'm sure we'll get that done, but adding detailed tests of components like this could fall off the priority list, to our detriment.  But writing tests for those now is simply not an option, both because it's more difficult long term and we have a *very* hard deadline coming up in the very near future that involves a demonstration of an integrated prototype, not a stable API.

PR #153, opened 2019-02-15
DM-17917: Add doUnresolved to ReferenceSourceSelectorTask

erykoff:


tests/test_sourceSelector.py
```diff
@@ -187,6 +187,7 @@ def setUp(self):
         self.config.magLimit.fluxField = "flux"
         self.config.doMagLimit = True
         self.config.doFlags = True
+        self.config.doUnresolved = False
```
TallJimbo: Isn't this the default?

erykoff: It is, but I'm following the redundancy in the ScienceSourceSelectorTask test.

PR #154, opened 2019-03-07
DM-17029: Update LoadReferenceObjectsTask to output fluxes in nanojansky

parejkoj:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -43,6 +43,87 @@
...
+        This supports the "write=False" mode of `bin/convert_to_nJy.py`.
+
+    Returns
+    -------
+    catalog : `lsst.afw.table.SimpleCatalog` or None
+        The converted catalog, or None if ``doConvert`` is False.
+
+    Notes
+    -----
+    Support for old units in reference catalogs will be removed after 18.0.
```
SimonKrughoff: I think this may not happen until after v19, right?  Maybe it's safer to say it will be removed after the release of late calendar year 2019.

tests/test_htmIndex.py
```diff
@@ -198,6 +199,7 @@ def testAgainstPersisted(self):
         ex2 = testCat.extract('*')
         self.assertEqual(set(ex1.keys()), set(ex2.keys()))
         for kk in ex1:
+            print(ex1[kk], ex2[kk])
```
SimonKrughoff: Does this really need to be printed?

tests/test_htmIndex.py
```diff
@@ -475,6 +480,36 @@ def testProperMotion(self):
...
 
+    def testLoadVersion0(self):
+        """Test reading a pre-written format_version=0 (Jy flux) catalog.
+        It should be converted to have nJy fluxes.
+        """
+        path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data/version0')
+        loader = LoadIndexedReferenceObjectsTask(butler=dafPersist.Butler(path))
+        self.assertEqual(loader.dataset_config.format_version, 0)
+        result = loader.loadSkyCircle(make_coord(10, 20), 5*lsst.geom.degrees, filterName='a')
+        # TODO: assert that log.warn messages are emitted.
```
SimonKrughoff: Is this TODO associated with a ticket?

tests/test_htmIndex.py
```diff
@@ -475,6 +480,36 @@ def testProperMotion(self):
...
+        self.assertFloatsEqual(catalog['b_flux']*1e9, result.refCat['b_flux'])
+        self.assertFloatsEqual(catalog['b_fluxSigma']*1e9, result.refCat['b_fluxErr'])
+
+    def testLoadVersion1(self):
+        """Test reading a format_version=1 catalog (fluxes unchanged)."""
+        path = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'data/version1')
+        loader = LoadIndexedReferenceObjectsTask(butler=dafPersist.Butler(path))
+        self.assertEqual(loader.dataset_config.format_version, 1)
+        result = loader.loadSkyCircle(make_coord(10, 20), 5*lsst.geom.degrees, filterName='a')
+        # TODO: assert that no log.warn is emitted.
```
SimonKrughoff: Same as above re: ticket for TODO.

bin.src/convert_refcat_to_nJy.py
```diff
@@ -0,0 +1,128 @@
...
+        sys.exit(-1)
+    if not is_old_schema(schema_file):
+        print("Catalog does not contain old-style fluxes; nothing to convert.")
+        sys.exit(0)
+
+    files = glob.glob(os.path.join(args.path, "*.fits"))
+    with concurrent.futures.ProcessPoolExecutor(max_workers=args.nprocesses) as executor:
+        futures = executor.map(process_one, files, itertools.repeat(args.write), itertools.repeat(args.quiet))
+        # so that exceptions don't get lost
+        for future in futures:
```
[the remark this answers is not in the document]
parejkoj: I reworded it:

```
        # we have to at least loop over the futures, otherwise exceptions will be lost
```

bin.src/convert_refcat_to_nJy.py
```diff
@@ -0,0 +1,128 @@
...
+            pass
+
+    if args.write:
+        with open(os.path.join(args.path, 'config.py'), 'a') as configFile:
+            configFile.write("\n\n# Updated refcat from version 0 to have nJy flux units\n")
+            configFile.write("config.format_version=1\n")
+        if not args.quiet:
+            print("Added `format_version=1` to config.py")
```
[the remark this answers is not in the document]
parejkoj: It's not a Task, and all of these print statements are outside the futures loop, so I think just `print` is fine here.

bin.src/convert_refcat_to_nJy.py
```diff
@@ -0,0 +1,128 @@
...
+    files = glob.glob(os.path.join(args.path, "*.fits"))
+    with concurrent.futures.ProcessPoolExecutor(max_workers=args.nprocesses) as executor:
+        futures = executor.map(process_one, files, itertools.repeat(args.write), itertools.repeat(args.quiet))
+        # so that exceptions don't get lost
+        for future in futures:
+            pass
+
+    if args.write:
+        with open(os.path.join(args.path, 'config.py'), 'a') as configFile:
+            configFile.write("\n\n# Updated refcat from version 0 to have nJy flux units\n")
```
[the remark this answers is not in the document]
parejkoj: There is no such thing as `format_version=0` in any existing files and there never will be since I introduced that version in this ticket, which also makes version 1 the default (with no way to write version 0).

I guess we could make an explicit check of `format_version` part of `is_old_schema`, in addition to the check of types? How does one read a config file like this outside of the `pex.Config` environment?

tests/test_htmIndex.py
parejkoj: I don't know. It was a todo for me while I was writing it, but I didn't actually implement that part, and I'm not sure how much it matters. Is it worth checking that we emit some kind of message for old catalogs?

bin.src/convert_refcat_to_nJy.py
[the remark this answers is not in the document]
SimonKrughoff: Sounds good.

bin.src/convert_refcat_to_nJy.py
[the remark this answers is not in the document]
SimonKrughoff: Oh, good point it's not a `Task`.

bin.src/convert_refcat_to_nJy.py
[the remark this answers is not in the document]
SimonKrughoff: I understand this may be more complicated than you want, but the safest thing is to load the config, set the version and write it back out.  This is also presumably future proof.

(untested) code snippet follows:
```
import lsst.meas.algorithms as meas_alg
config = meas_alg.ingestIndexReferenceTask.DatasetConfig()
config.load(os.path.join(args.path, 'config.py'))
config.format_version = 1
config.save(os.path.join(args.path, 'config.py'))
```

tests/test_htmIndex.py
SimonKrughoff: There are a lot of place we don't check log messages.  If you are not going to do it, I'm fine just removing the TODOs.

python/lsst/meas/algorithms/ingestIndexReferenceTask.py
```diff
@@ -26,20 +26,39 @@
...
+
+    Parameters
+    ----------
+    catalog - `~lsst.afw.table.SimpleCatalog`
+        Catalog to which metadata should be attached.  Will be modified
+        in-place.
+    """
+    md = PropertyList()
+    md.set("REFCAT_FORMAT_VERSION", LATEST_FORMAT_VERSION)
+    catalog.setMetadata(md)
```
parejkoj: @TallJimbo : would this overwrite the rest of the metadata in the catalog? Should I change it to do this first `md = catalog.getMetadata()`?

TallJimbo: That would indeed be safer.  I recall thinking that the initial `PropertySet` was going to be `None` in the cases that mattered here, but I could easily believe that was only the case in the first place I used this function, and then I forgot about it in the next place.

parejkoj:
@TallJimbo , @SimonKrughoff : should we re-review after I've added and tweaked Jim's additions to improve gen3 compatibility, or is my tweaking it enough of a review?

TallJimbo:
Your updates to my changes look good to me, and generally I feel like the spirit of code review has been met if two people have looked at the code.

PR #158, opened 2019-03-27
DM-18203: Set encoding for reading text catalogs.

jdswinbank:


tests/test_readTextCatalog.py
```diff
@@ -112,6 +114,25 @@ def testTooFewColumnNames(self):
...
 
+    def testEncoding(self):
+        """Test that we can set the text encoding.
+        """
+        # Attempting to read the table as pure ASCII data should fail because
+        # it contains a ``£``.
+        config = ReadTextCatalogTask.ConfigClass()
+        config.encoding = "ascii"
+        task = ReadTextCatalogTask(config=config)
+        # This will generate a ResourceWarning due to a NumPy bug.
```
timj: If you want to make the test output clean consider using a warnings.catch_warnings with a simplefilter to disable warning output for that command.

jdswinbank: I don't think you can, at least not trivially — I tried this, but the warning isn't generated in this code, but deep in the guts of Python: it comes from `/Users/jds/Projects/LSST/stack/python/miniconda3-4.5.4/envs/lsst-scipipe-fcd27eb/lib/python3.6/traceback.py:216` sometime after this test case has finished, so a context in this test case can't catch it.

I considered putting in a filter to drop all `ResourceWarnings` from `traceback.py:216`, but that seems less than optimal. Other suggestions welcome!

timj: Maybe it's fixed in the new numpy that we will be using from tomorrow...

jdswinbank: Pretty sure it's still broken in the latest version. :(

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -47,6 +47,11 @@ class ReadTextCatalogConfig(pexConfig.Config):
         default=',',
         doc='Delimiter to use when reading text reference files.  Comma is default.'
     )
+    encoding = pexConfig.Field(
+        dtype=str,
+        default="utf-8",
+        doc="Encoding of text reference file."
```
parejkoj: Do we really need to make this configurable? Can't we just make it `utf-8` forever and forget about it?

tests/test_readTextCatalog.py
parejkoj: Thanks for filing a numpy ticket on this and linking it here. However, should we have a jira ticket to remove this comment once numpy is fixed? I might be confusing in "future numpy is fixed" land...

parejkoj:
Your new commit is fine. Squash it.

PR #160, opened 2019-04-03
DM-10800: Merge matcherSourceSelector and matcherPessimisticSourceSelector

morriscb:
Add unittest for edge, interp and saturated.

### DIFF
--- a/python/lsst/meas/algorithms/matcherSourceSelector.py
+++ b/python/lsst/meas/algorithms/matcherSourceSelector.py
@@ -40,6 +40,13 @@ class MatcherSourceSelectorConfig(BaseSourceSelectorConfig):
         "(in the flux specified by sourceFluxType); <= 0 for no limit",
         default=40,
     )
+    excludePixelFlags = pexConfig.Field(
+        dtype=bool,
+        doc="Exclude objects that have saturated, interpolated, or edge "
+            "pixels using PixelFlags. For matchOptimisticB set this to False "
+            "to recover previous matcher selector behavior.",
+        default=True,
+    )
 
 
 @pexConfig.registerConfigurable("matcher", sourceSelectorRegistry)
@@ -82,6 +89,8 @@ class MatcherSourceSelectorTask(BaseSourceSelectorTask):
         self._getSchemaKeys(sourceCat.schema)
 
         good = self._isUsable(sourceCat)
+        if self.config.excludePixelFlags:
+            good = good & self._isGood(sourceCat)
         return Struct(selected=good)
 
     def _getSchemaKeys(self, schema):
@@ -96,6 +105,10 @@ class MatcherSourceSelectorTask(BaseSourceSelectorTask):
         self.fluxKey = schema[fluxPrefix + "instFlux"].asKey()
         self.fluxFlagKey = schema[fluxPrefix + "flag"].asKey()
         self.fluxErrKey = schema[fluxPrefix + "instFluxErr"].asKey()
+
+        self.edgeKey = schema["base_PixelFlags_flag_edge"].asKey()
+        self.interpolatedCenterKey = schema["base_PixelFlags_flag_interpolatedCenter"].asKey()
+        self.saturatedKey = schema["base_PixelFlags_flag_saturated"].asKey()
 
     def _isParent(self, sourceCat):
         """Return True for each source that is the parent source."""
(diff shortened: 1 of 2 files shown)


python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -40,6 +40,12 @@ class MatcherSourceSelectorConfig(BaseSourceSelectorConfig):
         "(in the flux specified by sourceFluxType); <= 0 for no limit",
         default=40,
     )
+    excludeSaturated = pexConfig.Field(
+        dtype=bool,
+        doc="Exclude saturated objects using PixelFlags. For matchOptimisticB "
+            "set this to False to recover previous matcher selector behavior.",
+        default=True,
+    )
```
erykoff: The config parameter is called `excludeSaturated`, but it also rejects things that are on the edge and have an interpolated center value.  I think this config name should be changed.

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -97,6 +105,10 @@ def _getSchemaKeys(self, schema):
         self.fluxFlagKey = schema[fluxPrefix + "flag"].asKey()
         self.fluxErrKey = schema[fluxPrefix + "instFluxErr"].asKey()
 
+        self.edgeKey = schema["base_PixelFlags_flag_edge"].asKey()
+        self.interpolatedCenterKey = schema["base_PixelFlags_flag_interpolatedCenter"].asKey()
+        self.saturatedKey = schema["base_PixelFlags_flag_saturated"].asKey()
+
```
erykoff: I know this matches the previous behavior, but I was curious about the fact that we care about whether any pixel is saturated, but only the center is interpolated?

python/lsst/meas/algorithms/matcherSourceSelector.py
```diff
@@ -132,46 +144,16 @@ def _isUsable(self, sourceCat):
...
         Return True for each source that is usable for matching, even if it may
         have a poor centroid.
 
         For a source to be usable it must:
-        - have a valid centroid
-        - not be deblended
-        - have a valid instFlux (of the type specified in this object's constructor)
-        - have adequate signal-to-noise
+        - Not be on a CCD edge.
+        - Not have an interpolated pixel value
```
erykoff: I believe this is only checking if the center is interpolated, not if the object has any interpolated pixels.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: So looking at the code for PixelFlags it seems as if what `Centroid` means is that the failure is within the 3x3 pixels around the centroid where as the other flags are set when the any pixel is "bad" within the `footprint`. I think the distinction here is that the flag "saturated" likely is a very bright object saturating the pixels and that interpolated is from a cosmic ray that has been removed. InterpolatedCenter here then only removes objects that have had a cosmic ray directly in their center. I ran a few tests with validate_drp and got consistently worse, but not catastrophically so, results using both `staturatedCetner` and just `interpolated`. The best option seems to be `interplatedCenter` and `saturated`. I'm letting a longer validate_drp run go using HSC, we'll see if these results hold similarly.

erykoff: Thanks, that's interesting.  I didn't realize it was 3x3, but I did know that the other flags were over the footprint.  Whatever you find, it'll be worth putting a comment in the code (or a link to a comment on the ticket) about these decisions.  Also, the docs below should be adjusted accordingly of course.

RobertLuptonTheGood: I don't think you want to throw away all saturated objects (that's at least one pixel in the Footprint).  There should be `saturatedCenter`

morriscb: So I was a bit wrong. Saturated pixels are also interpolated over in the default settings and the other interpolated Pixels are from cosmics and other image defects. So what in effect this cut is doing is throwing out every footprint with a saturated pixel and keeping those that have interpolated pixels over CRs and defects as long as they are not in the center 3x3 pixels. As I said, drp metrics seem to perform a bit worse when putting `Center` on both. I'm running HSC though the metrics to have more test data. I'll comment with the results on the ticket. Right now, I think this question is moving beyond the scope of the ticket so if we really want to investigate it, I would suggest making a ticket for it.

python/lsst/meas/algorithms/matcherSourceSelector.py
morriscb: does `excludePixelFlags` or `excludeSatInterpEdge` work? Any better ideas?

erykoff: Of those `excludePixelFlags` is better and allows us to change the exact flags being excluded later.

PR #161, opened 2019-04-22
DM-19373: Experimental Defects class for collecting Defect

timj:


python/lsst/meas/algorithms/defects.py
```diff
@@ -53,3 +72,350 @@ def policyToBadRegionList(policyFile):
...
+
+        for r in table:
+            record = r.extract("*")
+
+            if isFitsRegion:
+                if record["SHAPE"] != "BOX":
+                    log.warning("Defect lists can only be defined using BOX not %s",
+                                record["SHAPE"])
+                box = lsst.geom.Box2I.makeCenteredBox(lsst.geom.Point2D(record["X"], record["Y"]),
+                                                      lsst.geom.Extent2I(record["R"]))
```
czwa: Am I correct that this will always make the minimal box that contains the defect if it is not "BOX"?

timj: No. I've actually just fixed that logic. After the warning there was meant to be a `continue` so that it wouldn't break.  I did add support for POINT whilst I was there.

PR #183, opened 2019-10-30
DM-18036: Use Tables to read FITS

SimonKrughoff:
This also includes fixing the unit tests.
Using astropy.table.Table also solves issues dealing with endianness of values coming from FITS files.

### DIFF
--- a/tests/test_readFitsCatalog.py
+++ b/tests/test_readFitsCatalog.py
@@ -47,14 +47,14 @@ def makeFitsTable():
 
     # table for HDU 1
     cols1 = [
-        fits.Column(name='name', format='10A', array=["object 1", "object 2"]),
+        fits.Column(name='name', format='10A', array=np.array(["object 1", "object 2"])),
         fits.Column(name='ra', format='E', array=[10, 5]),
         fits.Column(name='dec', format='E', array=[-5, 45]),
         fits.Column(name='counts', format='J', array=[1000, 2000]),
         fits.Column(name='flux', format='D', array=[1.1, 2.2]),
         fits.Column(name='resolved', format='L', array=[True, False]),
     ]
-    hdu1 = fits.BinTableHDU.from_columns(fits.ColDefs(cols1))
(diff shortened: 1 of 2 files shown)


tests/test_readFitsCatalog.py
```diff
@@ -118,15 +124,25 @@ def testHDU1GivenNames(self):
...
 
     def testHDU2(self):
         """Test reading HDU 2 with original order"""
         config = ReadFitsCatalogTask.ConfigClass()
         config.hdu = 2
         task = ReadFitsCatalogTask(config=config)
         arr = task.run(FitsPath)
-        self.assertTrue(np.array_equal(arr, self.arr2))
+        for name in self.arr1.dtype.names:
+            # Table form disk ends up being bytes and in memory is unicode
```
erykoff: I think this should be "from disk".  But I'm confused why there are these two options below?  Why isn't it one or the other?

SimonKrughoff: I'll fix the typo.

I don't know why it does this, but the array constructed from a Table read from a fits file on disk has the strings as type `bytes`.  If the array comes from a Table constructed from a FITS object in memory, they are `unicode` even if I try to force them to be `bytes`.

I didn't want to try to figure out if it was a bug or not, so just handle the one string like column separately.  I know that's not particularly satisfying, but I'd prefer to figure out the difference without the pressure of the release blockage.  I'll file a ticket if that's ok with you.

erykoff: At least put that in the comments, though is it possible that this could bite a user?

erykoff: Hold on ... what ticket is this associated with?  It should be in the PR name.

SimonKrughoff: Gah.  Very sorry about that.  I was in a rush to make the PRs.  This is DM-18036.

SimonKrughoff:
@erykoff I know you already approved, but I simplified the test greatly.  Feel free to have another look if you wish.

PR #187, opened 2020-01-15
DM-23077: Change default behavior for interpolate

SimonKrughoff:
The default was to raise an error if interpolating outside the bounds of the curve.  The default now is to clamp to 0. outside the range of the curve.

### DIFF
--- a/python/lsst/meas/algorithms/simple_curve.py
+++ b/python/lsst/meas/algorithms/simple_curve.py
@@ -84,7 +84,7 @@ class Curve(ABC):
         pass
 
     @abstractmethod
-    def evaluate(self, detector, position, wavelength, kind='linear'):
+    def evaluate(self, detector, position, wavelength, kind='linear', bounds_error=False, fill_value=0):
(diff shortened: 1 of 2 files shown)


python/lsst/meas/algorithms/simple_curve.py
```diff
@@ -97,9 +97,16 @@ def evaluate(self, detector, position, wavelength, kind='linear'):
...
-        kind : `str`
-            The type of interpolation to do.  See documentation for
-            `scipy.interpolate.interp1d` for accepted values.
+        kind : `str`, optional
+            The type of interpolation to do (default is 'linear').
+            See documentation for `scipy.interpolate.interp1d` for
+            accepted values.
+        bounds_error : `bool`, optional
+            Raise error if interpolating outside the range of x
+            (default is False)?
```
timj: I'd put the question mark at the end of the line above and have the default state as a distinct item. Otherwise I keep thinking I'm being asked what the default is. Also add a Raises section to the docstring saying what this exception might be.

PR #189, opened 2020-01-17
DM-23056: Suppress FutureWarnings from LSST code

jdswinbank:


python/lsst/meas/algorithms/detection.py
```diff
@@ -698,8 +699,13 @@ def reEstimateBackground(self, maskedImage, backgrounds):
...
         self.log.info("Resubtracting the background after object detection")
-        maskedImage -= bg.getImageF()
-        backgrounds.append(bg)
+        maskedImage -= bg.getImageF(self.background.config.algorithm,
+                                    self.background.config.undersampleStyle)
+
+        actrl = bg.getBackgroundControl().getApproximateControl()
+        backgrounds.append((bg, getattr(afwMath.Interpolate, self.background.config.algorithm),
+                            bg.getAsUsedUndersampleStyle(), actrl.getStyle(), actrl.getOrderX(),
+                            actrl.getOrderY(), actrl.getWeighting()))
```
erykoff: I assume this is the new API, but boy that is a verbose yet opaque call.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -320,9 +326,16 @@ def fitBackground(self, maskedImage, nx=0, ny=0, algorithm=None):
...
 
-        bctrl = afwMath.BackgroundControl(algorithm, nx, ny,
-                                          self.config.undersampleStyle, sctrl,
-                                          self.config.statisticsProperty)
+        # This call to a deprecated BackgroundControl constructor is necessary
+        # to support the algorithm parameter; it should be replaced with
+        #
+        #     afwMath.BackgroundControl(nx, ny, sctrl, self.config.statisticsProperty)
+        #
+        # when algorithm has been deprecated and removed (DM-22814).
```
erykoff: Should this be marked TODO or is that optional when the ticket is mentioned?

python/lsst/meas/algorithms/detection.py
jdswinbank: 💯 

I found this totally impenetrable, and only worked it out by a combination of cargo-culting and trial-and-error. I think this is the best support for @kfindeisen's suggestion on [DM-23113](https://jira.lsstcorp.org/browse/DM-23113) that we should just back out these changes altogether. (Which we might still do, but I wouldn't be inclined to do so on a Friday afternoon!)

python/lsst/meas/algorithms/subtractBackground.py
jdswinbank: I'm not sure that we actually have a rule about that, but I think it would be a good idea to do so. I'll change it.

python/lsst/meas/algorithms/detection.py
kfindeisen: Yeah, `BackgroundList` is exactly what first made me think that maybe I shouldn't be touching this code. 😱

PR #191, opened 2020-03-28
DM-24060: Improve table creation inside Defects

timj:
Now assign whole rows rather than assigning elements
in a loop.


python/lsst/meas/algorithms/defects.py
```diff
@@ -259,25 +262,43 @@ def toFitsRegionTable(self):
...
+
+            for i, defect in enumerate(self._defects):
+                box = defect.getBBox()
+                center = box.getCenter()
+                # Correct for the FITS 1-based offset
+                xCol.append(center.getX() + 1.0)
+                yCol.append(center.getY() + 1.0)
+
+                width = box.getWidth()
+                height = box.getHeight()
```
TallJimbo: I think you can use `box.width` and `box.height` now, too.

PR #192, opened 2020-04-09
DM-24404: Squash warnings from astropy's failure to close ecsv files.

TallJimbo:


python/lsst/meas/algorithms/defects.py
```diff
@@ -613,7 +614,12 @@ def readText(cls, filename):
         defects : `Defects`
             Defects read from a FITS table.
         """
-        table = astropy.table.Table.read(filename)
+        with warnings.catch_warnings():
+            # Squash warnings due to astropy failure to close files; we think
+            # this is a real problem, but the warnings are even worse.
+            # https://github.com/astropy/astropy/issues/8673
+            warnings.simplefilter("ignore")
```
timj, line 621: Is it possible to catch the specific warning but let others through?

pllim: Yes, it is possible. If you want, you can catch it in many different levels. I think catching `ResourceWarning` from `astropy.io.ascii` should be specific enough that you don't have to worry about other warnings being masked? We apologize for any inconvenience caused.

Ref: https://docs.python.org/3/library/warnings.html#describing-warning-filters

TallJimbo: Thanks for the tip; I've switched to only catching `ResourceWarning` from `astropy.io.ascii` and can confirm that it works.

PR #196, opened 2020-04-17
DM-24472: Regenerate Gaia DR2 catalogs to correct coordinate error fields

parejkoj:


parejkoj:
Good point: I've added a note to the Config. That field being in different units would show up when running `check_ingested_reference_catalog.py`, because the sources wouldn't get cross-matched. I hope nobody is producing catalogs with other units for ra/dec right now, but if so we can make that configurable when the time comes.

PR #202, opened 2020-05-07
DM-24556: example implementation

jdswinbank:
@plazas This implements my suggestion for DM-24556, as we discussed the other day. Please take a look and see what you think.


plazas:
Looks good to me too. Thank you!

jdswinbank:
Thanks both! @plazas, if you are happy with this approach, I suggest merging these commits into your `tickets/DM-24556` branch and then you can merge that to `master` when ready.

PR #204, opened 2020-05-14
DM-24935: Add support for DataFrame and Table in ScienceSourceSelectorTask

erykoff:


python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -102,19 +106,23 @@ def run(self, sourceCat, sourceSelectedField=None, matches=None, exposure=None):
         RuntimeError
             Raised if ``sourceCat`` is not contiguous.
         """
-        if not sourceCat.isContiguous():
-            raise RuntimeError("Input catalogs for source selection must be contiguous.")
+        if not isinstance(sourceCat, DataFrame) and not isinstance(sourceCat, Table):
```
parejkoj, line 109: I think you want this to be `if isinstance(sourceCat, lsst.afw.table.SimpleCatalog)`, since that is explicitly the one that needs a contiguity check.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -31,6 +31,8 @@
 import abc
 import numpy as np
 import astropy.units as u
+from pandas import DataFrame
+from astropy.table import Table
```
parejkoj, line 35: Please make this `import astropy.table` and use `astropy.table.Table` below: the `Table` name is too generic to be a top level import without a namespace. I'd almost say the same for `DataFrame`, but I think that only exists in pandas.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -219,8 +229,15 @@ def apply(self, catalog):
             Boolean array indicating for each source whether it is selected
             (True means selected).
         """
-        primary = (catalog[self.primary]*u.nJy).to_value(u.ABmag)
-        secondary = (catalog[self.secondary]*u.nJy).to_value(u.ABmag)
+        if isinstance(catalog, DataFrame) or isinstance(catalog, Table):
+            primary = np.array(catalog[self.primary])
+            secondary = np.array(catalog[self.secondary])
```
parejkoj, line 234: I think this needs a comment saying why we need to convert to numpy here. I have a guess, but I'm not positive, so it should be spelled out.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -250,12 +267,19 @@ def apply(self, catalog):
             (True means selected).
         """
         flagField = self.fluxField + "_flag"
-        if flagField in catalog.schema:
-            selected = np.logical_not(catalog[flagField])
+        if isinstance(catalog, DataFrame) or isinstance(catalog, Table):
+            if flagField in catalog.columns:
```
parejkoj, line 271: Is there any way to restructure this so that we can take the `np.ones()` blocks out of both `if`s? If we set `schema = catalog.schema if isinstance(catalog, SimpleCatalog) else catalog.columns` at the start, we could do one `if flagField in schema` at outside the `isinstance` check and I think this whole block would then be simpler.

Alternately, a lambda or closure like `isFieldInCatalog(field)` that encapsulates the `schema` vs. `columns` split.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -290,12 +314,20 @@ def apply(self, catalog):
...
```
parejkoj, line 318: Same comment here regarding the "is field in column" question.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -327,12 +359,22 @@ def apply(self, catalog):
...
```
parejkoj, line 363: And again, re: flag in column.

tests/test_sourceSelector.py
```diff
@@ -67,6 +67,18 @@ def check(self, expected):
         self.assertListEqual([src.getId() for src in results.sourceCat],
                              [src.getId() for src, ok in zip(self.catalog, expected) if ok])
 
+        # Check with DataFrame version of catalog
+        results = task.run(self.catalog.asAstropy().to_pandas())
+        self.assertListEqual(results.selected.tolist(), expected)
+        self.assertListEqual(list(results.sourceCat['id']),
+                             [src.getId() for src, ok in zip(self.catalog, expected) if ok])
+
+        # Check with Table version of catalog
```
parejkoj, line 76: "with astropy.table.Table", please

python/lsst/meas/algorithms/sourceSelector.py
erykoff: I'll add a comment.  It's because the DataFrame returns some sort of sequence that doesn't play nice with numpy operators unless you explicitly cast it.  I don't know why, it's a Pandas-is-Special thing.

python/lsst/meas/algorithms/sourceSelector.py
erykoff: I'll add a little `_` method, I guess it's used enough to be worthwhile.

erykoff:
If one could set an array of `Flag` fields in an `afwTable` then it would be _possible_ to convert a `DataFrame` to an `afwTable` such that one could use a `sourceSelector`, but you'd still need to work around string columns (which is another headache, but thankfully the `sourceSelector` doesn't need any of these).  But you'd still have the overhead of defining a schema, converting columns, etc.  It would be fast, but would lead to tens of lines of extra code.  Or make these modifications so that you can run the `sourceSelector` directly.

python/lsst/meas/algorithms/sourceSelector.py
laurenam: I am about to add the following to `utils.py` in `pipe_analysis`:
```
def getSchema(catalog):
    """Helper function to determine "schema" of catalog.

    This will be the list of columns if the catlaog type is a pandas DataFrame,
    or the schema object if it is an `lsst.afw.table.SourceCatalog`.
    """
    if isinstance(catalog, pd.DataFrame):
        schema = catalog.columns
    elif isinstance(catalog, pd.Series):
        schema = [catalog.name, ]
    else:
        schema = catalog.schema
    return schema
```

parejkoj: If we are planning to use pandas more broadly, that shouldn't go into pipe_analysis, but rather some lower level LSST package.

tests/test_sourceSelector.py
parejkoj: I think you missed this one in your cleanups?

parejkoj:
Thanks for the new method: that makes the main logic a lot clearer.

PR #205, opened 2020-05-27
DM-24830: Create dataset class for processed bright star stamps

MorganSchmitz:
One of 3 PRs for [DM-24830](https://jira.lsstcorp.org/browse/DM-24830) (the other two being obs_base and obs_subaru).

This is the main one, and contains both the class itself, and a few tests.

Jenkins run [#31959](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/31959/pipeline)

### DIFF
--- a/python/lsst/meas/algorithms/brightStarStamps.py
+++ b/python/lsst/meas/algorithms/brightStarStamps.py
@@ -1,0 +1,311 @@
+# This file is part of meas_algorithms.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Collection of small images (stamps), each centered on a bright star.
+"""
+
+__all__ = ["BrightStarStamp", "BrightStarStamps"]
+
+import collections.abc
+from typing import NamedTuple
+from enum import Enum, auto
+
+import lsst.afw.image as afwImage
+import lsst.afw.fits as afwFits
+from lsst.geom import Box2I, Point2I, Extent2I
+from lsst.daf.base import PropertySet
+
+
+class RadiiEnum(Enum):
+    INNER_RADIUS = auto()
+    OUTER_RADIUS = auto()
+
+    def __str__(self):
+        return self.name
+
+
+class BrightStarStamp(NamedTuple):
+    """Single stamp centered on a bright star, normalized by its
+    annularFlux.
+    """
+    starStamp: afwImage.maskedImage.MaskedImageF
+    gaiaGMag: float
+    gaiaId: int
+    annularFlux: float
+
+
+class BrightStarStamps(collections.abc.Sequence):
+    """Collection of bright star stamps and associated metadata.
+
+    Parameters
+    ----------
+    starStamps : `collections.abc.Sequence` [`BrightStarStamp`]
+        Sequence of star stamps.
+    innerRadius : `int`, optional
+        Inner radius value, in pixels. This and ``outerRadius`` define the
+        annulus used to compute the ``"annularFlux"`` values within each
+        ``starStamp``. Must be provided if ``"INNER_RADIUS"`` and
+        ``"OUTER_RADIUS"`` are not present in ``metadata``.
+    outerRadius : `int`, optional
+        Outer radius value, in pixels. This and ``innerRadius`` define the
+        annulus used to compute the ``"annularFlux"`` values within each
+        ``starStamp``. Must be provided if ``"INNER_RADIUS"`` and
+        ``"OUTER_RADIUS"`` are not present in ``metadata``.
+    metadata : `lsst.daf.base.PropertyList`, optional
+        Metadata associated with the bright stars.
+
+    Raises
+    ------
+    ValueError
+        Raised if one of the star stamps provided does not contain the
+        required keys.
+    AttributeError
+        Raised if the definition of the annulus used to compute each star's
+        normalization factor are not provided, that is, if ``"INNER_RADIUS"``
+        and ``"OUTER_RADIUS"`` are not present in ``metadata`` _and_
+        ``innerRadius`` and ``outerRadius`` are not provided.
+
+    Notes
+    -----
+    A (gen2) butler can be used to read only a part of the stamps,
+    specified by a bbox:
+
+    >>> starSubregions = butler.get("brightStarStamps_sub", dataId, bbox=bbox)
+    """
+
+    def __init__(self, starStamps, innerRadius=None, outerRadius=None,
+                 metadata=None,):
+        for item in starStamps:
+            if not isinstance(item, BrightStarStamp):
+                raise ValueError(f"Can only add instances of BrightStarStamp, got {type(item)}")
+        self._starStamps = starStamps
+        self._metadata = PropertySet() if metadata is None else metadata.deepCopy()
+        # Add inner and outer radii to metadata
+        self._checkRadius(innerRadius, RadiiEnum.INNER_RADIUS)
+        self._innerRadius = innerRadius
+        self._checkRadius(outerRadius, RadiiEnum.OUTER_RADIUS)
+        self._outerRadius = outerRadius
+
+    def __len__(self):
+        return len(self._starStamps)
+
+    def __getitem__(self, index):
+        return self._starStamps[index]
+
+    def __iter__(self):
+        return iter(self._starStamps)
+
+    def append(self, item, innerRadius, outerRadius):
+        """Add an additional bright star stamp.
+
+        Parameters
+        ----------
+        item : `BrightStarStamp`
+            Bright star stamp to append.
+        innerRadius : `int`
+            Inner radius value, in4 pixels. This and ``outerRadius`` define the
+            annulus used to compute the ``"annularFlux"`` values within each
+            ``starStamp``.
+        outerRadius : `int`, optional
+            Outer radius value, in pixels. This and ``innerRadius`` define the
+            annulus used to compute the ``"annularFlux"`` values within each
+            ``starStamp``.
+        """
+        if not isinstance(item, BrightStarStamp):
+            raise ValueError(f"Can only add instances of BrightStarStamp, got {type(item)}.")
+        self._checkRadius(innerRadius, RadiiEnum.INNER_RADIUS)
+        self._checkRadius(outerRadius, RadiiEnum.OUTER_RADIUS)
+        self._starStamps.append(item)
+        return None
+
+    def extend(self, bss):
+        """Extend BrightStarStamps instance by appending elements from another
+        instance.
+
+        Parameters
+        ----------
+        bss : `BrightStarStamps`
+            Other instance to concatenate.
+        """
+        self._checkRadius(bss._innerRadius, RadiiEnum.INNER_RADIUS)
+        self._checkRadius(bss._outerRadius, RadiiEnum.OUTER_RADIUS)
+        self._starStamps += bss._starStamps
+
+    def getMaskedImages(self):
+        """Retrieve star images.
+
+        Returns
+        -------
+        maskedImages :
+            `list` [`lsst.afw.image.maskedImage.maskedImage.MaskedImageF`]
+        """
+        return [stamp.starStamp for stamp in self._starStamps]
+
+    def getMagnitudes(self):
+        """Retrieve Gaia G magnitudes for each star.
+
+        Returns
+        -------
+        gaiaGMags : `list` [`float`]
+        """
+        return [stamp.gaiaGMag for stamp in self._starStamps]
+
+    def getGaiaIds(self):
+        """Retrieve Gaia IDs for each star.
+
+        Returns
+        -------
+        gaiaIds : `list` [`int`]
+        """
+        return [stamp.gaiaId for stamp in self._starStamps]
+
+    def getAnnularFluxes(self):
+        """Retrieve normalization factors for each star.
+
+        These are computed by integrating the flux in annulus centered on the
+        bright star, far enough from center to be beyond most severe ghosts and
+        saturation. The inner and outer radii that define the annulus can be
+        recovered from the metadata.
+
+        Returns
+        -------
+        annularFluxes : list[`float`]
+        """
+        return [stamp.annularFlux for stamp in self._starStamps]
+
+    def selectByMag(self, magMin=None, magMax=None):
+        """Return the subset of bright star stamps for objects with specified
+        magnitude cuts (in Gaia G).
+
+        Parameters
+        ----------
+        magMin : `float`, optional
+            Keep only stars fainter than this value.
+        magMax : `float`, optional
+            Keep only stars brighter than this value.
+        """
+        subset = [stamp for stamp in self._starStamps
+                  if (magMin is None or stamp.gaiaGMag > magMin)
+                  and (magMax is None or stamp.gaiaGMag < magMax)]
+        # This is an optimization to save looping over the init argument when
+        # it is already guaranteed to be the correct type
+        instance = BrightStarStamps((), metadata=self._metadata)
+        instance._starStamps = subset
+        return instance
+
+    @property
+    def metadata(self):
+        return self._metadata.deepCopy()
+
+    def _checkRadius(self, radiusValue, metadataEnum):
+        """Ensure provided annulus radius is consistent with that present
+        in metadata. If metadata does not contain annulus radius, add it.
+        """
+        # if a radius value is already present in metadata, ensure it matches
+        # the one given
+        metadataName = str(metadataEnum)
+        if self._metadata.exists(metadataName):
+            if radiusValue is not None:
+                if self._metadata[metadataName] != radiusValue:
+                    raise AttributeError("BrightStarStamps instance already contains different annulus radii "
+                                         + f"values ({metadataName}).")
+        # if not already in metadata, a value must be provided
+        elif radiusValue is None:
+            raise AttributeError("No radius value provided for the AnnularFlux measurement "
+                                 + f"({metadataName}), and none present in metadata.")
+        else:
+            self._metadata[metadataName] = radiusValue
+        return None
+
+    def writeFits(self, filename):
+        """Write a single FITS file containing all bright star stamps.
+        """
+        # ensure metadata contains current number of objects
+        self._metadata["N_STARS"] = len(self)
+
+        # add full list of Gaia magnitudes, IDs and annularFlxes to shared
+        # metadata
+        self._metadata["G_MAGS"] = self.getMagnitudes()
+        self._metadata["GAIA_IDS"] = self.getGaiaIds()
+        self._metadata["ANNULAR_FLUXES"] = self.getAnnularFluxes()
+
+        # create primary HDU with global metadata
+        fitsPrimary = afwFits.Fits(filename, "w")
+        fitsPrimary.createEmpty()
+        fitsPrimary.writeMetadata(self._metadata)
+        fitsPrimary.closeFile()
+
+        # add all stamps and mask planes
+        for stamp in self.getMaskedImages():
+            stamp.getImage().writeFits(filename, mode='a')
+            stamp.getMask().writeFits(filename, mode='a')
+        return None
+
+    @classmethod
+    def readFits(cls, filename):
+        """Read bright star stamps from FITS file.
+
+        Returns
+        -------
+        bss : `BrightStarStamps`
+            Collection of bright star stamps.
+        """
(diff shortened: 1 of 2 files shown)


python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+        # extract necessary info from metadata
+        visitMetadata = afw.fits.readMetadata(filename, hdu=0)
+        nbStarStamps = visitMetadata.getInt("N_STARS")
+        innerRadius = visitMetadata.get("INNER_RADIUS")
+        outerRadius = visitMetadata.get("OUTER_RADIUS")
+        gaiaGMags = visitMetadata.getArray("G_MAGS")
+        gaiaIds = visitMetadata.getArray("GAIA_IDS")
+        annularFluxes = visitMetadata.getArray("ANNULAR_FLUXES")
+        # check if a bbox was provided
+        kwargs = {}
```
natelust, line 260: Why are you creating a dict for one value?

tests/test_brightStarStamps.py
```diff
@@ -0,0 +1,143 @@
...
+class BrightStarStampsTestCase(lsst.utils.tests.TestCase):
+    """Test BrightStarStamps.
+    """
+    def setUp(self):
+        stampSize = (25, 25)
+        # create dummy star stamps
+        starImages = [afwImage.maskedImage.MaskedImageF(*stampSize)
+                      for _ in range(3)]
+        for starIm in starImages:
+            starImArray = starIm.getImage().getArray()
```
natelust, line 41: use the property access here:
```python
starImArray = starIm.image.array
```

tests/test_brightStarStamps.py
```diff
@@ -0,0 +1,143 @@
...
+            starImArray += np.random.rand(*stampSize)
```
natelust, line 42: If you are using rand in a test, you should set the seed at the top of the test file. This will produce exactly the same output each time, which is useful for tracking down errors.

tests/test_brightStarStamps.py
```diff
@@ -0,0 +1,143 @@
...
+        del self.faintObjIdx
+
+    def testIO(self):
+        """Test the class' write and readFits methods.
+
+        The ``options`` argument to the read method is only used to read
+        sub-BBoxes, which is handled by the Butler. Tests of this are done in
+        afw.
+        """
+        filename = "brightstarStamps.fits"
```
natelust, line 70: You should use the temp file context manager [here](https://docs.python.org/3/library/tempfile.html). This will create you a temporary file in the appropriate place for a given os, and make sure it is cleaned up at the end.

tests/test_brightStarStamps.py
```diff
@@ -0,0 +1,143 @@
...
+        self.assertEqual(len(brightOnly), 2)
+        self.assertFalse("faint" in brightOnly.getGaiaIds())
+
+        faintOnly = self.bss.selectByMag(magMin=7)
+        self.assertEqual(len(faintOnly), 1)
+        self.assertEqual(faintOnly.getGaiaIds()[0], "faint")
+        brightObj = self.bss[self.faintObjIdx]
+        self.assertMaskedImagesAlmostEqual(brightObj["starStamp"], faintOnly.getMaskedImages()[0])
+
+    def testMissingKeyHandling(self):
```
natelust, line 94: Using the record type for storage I talked about above, this test will change.

tests/test_brightStarStamps.py
```diff
@@ -0,0 +1,143 @@
...
+        _ = brightStarStamps.BrightStarStamps(self.starStamps, innerRadius=self.innerRadius,
+                                              outerRadius=self.outerRadius,
+                                              metadata=metadata)
+        # An exception should be raised if they do not
+        with self.assertRaises(AttributeError):
+            _ = brightStarStamps.BrightStarStamps(self.starStamps, innerRadius=self.innerRadius/2,
+                                                  metadata=metadata)
+        with self.assertRaises(AttributeError):
+            _ = brightStarStamps.BrightStarStamps(self.starStamps, outerRadius=self.outerRadius + 1,
+                                                  metadata=metadata)
```
natelust, line 130: What about testing supplying the radius if it does not exist in the metadata, which is probably the only time the arguments will be used.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Collection of small images (stamps), each centered on a bright star.
+"""
+
+import collections.abc
+import numpy as np
+
+from lsst import afw
```
natelust, line 28: In general we dont import just afw. You will more commonly see
```python
import lsst.afw.fits as afwFits
```

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+from lsst import geom
```
natelust, line 29: because you are using very few symbols, its probably worth the saved characters later to do
```python
from lsst.geom import Box2I, Point2I, Extent2I
```

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+        gaiaGMags : `list` [`float`[
+        """
+        return [stamp["gaiaGMag"] for stamp in self._starStamps]
+
+    def getGaiaIds(self):
+        """Retrieve Gaia IDs for each star.
+
+        Returns
+        -------
+        gaiaIds : `list` [`str`]
```
natelust, line 131: Can I ask why str here?

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+        # if a radius value is already present in metadata, ensure it matches
+        # the one given
+        if self._metadata.exists(metadataName):
+            if radiusValue is not None:
+                if self._metadata.get(name=metadataName) != radiusValue:
+                    raise AttributeError("BrightStarStamps instance already contains different annulus radii "
+                                         + f"values ({metadataName}).")
+        # if not already in metadata, a value must be provided
+        elif radiusValue is None:
+            raise AttributeError("No radius value provided for the AnnularFlux measurement "
```
natelust, line 212: This msg should say something about no value provided, and none present in the metadata, so the user can track down that there are two different places it could come from

natelust: To be honest I am not entirely sure why this exists, the user must already supply the meta data field, and the calculations for the stamps must be done prior to creating this class (I guess append could add some if you had an empty list) The radii should be known at the point of creating this class then. Why have the arguments instead of just enforcing they are in the meta data passed in. Is it just a convenience for the user? I think the class would be simpler if the radii just came in with the meta data. Its a bit confusing to have different ways of setting it, the user might not know which one to use, or think they must use both.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+from lsst import geom
+from lsst.daf.base import PropertySet
+
+
+class BrightStarStamps(collections.abc.Sequence):
+    """Collection of bright star stamps and associated metadata.
+
+    Parameters
+    ----------
+    starStamps : `list` [`dict`]
```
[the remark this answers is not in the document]
MorganSchmitz: Any MutableSequence would do. Changed.

Note I'm not sure if this should be `abc.MutableSequence` in the docstring then, since we do not import `collections.abc`. Is it okay to reference an unimported library in a docstring?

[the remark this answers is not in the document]
natelust: I think sphinx does its own imports and look-ups. MutableSequence is a derived class of Sequence, and as you are not using any mutability aspect of the input, it is better to tell people you just need a sequence (i.e. someone could pass a tuple and it would work just fine)

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+        magMax : `float`, optional
+            Keep only stars brighter than this value.
+        """
+        selector = np.array([True for _ in range(len(self))])
+        if magMin is not None:
+            selector *= np.array(self.getMagnitudes()) > magMin
+        if magMax is not None:
+            selector *= np.array(self.getMagnitudes()) < magMax
+        selIdx = np.where(selector)[0]
+        return BrightStarStamps([self._starStamps[j] for j in selIdx], metadata=self._metadata)
```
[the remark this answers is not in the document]
MorganSchmitz: This is great. I learned so much from this - thanks Nate!

I went with _all_ of your suggestions, including folding in the retrieval of magnitudes to avoid the GetMagnitudes loop. I think it still makes for readable code - at least much more so than my initial 13 loops.

python/lsst/meas/algorithms/brightStarStamps.py
MorganSchmitz: I can't rightly say... Fixed.

MorganSchmitz: Ah! Yes I can, though admittedly there's probably a much better way to go about this.

It is so that `kwargs` can be passed along to `ImReader` and `maskReader` below (lines 273-274), regardless of whether a bbox was specified. If not, it's an empty dictionary.

python/lsst/meas/algorithms/brightStarStamps.py
MorganSchmitz: Right, this is confusing as I tried to take into account how these objects will be generated and used. There are a lot of disparate elements that led me to write things this way:
* These `starStamps` objects have already been normalized, using their annularFlux value, for set values of annular radii. It only makes sense to store them together within one `BrightStarStamps` instance if they have all used the same annulus definition.
* We will want to write a great many of these to disk; we need to include the radii used in the written fits. It seems the best way to do so is to add these values to the metadata, that will then be set within the `writeFits` method (line 233 in this version of the file).
* As discussed in the thread about the metadata getter, `self._metadata` is likely to be the full metadata of the `Exposure` from which the stamps were extracted, with extra fields added here. The “user” who creates these instances (DM-25304) has this readily available on the one hand, and the radii values he used on the other. I thought it would make sense to have the class itself merge the two internally, so this user only needs to dump the exposure metadata, and the two radii values, at initialization.
* Another use case will be the reading, concatenation, and further use (DM-25305) of several BrightStarStamps, saved to disk. We very much want to ensure we are concatenating bright star stamps that used the same annular flux radii; but since we’re reading them from disk, that information is in the metadata now.

I do think the duplication here is good, or at least necessary: we want users to be able to provide their annular radii when they create a new instance, and dump metadata at the same time. If that metadata comes from another BrightStarStamps (likely read from disk), right then is a good time to fail if the radii used do not match those in the metadata. Otherwise, they can safely be added.

Having written all this, however, I realize this would perhaps look a little less out-of-place with an `extend` method, so I added one. This is what will be used for concatenation in DM-25304, rather than `append`; and it makes it clearer (I think) why we want this `_checkRadius` there. This led me to shuffling things around a little; `_checkRadius` is called directly in the init now, and the two radii are also kept as private members.

Does that make sense?

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,280 @@
...
+        """Retrieve metadata associated with these `Defects`.
+
+        Returns
+        -------
+        meta : `lsst.daf.base.PropertyList`
+            Metadata. The returned `~lsst.daf.base.PropertyList` can be
+            modified by the caller and the changes will be written to
+            external files.
+        """
+        return self._metadata
```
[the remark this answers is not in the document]
MorganSchmitz: My reasoning for having it as a private member with an explicit getter, which might very well not be sound, is the following.

The metadata of the class absolutely needs to contain some "new" information (annular radii especially, but also the Gaia magnitudes and ids), specific to the bright star stamps. In many use cases, it might also contain the metadata from the visit from which the stamps were extracted. 

My intention was for this to be transparent to the user. When these objects are created (DM-25304), the user can just pass on the exposure metadata unchanged, along with the annular radii they set when processing the stamps, and be sure the metadata will be written along with the stamps and new extra information.

Conversely, I was hoping this design would strongly imply the user probably should not alter the metadata manually.

A very similar design is used for defects: https://github.com/lsst/meas_algorithms/blob/b97ba8bf064112940b133beb7b5d8f5936140fec/python/lsst/meas/algorithms/defects.py#L219

python/lsst/meas/algorithms/brightStarStamps.py
MorganSchmitz: I think I figured it made sense for unique ids to be strings, even if (as seems to be the case here with Gaia DR2) they happen to be ints. Now that I actually take a step back and think about it, I cannot really justify it, I suppose.

Would you recommend specifying int, or not providing the type at all?

python/lsst/meas/algorithms/brightStarStamps.py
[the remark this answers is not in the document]
natelust: Having the object contain new information or not makes no difference to the access pattern of the instance attribute, you could still have `.metadata` even if you mutate it. An attribute is just a flatter way to access data contained by the class. However your comment does bring up some things:
* Since you are mutating the metadata passed into the class, you really should create a copy of it in `__init__`, since you don't know if your changes will affect something down stream
* If you want the user to not modify the metadata, it is better to return a copy of your internal metadata object in a getter
* but if you really want to make it read only in a pythonic way, you should keep it as a private attribute `_metadata` and have a metadata property like I demoed above. This will prevent someone from reassigning the `.metadata` to anything. If it is important that someone not assign to the property set you should also return a copy in the property, so that they can use values but any changes they make do not affect your class.. That will allow your class to have an interface such that someone could write `brightStars.metadata['some_property']` while still keeping the safety you want.

Putting that together would look something like:
```python
class BrightStarStamps:
    def __init__(..., metadata, ...):
        ...
        self._metadata = metadata.copy()
        ...
    @property
    def metadata(self):
        return self._metadata.copy()
```
Note copy is a shallow copy, so the container is new, and the keys are new, but the values are not duplicated. The amount of copying for protection is up to you, but as I suspect this is not a rapidly accessed member speed is probably not a concern.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,274 @@
...
+
+    def _checkRadius(self, radiusValue, metadataName):
+        """Ensure provided annulus radius is consistent with that present
+        in metadata. If metadata does not contain annulus radius, add it.
+        """
+        # if a radius value is already present in metadata, ensure it matches
+        # the one given
+        if self._metadata.exists(metadataName):
+            if radiusValue is not None:
+                if self._metadata.get(name=metadataName) != radiusValue:
```
natelust, line 203: use the `__getitem__` interface:
```python
if self._metadata[metadataName] != radiusValue:
```
Change this throughout the code

python/lsst/meas/algorithms/brightStarStamps.py
natelust: Yes it does. Its really just a question of who is responsible for adding these values to the metadata. If you want the class to handle it in the case of arguments as a convenience to the user, that's fine. I do like the addition of the extend method. However, in regards to your first built point, you actually have no mechanism in place to verify that the radii are the same in the append method. You should either require radii as args, put a big warning in the method docstring, or require adding elements be done by calling extend with a new instance.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,274 @@
...
+                    raise AttributeError("BrightStarStamps instance already contains different annulus radii "
+                                         + f"values ({metadataName}).")
+        # if not already in metadata, a value must be provided
+        elif radiusValue is None:
+            raise AttributeError("No radius value provided for the AnnularFlux measurement "
+                                 + f"({metadataName}), and none present in metadata.")
+        else:
+            self._metadata.add(name=metadataName, value=radiusValue)
```
natelust, line 211: use `__setitem__` interface as mentioned above

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,274 @@
...
+    def extend(self, bss):
+        """Extend BrightStarStamps instance by appending elements from another
+        instance.
+
+        Parameters
+        ----------
+        bss : `BrightStarStamps`
+            Other instance to concatenate.
+        """
+        self._checkRadius(bss._innerRadius, "INNER_RADIUS")
```
natelust, line 117: Though in the case of this class it is rather limited in use, it is best to not use magic strings if it can be avoided. Magic strings are susceptible to typos by someone at some future point, which are impossible to lint, which leads to an error not being found until some future runtime error.

If you can enumerate all the possibilities (such as this case) it is better to use pythons Enum class like follows

```python
from enum import Enum, auto

class RadiiEnum(Enum):
    INNER_RADIUS = auto()  # auto because we dont actually care what the enumeration number is
    OUTER_RADIUS = auto()
```
This allows you to use and check like follows
```python
self._checkRadius(bss._outerRadius, RadiiEnum.INNER_RADIUS)
....
    def _checkRadius(self, radiusValue, metadataEnum):
        if metadataEnum is RadiiEnum.INNER_RADIUS:
            ....
        else:
            ....
```
Linters will then be able to verify everything is correct every time it is used. However we can do one better for your case:
```python
class RadiiEnum(Enum):
    INNER_RADIUS = auto()
    OUTER_RADIUS = auto()

    def __str__(self):
        return self.name #  name is the name of what ever variant the value happened to be
```
This will turn the enum into a string with the correct name when you use it as such:
```python
    def _checkRadius(self, radiusValue, metadataEnum):
        metadataName = str(metadataEnum)
       #  The rest of your function can now remain the same
```
This way users can just use the enum values which will be checked by the linter, editors can auto complete, and it is self documenting in that all the possible options are enumerated (you know that their isnt some other metadata field you could check). The only failure can then be if you spell your enum name wrong, and that can be checked by a simple unit test.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,274 @@
...
+        # if not already in metadata, a value must be provided
+        elif radiusValue is None:
+            raise AttributeError("No radius value provided for the AnnularFlux measurement "
+                                 + f"({metadataName}), and none present in metadata.")
+        else:
+            self._metadata.add(name=metadataName, value=radiusValue)
+        return None
+
+    def writeFits(self, filename):
+        """Write a single FITS file containing all bright star stamps.
```
natelust, line 215: use the get/set item interfaces of property set where appropriate

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,295 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+#
+"""Collection of small images (stamps), each centered on a bright star.
+"""
+
+__all__ = ["BrightStarStamp", "BrightStarStamps"]
```
natelust, line 25: I know you will not see this in a lot of code we have, because people often get this wrong (reviewers included) but __all__ needs to be the first thing in the file, before the copyright

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -0,0 +1,274 @@
+# This file is part of meas_algorithms.
```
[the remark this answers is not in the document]
timj: The `__all__` should be after the license according to the style guide: https://developer.lsst.io/python/style.html?highlight=__all__#standard-code-order-should-be-followed

python/lsst/meas/algorithms/brightStarStamps.py
timj: No. https://developer.lsst.io/python/style.html?highlight=__all__#standard-code-order-should-be-followed 

To be specific:

* Shebang line, #! /usr/bin/env python (only for executable scripts)
* Module-level comments (such as the license statement)
* Module-level docstring
* `__all__ = [...]` statement, if present

python/lsst/meas/algorithms/brightStarStamps.py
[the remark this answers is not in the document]
natelust: I could have sworn I had read that differently not that long ago, but git history tells me otherwise, I feel as if I am loosing my memory or something. Thanks @timj for pointing that out.

python/lsst/meas/algorithms/brightStarStamps.py
TallJimbo: This is getting further off topic, but I notice that `__future__` imports aren't in that list in the dev guide, and I've been assuming they need to go before `__all__`.  Anyone know if that's correct?

natelust: Jim if you are thinking about the annotation future import, it MUST be the absolute first thing in a file, because it changes how the interpreter works.

timj: I'm pretty sure python complains if the `__future__` is not the first line of code since the future applies to the entire file.

natelust: Other future things may or may not be safe to be put other places.

natelust: and by first, I mean before __all__ or any other executed code, things like docstring and comments are not interpreted and or parsed, so it does not really matter

natelust: In case you care to know why, here is the page on future https://docs.python.org/3/library/__future__.html. basically there can be a CompilerFlag that changes how code is compiled into python byte code.

PR #206, opened 2020-06-17
DM-24277: Apply proper motion and parallax while loading refcats in Jointcal

parejkoj:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -1458,12 +1457,12 @@ def applyProperMotionsImpl(log, catalog, epoch):
 
     Adjust position and position error in the ``catalog``
     for proper motion to the specified ``epoch``,
-    modifying the catalong in place.
+    modifying the catalog in place.
 
     Parameters
     ----------
     log : `lsst.log.log`
```
PaulPrice, line 1464: `lsst.log.Log`

PR #208, opened 2020-07-16
DM-25979: IngestIndexReferenceTask throws if coord_err_unit is None

jdswinbank:
AKA “that looks more fun than proof rereading slides”.

### DIFF
--- a/tests/nopytest_ingestIndexReferenceCatalog.py
+++ b/tests/nopytest_ingestIndexReferenceCatalog.py
@@ -19,6 +19,13 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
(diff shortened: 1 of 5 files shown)


tests/nopytest_ingestIndexReferenceCatalog.py
```diff
@@ -68,6 +70,10 @@ def testIngestTwoFilesTwoCores(self):
         self.checkAllRowsInRefcat(loader, skyCatalog1, config)
         self.checkAllRowsInRefcat(loader, skyCatalog2, config)
 
+    def testIngestTwoFilesTwoCores(self):
+        for withRaDecErr in [True, False]:
+            self._run_testIngestTwoFilesTwoCores(withRaDecErr)
```
parejkoj, line 75: Can you please make this two separate calls? That way, if one fails, it will be much more obvious which one.

PR #209, opened 2020-07-18
DM-26007: defaultFilter is not used if a filterName is given to loadSkyCircle

parejkoj:


tests/test_loadReferenceObjects.py
```diff
@@ -121,8 +121,12 @@ def testFilterAliasMap(self):
                 # camera filters aliases are named <filter>_camFlux
                 if "camr" in filterMap:
                     self.assertEqual(getRefFluxField(refSchema, "camr"), "camr_camFlux")
+                # If defaultFilter is specified, use that instead if there's no mapping.
```
r-owen, line 124: Please clarify this comment. What is "that"? I suggest moving the comment and changing it to something along the lines of the following (assuming it is obvious that this test has no filterMap):
```
elif defaultFilter != "":
    # defaultFilter specified but we have no filterMap;
    # The chosen filter should be "camFlux"
    self.assertEqual(...)
else:
    # ... (please describe this case)
```

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -719,7 +722,9 @@ def getRefFluxField(schema, filterName=None):
     if not isinstance(schema, afwTable.Schema):
         raise RuntimeError("schema=%s is not a schema" % (schema,))
     if filterName:
-        fluxFieldList = [filterName + "_camFlux", filterName + "_flux"]
+        # NOTE: include "camFlux" in the list for the case where
+        # `defaultFilter` was specified in the refcat loader config.
+        fluxFieldList = [filterName + "_camFlux", filterName + "_flux", "camFlux"]
```
r-owen, line 727: This code will silently use "camFlux" even if you have a filter map, but you misspell `filterName` or the associated entry in the filter map. That worries me.

If there is a reasonable way to test if you have a filter map, then I suggest the following, instead:
```
if filterName and a filter map exists:
    fluxFieldList = [filterName + "_camFlux", filterName + "_flux"]
else:
    fluxFieldList = ["camFlux"]
```

tests/test_loadReferenceObjects.py
```diff
@@ -149,9 +153,13 @@ def testFilterAliasMap(self):
                 if "camr" in filterMap:
                     fluxKey, fluxErrKey = getRefFluxKeys(refSchema, "camr")
                     self.assertEqual(refCat[0].get(fluxErrKey), 0.111)
+                # If defaultFilter is specified, use that instead if there's no mapping.
```
r-owen, line 156: I have the same comment about this comment as immediately above.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -1076,6 +1103,14 @@ def addAliasesForOneFilter(filterName, refFilterName):
                 camFluxErrName = camFluxName + "Err"
                 aliasMap.set(camFluxErrName, refFluxErrName)
 
+        if self.config.anyFilterMapsToThis is not None:
+            refFluxName = self.config.anyFilterMapsToThis + "_flux"
+            if refFluxName not in schema:
+                msg = f"Unknown reference filter for anyFilterMapsToThis='{refFluxName}'"
+                raise RuntimeError(msg)
+            aliasMap.set("anyFilterMapsToThis", refFluxName)
+            return  # this is mutually exclusive with filterMap
```
r-owen, line 1112: Please consider moving this code block before the definition of addAliasesForOneFilter, since this code block doesn't use that function. I found it confusing to read through a function definition that was not used.

PR #218, opened 2020-11-06
DM-27364: Initialize counter before forking for python 3.8 support on mac

brianv0:
Fixes python 3.8 multiprocessing start method of `spawn` on macOS.

### DIFF
--- a/python/lsst/meas/algorithms/ingestIndexManager.py
+++ b/python/lsst/meas/algorithms/ingestIndexManager.py
@@ -36,9 +36,9 @@ from lsst.afw.image import fluxErrFromABMagErr
 
 # global shared counter to keep track of source ids
 # (multiprocess sharing is most easily done with a global)
-COUNTER = 0
+COUNTER = multiprocessing.Value('i', 0)
 # global shared counter to keep track of number of files processed.
-FILE_PROGRESS = 0
+FILE_PROGRESS = multiprocessing.Value('i', 0)
 
 
 class IngestIndexManager:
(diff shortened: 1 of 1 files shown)


python/lsst/meas/algorithms/ingestIndexManager.py
```diff
@@ -36,9 +36,9 @@
 
 # global shared counter to keep track of source ids
 # (multiprocess sharing is most easily done with a global)
-COUNTER = 0
+COUNTER = multiprocessing.Value('i', 0)
 # global shared counter to keep track of number of files processed.
-FILE_PROGRESS = 0
+FILE_PROGRESS = multiprocessing.Value('i', 0)
```
parejkoj, line 41: Could you use the ctypes type (I think `c_int`, or maybe just `int`) instead of `'i'`` here? The `'i'` made me think that was the name of the counter or something.

parejkoj (commented):
Thanks for catching this.

Please fix the PR name to match the `DM-NNNN: {{JIRA Ticket Title}}` format specified in the dev guide.

https://developer.lsst.io/work/flow.html#make-a-pull-request

python/lsst/meas/algorithms/ingestIndexManager.py
```diff
@@ -21,6 +21,7 @@
 
 __all__ = ["IngestIndexManager", "IngestGaiaManager"]
 
+from ctypes import c_int
```
parejkoj, line 24: I would suggest `import ctypes` and then using it explicitly below. It only shows up twice in the code, so it's not like a `from ...` import is really saving any typing.

parejkoj:
Don't forget the PR name change...

PR #219, opened 2020-11-11
DM-27501: Make a generic Stamps class

SimonKrughoff:


python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,209 @@
...
+    >>> starSubregions = butler.get("brightStarStamps_sub", dataId, bbox=bbox)
+    """
+
+    def __init__(self, stamps, metadata=None, has_mask=True, has_variance=True):
+        for item in stamps:
+            if not isinstance(item, Stamp):
+                raise ValueError(f"Can only add instances of Stamp, got {type(item)}")
+        self._stamps = stamps
+        self._metadata = PropertySet() if metadata is None else metadata.deepCopy()
+        self._metadata['HAS_MASK'] = has_variance
```
ktlim, line 83: Bug here.  It's better to store metadata as actual attributes if possible rather than generic FITS headers; it helps to avoid bugs like this.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,209 @@
...
+
+    def __getitem__(self, index):
+        return self._stamps[index]
+
+    def __iter__(self):
+        return iter(self._stamps)
+
+    def _refresh_metadata(self):
+        # ensure metadata contains current number of objects
+        self._metadata["N_STAMPS"] = len(self)
```
ktlim, line 97: I wouldn't think that this method should be necessary at all, if `N_STAMPS` is generated only when necessary.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -42,17 +39,24 @@ def __str__(self):
...
+    def factory(cls, stamp, metadata, idx):
+        gaiaGMags = metadata.getArray("G_MAGS")
+        gaiaIds = metadata.getArray("GAIA_IDS")
+        annularFluxes = metadata.getArray("ANNULAR_FLUXES")
+        return cls(stamp=stamp, ra=Angle(numpy.nan), dec=Angle(numpy.nan), size=-1,
+                   gaiaGMag=gaiaGMags[idx], gaiaId=gaiaIds[idx], annularFlux=annularFluxes[idx])
+
 
-class BrightStarStamps(collections.abc.Sequence):
+class BrightStarStamps(Stamps):
```
ktlim, line 59: It's not clear that `Stamps` is giving you much except a generic metadata element (that I think may be better avoided) and `readFits()` and `writeFits()` methods.  You might be better off having this be explicitly based on a `list` of `BrightStarStamp` with additional attributes and then having external `readFits()` and `writeFits()` methods that can handle the list (and may need to be overridden anyway in this class).

SimonKrughoff: I'm specifically trying to not have the `readFits` and `writeFits` methods overridden in this class.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,335 @@
...
+    def factory(cls, stamp, metadata, index):
+        """A factory method to construct an instance of this class
+        given a masked image and some metadata.
+
+        Parameteres
+        -----------
+        stamp : `lsst.afw.image.MaskedImageF`
+            The pixel data to associate with the instance
+        metadata : `dict`
+            A mapping of metadata used to populate this instance
```
timj, line 153: It's a dict that has three required keys and whose values must be a sequence.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,335 @@
...
+        index : `int`
+            The index value (0 based) for looking up values from ``metadata``
+
+        Returns
+        -------
+        stamp : `lsst.meas.algorithms.Stamp`
+            An instance of this class
+        """
+        if 'RA_DEG' in metadata.keys() and 'DEC_DEG' in metadata.keys() and 'SIZE' in metadata.keys():
```
timj, line 162: Please drop `.keys()` from here since they are not needed.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,335 @@
...
+    def readFitsWithOptions(cls, filename, options):
+        """Build an instance of this class with options.
+
+        Parameters
+        ----------
+        filename : `str`
+            Name of the file to read
+        options : `PropertySet`
+            Collection of metadata parameters
+        """
```
timj, line 320: Missing what it returns.

timj (commented):
Here are some initial comments. It's taking me a long time to get through it so felt I should send what I already have.  I am worried about the fragility of the FITS file we end up with and I'm also worried that we use "stamps" lower case and "Stamps" uppercase and the lowercase one is sometimes referring to a MaskedImage and not a "Stamp".

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -92,26 +112,67 @@ class BrightStarStamps(collections.abc.Sequence):
...
         self._checkRadius(outerRadius, RadiiEnum.OUTER_RADIUS)
         self._outerRadius = outerRadius
 
-    def __len__(self):
-        return len(self._starStamps)
+    def _refresh_metadata(self):
+        """Refresh the metadata.  Should be called before writing this object out.
+        """
+        # ensure metadata contains current number of objects
+        self._metadata["N_STAMPS"] = len(self)
```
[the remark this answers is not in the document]
SimonKrughoff: OK.  I got rid of _refresh_metadata in the base class too

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,335 @@
...
+        A string indicating the output filename
+    stamps : iterable of `lsst.afw.image.MaskedImageF`
+        An iterable of maked images
+    metadata : `PropertySet`
+        A collection of key, value metadata pairs
+    """
+    # create primary HDU with global metadata
+    fitsPrimary = afwFits.Fits(filename, "w")
+    fitsPrimary.createEmpty()
+    fitsPrimary.writeMetadata(metadata)
```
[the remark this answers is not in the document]
SimonKrughoff: I set it with `len(stamps)` immediately before writing it.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,335 @@
...
+
+
+class StampsBase(abc.ABC, Sequence):
+    """Collection of  stamps and associated metadata.
+
+    Parameters
+    ----------
+    stamps : iterable
+        This should be an iterable of dataclass objects
+        a la ``lsst.meas.algorithms.Stamp``.
```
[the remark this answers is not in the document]
SimonKrughoff: I am calling the variable `stamp_im` or `stamp_ims` if it's a MaskedImage or list of them.  Otherwise I call them `stamp` or `stamps`.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,335 @@
...
+        kwargs["bbox"] = bbox
+    # read stamps themselves
+    stamps = []
+    for idx in range(nStamps):
+        n_parts = 1  # Always have at least an image
+        for key in ['HAS_MASK', 'HAS_VARIANCE']:
+            if metadata[key]:
+                n_parts += 1
+        parts = {}
+        imReader = afwImage.ImageFitsReader(filename, hdu=n_parts*idx + 1)
```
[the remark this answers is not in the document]
SimonKrughoff: I believe I have made the change appropriately.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,335 @@
...
+        if metadata['HAS_MASK']:
+            # Alwasy stored right after image if present
+            maskReader = afwImage.MaskFitsReader(filename, hdu=n_parts*idx + 2)
+            parts['mask'] = maskReader.read(**kwargs)
+        if metadata['HAS_VARIANCE']:
+            # Maybe either right after image or two after
+            varReader = afwImage.ImageFitsReader(filename, hdu=n_parts*idx + n_parts)
+            parts['variance'] = varReader.read(**kwargs)
+        maskedImage = afwImage.MaskedImageF(**parts)
+        stamps.append(stamp_factory(maskedImage, metadata.toDict(), idx))
```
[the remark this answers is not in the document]
SimonKrughoff: I believe so, otherwise I can't increment into the array associated with a particular key.  Though I certainly call it just once.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: I think I've addressed this, but let me know if it's not complete enough

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+        stamp objects a la `lsst.meas.alrogithm.Stamp`
+    options : `PropertySet`
+        A collection of parameters.  If certain keys are available
+        (``llcX``, ``llcY``, ``width``, ``height``), a bounding box
+        is constructed and passed to the ``FitsReader`` in order
+        to return a sub-image.
+
+    Returns
+    -------
+    stamps, metadata : `list` of dataclass objects like `Stamp`, PropertySet
```
timj, line 99: I think we write this as 

```
stamps : `list` of ...
    Some stamps whatever
metadata : `PropertySet
   The metadata
```
rather than a single line.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+        else:
+            raise ValueError(f"Unknown extension type: {md['EXTNAME']}")
+        stamp_parts.setdefault(md['EXTVER'], {})[md['EXTNAME'].lower()] = reader.read(**kwargs)
+        idx += 1
+    # construct stamps themselves
+    stamps = []
+    meta_dict = metadata.toDict()
+    for k in stamp_parts:
+        maskedImage = afwImage.MaskedImageF(**stamp_parts[k])
+        # Assume the value of EXTVER is the index into the metadata
```
timj, line 131: EXTVER starts at 1 and not 0. Does that matter?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+        is constructed and passed to the ``FitsReader`` in order
+        to return a sub-image.
+
+    Returns
+    -------
+    stamps, metadata : `list` of dataclass objects like `Stamp`, PropertySet
+        A tuple of a list of `Stamp`-like objects and a collection of metadata.
+    """
+    # extract necessary info from metadata
+    metadata = afwFits.readMetadata(filename, hdu=0)
```
timj, line 103: Isn't this a PropertyList and not a PropertySet?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+            reader = afwImage.ImageFitsReader(filename, hdu=idx)
+        elif md['EXTNAME'] == 'MASK':
+            reader = afwImage.MaskFitsReader(filename, hdu=idx)
+        else:
+            raise ValueError(f"Unknown extension type: {md['EXTNAME']}")
+        stamp_parts.setdefault(md['EXTVER'], {})[md['EXTNAME'].lower()] = reader.read(**kwargs)
+        idx += 1
+    # construct stamps themselves
+    stamps = []
+    meta_dict = metadata.toDict()
```
timj, line 128: Maybe comment here to say that indexing into a PropertyList has different semantics?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+        Returns
+        -------
+        stamp : `Stamp`
+            An instance of this class
+        """
+        if 'RA_DEG' in metadata and 'DEC_DEG' in metadata and 'SIZE' in metadata:
+            return cls(stamp_im=stamp_im,
+                       position=SpherePoint(Angle(metadata['RA_DEG'][index], degrees),
+                                            Angle(metadata['DEC_DEG'][index], degrees)),
+                       size=metadata['SIZE'][index])
```
timj, line 185: Why can't this be worked out from the masked image?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+    metadata : `lsst.daf.base.PropertyList`, optional
+        Metadata associated with the bright stars.
+    use_mask : `bool`, optional
+        If ``True`` read and write the mask data.  Default ``True``.
+    use_variance : `bool`, optional
+        If ``True`` read and write the variance data. Default ``True``.
+
+    Notes
+    -----
+    A (gen2) butler can be used to read only a part of the stamps,
```
timj, line 207: so can a gen3 butler.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def extend(self, s):
+        """Extend Stamps instance by appending elements from another instance.
+
+        Parameters
+        ----------
+        s : `list` [`object`]
+            List of Stamp-like object to append.
+        """
+        self._stamps += s._stamps
```
timj, line 288: Don't you want to check that you are not appending random objects?  Don't you need to check that the thing even has a `_stamps` property and tell people something reasonably rather than triggering an AttributeError?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+
+    def append(self, item):
+        """Add an additional bright star stamp.
+
+        Parameters
+        ----------
+        item : `object`
+            Stamp-like object to append.
+        """
+        if not hasattr(item, 'stamp'):
```
timj, line 275: Would it help if Stamp and BrightStarStamp inherited from AbstractStamp? Then you could do an isinstance check.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -0,0 +1,351 @@
...
+        Returns
+        -------
+        maskedImages :
+            `list` [`lsst.afw.image.maskedImage.maskedImage.MaskedImageF`]
+        """
+        return [stamp.stamp_im for stamp in self._stamps]
+
+    @property
+    def metadata(self):
+        return self._metadata.deepCopy()
```
timj, line 302: You should think about whether the deep copy is the right thing to do. It makes it hard for people outside to change anything inside. For afw the getMetadata returns the actual thing that you can then update. If your use case is to tightly control this metadata then that's fine but maybe worth adding a docstring here.

timj (approved):
Looks much better. Thanks.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: Oh, good catch.  I think it may.  Let me look.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: Maybe I should assert that the size of the postage stamp is the same as the size that was asked for?

timj: I'm not sure why size needs to be specified in the constructor. Maybe I've misunderstood that metadata field.

Checking that the metadata is consistent with the masked image seems reasonable, but if they are meant to be the same then there is no need to pass it in to the constructor.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: I just made EXTVER start at one so it would line up with the extension number.  I think that's a bad idea since it will also be used as an index into an array of values for building the stamps.  I'm changing to make it zero based

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: I frankly don't know the importance of that difference.  I can check.

python/lsst/meas/algorithms/stamps.py
timj: I have no idea if 0 is allowed for EXTVER. You may have to read the standard a bit harder to see if 0 is allowed but one option is to create the file starting at 0 and see if fitsverify is happy with it.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: OK

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: Will do

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: That's initially how I started, but then the two didn't share any attributes so the only thing that the abstract class would have in it would be an abstract `factory` class method.  I'll do that and you can tell me if you like it or not.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: I don't know what the use case @MorganSchmitz was trying solve was.  I'll try to close the loop on that.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: OK.  The standard specifically says it must be > 0.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: My understanding is that is how `dataclass` objects work, maybe I can give it a default so I can omit it from the constructor

SimonKrughoff: You are saying take size out of the dataclass entirely since it is an attribute of the masked image anyway.  I see your point.

timj: Yes, I am asking why you need to set it when you know it already. Using SIZE to check consistency during construction is fine but I don't understand why it needs to be part of the dataclass.

SimonKrughoff: OK.  I'm leaving it out, then.

timj: You can of course add a `getSize()` method if the size is needed, and that can query the masked image.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: You're right.  It's a PropertyList.  I'll change the docs.

SimonKrughoff: But the rest of the code uses `PropertySet`.  Is there a functional difference?

timj: PropertyList is a subclass of PropertySet but PropertyList is optimized for FITS headers (so for example comments are allowed). PropertyList deals with hierarchy differently to PropertySet.  I imagine no-one really cares about the difference if you are mostly using it as a simple dict.

SimonKrughoff: OK.  I'll change to `PropertyList` everywhere

python/lsst/meas/algorithms/stamps.py
MorganSchmitz: Yep, tight control of this metadata was what I had in mind. Initially I just had a non-property getter return a `_metadata` attribute to simply _imply_ users are unlikely to want to change things in the metadata. Then following [discussions with Nate during review](https://github.com/lsst/meas_algorithms/pull/205#discussion_r462333347), I went with the copy.

My reasoning was that we are using the metadata to read and write a lot of ad hoc information - I believe for the generic class that'll just be number of stamps, and toggles for whether the variance and mask planes are present. But in my original case of bright stars, I also have Gaia object IDs, annular flux measurements, and parameters related to the latter that are absolutely necessary to use the data. So I figured I really wanted to avoid users accidentally changing that.

I don't know if that all makes sense? It is by no means crucial to have it be a copy; again, I had it return the actual thing initially. Feel free to change to that if that sounds like a better option!

ktlim: I haven't been following this closely, but if you are using FITS headers to persist structural information or actual data which is retrieved from the in-memory Python class via attributes or methods, I think it is best practice to strip that information from any flexible key/value metadata loaded from the FITS headers and re-create it on output.  There should only be one copy of that data in any given object.  See how headers are handled in Exposure.

timj: The current PR removes the use of this metadata where a property is more appropriate (so for example the number of "stamps" is not stored there any more).

MorganSchmitz: I'll make sure to change the bright star subclass to do the same, or at least strip the copy in the metadata like KT was saying, as soon as this is merged!

SimonKrughoff: OK, I'm going to return the actual object.  Thanks for all the input.

PR #222, opened 2020-12-16
DM-28040: Fix bugs in Stamps persistence

SimonKrughoff:


tests/test_stamps.py
```diff
@@ -81,6 +85,11 @@ def testIO(self):
         """
         self.roundtrip(make_stamps())
 
+    def testIOone(self):
+        """Test the class' write and readFits methods.
```
MorganSchmitz, line 89: Might be worth adding "when there is only a single stamp in the collection" to the docstring here

MorganSchmitz (approved):
Looks good to me! Nice that we could get rid of converting metadata to dict before passing it on the factory as a side effect.

Thanks a lot for taking the time to make the fix for brightStarStamp as well!

PR #229, opened 2021-02-24
DM-28858: Compare the data products of the gen2 vs. gen3 ci_hsc's up to Single Frame Processing

laurenam:


python/lsst/meas/algorithms/measureApCorr.py
```diff
@@ -177,7 +177,7 @@ def __init__(self, schema, **kwds):
         Task.__init__(self, **kwds)
         self.refFluxKeys = FluxKeys(self.config.refFluxName, schema)
         self.toCorrect = {}  # dict of flux field name prefix: FluxKeys instance
-        for name in getApCorrNameSet():
+        for name in sorted(list(getApCorrNameSet())):
```
TallJimbo, line 180: `sorted(apCorrNameSet())` is all you need; `sorted` can take any iterable, and always returns a new `list`.

TallJimbo (approved):
Looks good.  If I'm understanding the logic correctly, I think the comments are a bit pessimistic: given that you've started with an explicitly not-sorted list, how could the test pass as a fluke?

python/lsst/meas/algorithms/measureApCorr.py
TallJimbo: Sounds good; PR approved.

laurenam:
Ah, well, I was following the useful comment you made when doing the same in `meas_base` (https://github.com/lsst/meas_base/commit/2120646ad5f5dd8e4a7f2fe05180baee736fe1b0).  I did also test this fragility.  If I use the updated test (and just that one for simplicity...i.e. I disabled all but `def testAddFields(self)` for the following) repeatedly with no sorting I get (warning: ugly bash loops below!):
```
$ for i in {1..20}; do python tests/test_measureApCorr.py; done &> test.log
$ grep FAILED test.log | wc
     12      24     444
$ grep OK test.log | wc
      8       8     160
```
So...without sorting, the test for sorted keys actually passes 8/20 times!  On the other hand, with the sorting added, I get the following:
```
$ for i in {1..20}; do python tests/test_measureApCorr.py; done &> testSorted.log
$ grep FAILED testSorted.log | wc
      0       0       0
$ grep OK testSorted.log | wc
     20      20     400
```
I did a quick search to see if this is expected and this post (https://stackoverflow.com/questions/26420945/python-order-of-elements-in-set) tells me that:
> You cannot have order in sets. and there is no way to tell how Python orders it.

so I guess it is the expected behavior and my wording is actually not all that off?

TallJimbo:
Ah, yup, you were right all along.  I hadn't realized that after we set up the initial container, it goes through a `set`.  Mind adding that fact to one of the comments, as my past self clearly failed to do?

PR #230, opened 2021-03-01
DM-28717: Allow for Stamps formatter to handle bbox

MorganSchmitz:
One of four PRs for [DM-28717](https://jira.lsstcorp.org/browse/DM-28717), the others being afw, obs_base and daf_butler.


SimonKrughoff (changes requested):
It would be nice to have a test that tests this functionality.  I probably should have implemented one when I originally generalized this, but now is better than never.

I don't think it needs to do more than make sure the resulting image has the correct dimensions

MorganSchmitz:
Good call. Done!

PR #231, opened 2021-03-09
DM-28936: Try to get calibration source selection consistent between gen2 and gen3 middleware

laurenam:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -296,26 +297,23 @@ def _makeBoxRegion(BBox, wcs, BBoxPadding):
...
+        # region in at least one dimension, in which case all reference catalogs
+        # must be checked fully against the input bounding box.
         if innerLocalBBox.getDimensions() == geom.Extent2D(0, 0):
-            innerSkyRegion = sphgeom.Box()
-        else:
-            innerBoxCorners = innerLocalBBox.getCorners()
-            innerSphCorners = [wcs.pixelToSky(corner).getVector() for corner in innerBoxCorners]
-
-            innerSkyRegion = sphgeom.ConvexPolygon(innerSphCorners)
+            innerLocalBBox = geom.Box2D(BBox)
```
TallJimbo, line 305: Two things here confuse me:
 - This comment says "in at least one dimension" both before and after these changes, but the code clearly says "in both dimensions".
 - Did you remove the else clause because we're _actually_ hitting the case where the bbox is zero-sized in both dimensions _and_ converting an empty `Box2D` to a `sphgeom.Box` does not produce an empty `sphgeom.Box`?  Both of those clauses being true (where the first seems unlikely, and the latter seems like a separate bug) seem to be the only way the removal of the `else` clause could affect what happens.

Or have I just misread the logic here?

laurenam: Yes, the wording and implementation of this function really confused me too.  In playing around with the `BBoxPadding` that gets passed in, I went to the extreme (as changing it wasn’t seeming to have an effect on the trimming of the catalog), I tried a value of 2500.  This should, in principle, only shrink one dimension of an HSC chip down to zero...but I did hit this clause (maybe we don’t allow an single-dimension zero without setting the other to zero as well...that seems to be sensible, actually) and then hit an error when the code tried to return `innerSphCorners` as it did not get defined in that clause.  Regardless of dimensions, it seemed setting the inner box to a NaN-box was not a good thing to do, so I set it back to the original bbox (I _think_ this is in line with the wording!) and that way the other settings could be moved out of the clause, thus avoiding the error.

As a side note, I’m not clear on the usefulness of this shrunken bbox, but I may be missing “the point”.

TallJimbo: > we don’t allow an single-dimension zero without setting the other to zero as well

Exactly right, and I should have thought of this earlier.  But the comment is certainly confusing if you don't know that.  Mind fixing it here?

> Regardless of dimensions, it seemed setting the inner box to a NaN-box was not a good thing to do, so I set it back to the original bbox (I think this is in line with the wording!) and that way the other settings could be moved out of the clause, thus avoiding the error.

Ah, ok, what I had missed is that you are setting this to the original bbox, and not to the empty one.  Still a weird way to get out of this edge case, but let's not mess with it now.  Mind commenting to that effect, too?  Something like
```
# This probably isn't the right way to handle an empty inner bbox, but it's what calling code currently expects.
```

laurenam: Ok, I've reworded, but this last comment has me thinking more about my side note above.  I don't really understand the value and use of this "inner" -- sure to be contained within the detectors actual sky -- bounding box. 
 It is used to skip the reference cat trimming loop if the reference catalog lies entirely within this inner region, which does makes sense...but if this case is true, it means the reference catalog loaded does not cover anywhere near what is needed/being asked for (by a huge factor in area since it is shrinking by the amount we really want it to be grown by!  I've actually also added a `log.warn` if that triggers since the user should at least be aware of this situation...) , but if the main goal is to have a box that is almost certain to be contained within the true WCS region of the dataset, we could set it to "shrink by max of BBoxPadding or N% of the number of pixels in the shorter dimension".

TallJimbo: I don't think we want that warning in `_filterFunction` - I think that'll cause it to warn if any one of the reference catalog shards is within the region requested, rather than warning only if the region covered by _all_ of the shards is within the region requested.  And that's why the logic makes sense.

That said, I don't think we actually have any reference catalogs whose shard size is small enough for this to happen, but it's hypothetically possible.  That raises the question of whether this is unnecessary complexity, because it's optimizing a case that doesn't happen in practice, but I think that's probably out of scope for this ticket.

laurenam: Ah, ok..now I think I understand the logic (I wasn’t thinking about this per-shard).  I totally agree on all of the above and will get rid of the warn!

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -290,146 +292,157 @@ def _makeBoxRegion(BBox, wcs, BBoxPadding):
...
+            Raised if no reference catalogs could be found for the specified
+            region.
+        TypeError
+            Raised if the loaded reference catalogs do not have matching
+            schemas.
         """
+        bboxPadding = self.config.pixelMargin
         innerSkyRegion, outerSkyRegion, _, _ = self._makeBoxRegion(bbox, wcs, bboxPadding)
+        paddedBbox = copy.deepcopy(bbox)
+        paddedBbox.grow(bboxPadding)
```
natelust, line 373: I can believe this works in this particular case, but it is not solving the original problem that the boxPadding was intended for.

If you take an arbitrary bounding box and attempt to use the corners of the bounding box to define a convex polynomial on the sky, you can end up in a situation where there is area of the bounding box that does not actually fall inside the outer convex polynomial (since you are trying to put a rectangular box onto a spherical grid). You could use the same value for both, but it is an additional growth factor (see below for more)

The bboxPadding argument was designed to handle that issue and not growing the bbox to worry about astrometric errors. I designed this such that if you wanted a larger box, you would grow the box before calling this method.

However, I take it that the calling code does not do this per-say if we want it to be the same for gen2 and gen3 at this point.

The correct fix for this then would be to leave the bboxPadding argument (because it is independent of the pixelMargin parameter, and is not part of the gen2 implementation) and grow the bounding box here because it seems we must for what is going on with gen2.

Something like:
```
def loadPixelBox(self, bbox, wcs, filterName=None, epoch=None, photoCalib=None, bboxPadding=100):
    ...
    paddedBbox = Box2I(bbox)
    paddedBbox.grow(self.config.pixelMargin)
    innerSkyRegion, outerSkyRegion, _, _ = self._makeBoxRegion(paddedBbox, wcs, bboxPadding)
```

You could rename the `bboxPadding` parameter to reflect that it affects the generation of the spherical box creation instead, if you want to disambiguate.

What was written

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -290,146 +292,157 @@ def _makeBoxRegion(BBox, wcs, BBoxPadding):
...
-        `lsst.pex.exception.RuntimeError`
-            Raised if no reference catalogs could be found for the specified region
-
-        `lsst.pex.exception.TypeError`
-            Raised if the loaded reference catalogs do not have matching schemas
-
+        RuntimeError
+            Raised if no reference catalogs could be found for the specified
+            region.
+        TypeError
```
natelust, line 436: This is actually a `pex.exceptions.TypeError` If you dont want it to be, we need to change the code.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -290,146 +292,157 @@ def _makeBoxRegion(BBox, wcs, BBoxPadding):
...
+            box (padded by self.config.pixelMargin).
 
         Raises
         ------
-        `lsst.pex.exception.RuntimeError`
-            Raised if no reference catalogs could be found for the specified region
-
-        `lsst.pex.exception.TypeError`
-            Raised if the loaded reference catalogs do not have matching schemas
+        RuntimeError
```
natelust, line 363: The type of these Error was correct before the change

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -290,146 +292,157 @@ def _makeBoxRegion(BBox, wcs, BBoxPadding):
...
         outerBoxCorners = outerLocalBBox.getCorners()
         outerSphCorners = [wcs.pixelToSky(corner).getVector() for corner in outerBoxCorners]
-
         outerSkyRegion = sphgeom.ConvexPolygon(outerSphCorners)
 
         return innerSkyRegion, outerSkyRegion, innerSphCorners, outerSphCorners
 
-    def loadPixelBox(self, bbox, wcs, filterName=None, epoch=None, photoCalib=None, bboxPadding=100):
-        """Load reference objects that are within a pixel-based rectangular region
+    def loadPixelBox(self, bbox, wcs, filterName=None, epoch=None, photoCalib=None):
```
natelust, line 325: This needs to change back, see below (feel free to ask more questions if what I wrote was not clear)

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -558,8 +563,7 @@ def joinMatchListWithCatalog(self, matchCat, sourceCat):
         """
         return joinMatchListWithCatalogImpl(self, matchCat, sourceCat)
 
-    @classmethod
-    def getMetadataBox(cls, bbox, wcs, filterName=None, photoCalib=None, epoch=None, bboxPadding=100):
+    def getMetadataBox(self, bbox, wcs, filterName=None, photoCalib=None, epoch=None):
```
natelust, line 566: This needs to be changed back, see other comment

natelust (changes requested):

python/lsst/meas/algorithms/loadReferenceObjects.py
timj: We should not be raising pex_exception RuntimeError or TypeError in python code. They should only be raised in C++ code. It would be better to change this code to use the normal python exceptions. (I still don't understand why pybind11 doesn't translate those exceptions to the standard Python ones at the interface -- you can catch a pexExcept.RuntimeError as a normal RuntimeError so there's generally no reason to use the pex_exceptions code in python at all).

PS pex_exceptions is imported twice at the top of this file with different aliases.

timj: Dev guide: https://developer.lsst.io/python/style.html?highlight=pex_exceptions#python-exceptions-should-be-raised-and-checked-in-python-code

natelust: I have no preference here. I think I just copy and pasted them from some code when I hacked this together quickly ages ago. I am happy for them to change, I just wanted to point out that the type info was not actually correct.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -901,15 +905,15 @@ def loadPixelBox(self, bbox, wcs, filterName=None, photoCalib=None, epoch=None):
         wcs : `lsst.afw.geom.SkyWcs`
             WCS; used to convert pixel positions to sky coordinates
             and vice-versa.
-        filterName : `str`
+        filterName : `str` or `None`, optional
```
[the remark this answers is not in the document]
TallJimbo: It's a little more subtle than that: `optional` here means "doesn't need to be passed", while `typing.Optional[T]` means "`T` or `None`.  But in any case, the standard practice would be to just write this as
```
filterName: `str`, optional
```

python/lsst/meas/algorithms/loadReferenceObjects.py
laurenam: I see your point, but I still don't think this will work because the `region` getting passed to `loadRegion` will be the `outerSkyRegion` that only got the `bboxPadding` added to it (i.e. `config.pixelMargin` will again be effectively ignored).  It's also not possible for someone running a pipeline to have any control over the `bboxPadding` value (not to mention in the two places it gets a default of 100), which seems very dangerous as a fixed "number of pixels" default is unlikely to be appropriate for all cameras/detectors.

I would think that it's reasonable for the user to be responsible for configuring the `config.pixelMargin` to accommodate this case.  Also, I think with the way the refcats are partitioned and the fact each partition that has any amount of overlap with the `region` under consideration will get loaded, it should be really unlikely(unlucky!) that the effective loaded region (with proper `pixelMargin` padding) will leave us with any missing area.  If you really don't trust the `config.pixelMargin` setting, I think this could only work if you are taking `bboxPadding` as "in addition to", so that the `outerSkyRegion` gets padded with `config.pixelMargin + bboxPadding`...but that just becomes convoluted and confusing (but, even then, I'd recommend the default be set to 0, not 100).

So, I'm still advocating getting rid of this option.  Let me know what you think...if you disagree, then we still need a solution to the fact that, *at minimum*, the padding controlling the size of the load regions needs obey `config.pixelMargin`.   Perhaps we could add to the `pixelMargin` doc string to emphasize this point?

[I will work on explicitly testing my statement above with plots such as those I attached to the ticket, but that involves strategically placed breaks to persist the loaded catalog, and I have to remind myself where they go!  So please do consider the logic once again while I'm at it...and stop me if you think it's unnecessary overkill!]

python/lsst/meas/algorithms/loadReferenceObjects.py
laurenam: See comments below.

python/lsst/meas/algorithms/loadReferenceObjects.py
laurenam: See my comment above.

python/lsst/meas/algorithms/loadReferenceObjects.py
natelust: I do disagree, the `outerSkyRegion` must be effectively `config.pixelMargin + bboxPadding` which is what happened in the example I put in the comment (notice the argument change to `_makeBoxRegion`. You are correct that the `paddedBbox.grow()` is needed in this method because we cant yet change the logic in the calling code.

I dont think I explained my self adequately above. The problem is that if you have a bbox grown by `pixelMargin` from which you then generate a bounding convex polynomial without growing a second amount, you can get a situation where there is on sky area for which `bbox.contains(coord_converted_to_x_y)` reports `True` but `outerRegion.contains(coord)` reports `False` due to how the polygons get inscribed vs the tangent plane projection, which is what the bbox is.

This would lead to a situation where some ref_cats may get rejected because their stored region dose not intersect the outerRegion, but some records w/n that ref_cat would actually fall within the bounding box.

You are right that currently the code does give an option to change the call signature to pass an alternative padding, but I consider that a bug we need to resolve post gen2. This loader will need refactored anyway, and we really should be moving it futher away from being a subtask rather than trying to conform to the old way of doing it. I put that option in there, as it was large enough in all the testing I did, and I didn't want to yet add a new config option to parent task config classes, as then that would have to be supported well forward.

python/lsst/meas/algorithms/loadReferenceObjects.py
natelust: Like I said in the other comment I am happy for them to be changed as you did. The reason that this was a `pexExceptions.TypeError` was because we were raising it explicitly here: https://github.com/lsst/meas_algorithms/blob/master/python/lsst/meas/algorithms/loadReferenceObjects.py#L460 before the change (that is unless you changed it before the force push too, and I somehow was mistakenly looking at master, which is something I would do) and I was thinking it should be that type in `loadPixelBox` as well because that calls this method and it could have been raised here. That said its all moot because you changed the raise.

python/lsst/meas/algorithms/loadReferenceObjects.py
laurenam: Ah...sorry, I did miss that you passed in the `paddedBbox` as well as `bboxPadding` to `_makeBoxRegion`.  Ok -- while my concerns about the unchangeable default still stand -- I will make your suggested change  and do some further testing (as this should load more area than before, I think the filtered catalog should still end up looking the same...but an explicit check is warranted!)  I'll post on the ticket once I have some results.

What do you think of `bboxPaddingForSphere` as a name for this variable (that's terrible...please do suggest something better!!  But I am keen to be more specific in the name).

natelust: The end catalog should be the same, in all but the weirdest cases (which we may not even ever hit with these instruments), but it would be one of those things that would be really hard for someone to trace down later if it did come up. 

maybe `bboxToSpherePadding` ? about as long as what you wrote. I'm not good at coming up with something either, which is how we got the confusing name to start with :smile:

laurenam: Ok, for now I've gone with the following (suggestions for improvement welcome):
```
        bboxToSpherePadding : `int`, optional                                                                  
            Padding to account for translating a set of corners into an                                        
            spherical (convex) boundary that is certain to encompase the                                       
            enitre area covered by the box.                             
```

laurenam: Wow...tips, dude!  We must have been typing simultaneously!

python/lsst/meas/algorithms/loadReferenceObjects.py
laurenam: Effectively reverted.

python/lsst/meas/algorithms/loadReferenceObjects.py
laurenam: Effectively reverted.

natelust (approved):

PR #232, opened 2021-03-25
DM-29412: Update StampsBase so that it can serve as StorageClass for subclasses

jbkalmbach:


python/lsst/meas/algorithms/stamps.py
```diff
@@ -33,9 +33,11 @@
 import lsst.afw.fits as afwFits
 from lsst.geom import Box2I, Point2I, Extent2I, Angle, degrees, SpherePoint
 from lsst.daf.base import PropertyList
+from lsst.daf.butler.core.utils import getFullTypeName
+from lsst.utils import doImport
 
 
-def writeFits(filename, stamp_ims, metadata, write_mask, write_variance):
+def writeFits(filename, stamp_ims, metadata, type_name, write_mask, write_variance):
```
timj, line 40: @SimonKrughoff should we start writing a version number to these files?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -291,7 +295,15 @@ def readFitsWithOptions(cls, filename, options):
         options : `PropertyList`
             Collection of metadata parameters
         """
-        raise NotImplementedError
+        # Load metadata to get class
+        metadata = afwFits.readMetadata(filename, hdu=0)
+        type_name = metadata["STAMPCLS"]
```
timj, line 300: Consider doing:

```python
type_name = metadata.get("STAMPCLS")
if type_name is None:
    raise RuntimeError(f"No class name in file {filename}. Unable to instantiate correct stamps subclass."
                       " Is this an old version format Stamps file?")
```

Without that check you'll get a confusing `KeyError("STAMPCLS")` which won't be helpful to the general user.

One other thing that worries me a little is that since this is no longer an abstract method, it's possible for a subclass to decide not to implement it themselves. If that was to happen things would blow up really spectacularly since I think you'd get a recursion error.

Maybe short circuit that by doing something like:

```python
if cls is not StampsBase:
    raise NotImplementedError(f"Please implement specific FITS reader for class {cls}")
```
as an equivalent to the abstract method that was there before.

timj (commented):
Looks okay and I do have a couple of suggestions that should make it more robust against confusing error messages. I'd like a second opinion from @SimonKrughoff .

python/lsst/meas/algorithms/stamps.py
jbkalmbach: Thanks. Added in the error message and some tests to look for them.

python/lsst/meas/algorithms/stamps.py
SimonKrughoff: I guess so since we are breaking the API, but does that mean we have to also check the version and fall back to the old API when `STAMPCLS` is missing?  Seems like a lot of overhead, but if Morgan needs to read old stamps with new code, it will cause a problem.  I don't have any old STAMPS around, so am happy to just move on with the new order

timj: For now adding the VERSION won't have much effect because the only thing that really matters is STAMPCLS being present or not. If STAMPCLS is missing we can only complain and can't do anything to change that with versioning.  I'm more thinking about a future world where we decide to make another change and it would be handy if we could have a VERSION=1 vs VERSION=2 header (absence meaning VERSION == 0).  We've shown in our serialization of Exposure that versioning does become useful eventually.

SimonKrughoff: Fair enough, then I'm happy to take your suggestions

jbkalmbach: Ok, I updated to add a version number in the metadata. For now is that all I should do?

timj: Yes, seems fine.

PR #233, opened 2021-03-30
DM-29501: replace unary_function, binary_function, deprecated in C++17

mwittgen:


src/CR.cc
```diff
@@ -73,7 +73,7 @@ class IdSpan {
 /**
  * comparison functor; sort by ID, then by row (y), then by column range start (x0)
  */
-struct IdSpanCompar : public std::binary_function<const IdSpan::ConstPtr, const IdSpan::ConstPtr, bool> {
+  struct IdSpanCompar {
```
erykoff, line 76: Is it really as simple as just removing the `binary_function` bit here?  Huh!

erykoff (approved):

src/CR.cc
mwittgen: Looks like copy/paste from existing code. Unless I am missing something.

PR #234, opened 2021-03-31
DM-28803: Add capability to discard stars with NaN annuli

leeskelvin:


python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -124,6 +125,9 @@ def measureAndNormalize(self, annulus, statsControl=afwMath.StatisticsControl(),
         # compute annularFlux
         annulusStat = afwMath.makeStatistics(annulusImage, statsFlag, statsControl)
         self.annularFlux = annulusStat.getValue()
+        if np.isnan(self.annularFlux):
+            raise ValueError("Annular flux computation failed, likely because "
```
taranu, line 129: ValueError is "Raised when an operation or function receives an argument that has the right type but an inappropriate value,", so perhaps this should be a RuntimeError or ArithmeticError (or just make a very specific subclass called NanAnnularFluxError or similar).

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -254,9 +261,15 @@ def initAndNormalize(cls, starStamps, innerRadius, outerRadius,
...
-                                      badMaskPlanes=badMaskPlanes)
+        for j, stamp in enumerate(bss._stamps):
+            try:
+                stamp.measureAndNormalize(annulus, statsControl=statsControl, statsFlag=statsFlag,
+                                          badMaskPlanes=badMaskPlanes)
+            except ValueError:
+                if discardNanFluxObjects:
+                    bss._stamps.pop(j)
+                else:
+                    stamp.annularFlux = np.nan
```
taranu, line 272: It's not obvious to me what use this would serve, so it may be worth explaining here or in the function docstring under Notes.

python/lsst/meas/algorithms/brightStarStamps.py
leeskelvin: `RuntimeError` seems good to me, will update, thanks.

python/lsst/meas/algorithms/brightStarStamps.py
leeskelvin: Which aspect are you referring to in particular: the `if/else`, the `pop`, setting the `stamp.annularFlux` to `np.nan`, or something else? Happy to add a brief comment into the text to clarify.

taranu: The setting to nan part. A brief comment on why you might want to keep the stamp anyway would be helpful (my only immediate guess would be for bookkeeping so that you know that it tried and failed to measure something).

leeskelvin: Thanks Dan. Added the following comment block into the file:
```
# Optionally keep NaN flux objects, for bookkeeping purposes,
# and to avoid having to re-find and redo the preprocessing
# steps needed before bright stars can be subtracted.
```

PR #242, opened 2021-06-07
DM-30445: Add optional Archive element to stamps

MorganSchmitz:
[Jira ticket](https://jira.lsstcorp.org/browse/DM-30445)

[Jenkins run 34358](https://ci.lsst.codes/blue/organizations/jenkins/stack-os-matrix/detail/stack-os-matrix/34358/pipeline)

### DIFF
--- a/python/lsst/meas/algorithms/stamps.py
+++ b/python/lsst/meas/algorithms/stamps.py
@@ -292,7 +292,7 @@ class StampsBase(abc.ABC, Sequence):
     A butler can be used to read only a part of the stamps,
     specified by a bbox:
 
-    >>> starSubregions = butler.get("brightStarStamps_sub", dataId, bbox=bbox)
+    >>> starSubregions = butler.get("brightStarStamps", dataId, parameters={'bbox': bbox})
     """
 
     def __init__(self, stamps, metadata=None, use_mask=True, use_variance=True,
--- a/python/lsst/meas/algorithms/stamps.py
+++ b/python/lsst/meas/algorithms/stamps.py
@@ -32,13 +32,14 @@ from typing import Optional
 
 import lsst.afw.image as afwImage
 import lsst.afw.fits as afwFits
+import lsst.afw.table as afwTable
 from lsst.geom import Box2I, Point2I, Extent2I, Angle, degrees, SpherePoint
 from lsst.daf.base import PropertyList
 from lsst.daf.butler.core.utils import getFullTypeName
 from lsst.utils import doImport
 
 
-def writeFits(filename, stamp_ims, metadata, type_name, write_mask, write_variance):
+def writeFits(filename, stamp_ims, metadata, type_name, write_mask, write_variance, write_archive=False):
     """Write a single FITS file containing all stamps.
 
     Parameters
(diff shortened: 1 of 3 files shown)


python/lsst/meas/algorithms/stamps.py
```diff
@@ -113,35 +125,44 @@ def readFitsWithOptions(filename, stamp_factory, options):
...
+            md = afwFits.readMetadata(filename, hdu=idx+1)
+            if md['EXTNAME'] in ('IMAGE', 'VARIANCE'):
+                reader = afwImage.ImageFitsReader(filename, hdu=idx+1)
+            elif md['EXTNAME'] == 'MASK':
+                reader = afwImage.MaskFitsReader(filename, hdu=idx+1)
+            elif md['EXTNAME'] == 'ARCHIVE_INDEX':
+                f.setHdu(idx+1)
+                archive = afwTable.io.InputArchive.readFits(f)
+                continue
+            elif md['EXTTYPE'] == 'ARCHIVE_DATA':
```
SimonKrughoff, line 161: I'm not clear what this is doing.  I'd naively assume that the extension with EXTNAME = ARCHIVE_INDEX would have EXTTYPE = ARCHIVE_DATA, or is it that the archive system knows what archive data to read?

tests/test_stamps.py
```diff
@@ -46,15 +49,32 @@ def make_stamps(n_stamps=3):
...
-                  for stampIm, ra, dec in zip(stampImages, ras, decs)]
+                                                         geom.Angle(dec, geom.degrees)),
+                               archive_element=ae)
+                  for stampIm, ra, dec, ae in zip(stampImages, ras, decs, archive_elements)]
     metadata = PropertyList()
     metadata['RA_DEG'] = ras
     metadata['DEC_DEG'] = decs
 
-    return stamps.Stamps(stamp_list, metadata=metadata)
+    return stamps.Stamps(stamp_list, metadata=metadata, use_archive=True)
```
SimonKrughoff, line 62: I think we also need a test that uses the default of `use_archive=False`.  I don't think we necessarily need a duplicate test for True and False for all tests, but maybe test with the default in most tests and have a specific test that uses True?

SimonKrughoff (approved):
I do think we need to test with `use_archive=True` and `use_archive=False`, but other than that, it looks good.

python/lsst/meas/algorithms/stamps.py
MorganSchmitz: They're two (at least) separate HDUs. `InputArchive` knows how to handle both if you just give it the one that has EXTNAME `ARCHIVE_INDEX`. This line is just to avoid raising ValueError when we encounter the other HDU.s that was written by `OutputArchive`, which will always have EXTTYPE `ARCHIVE_DATA`.

Following the discussions in [DM-28716](https://jira.lsstcorp.org/browse/DM-28716), the _EXTNAME_ for those HDUs actually contain the type of Persistable that have been stored, so I can't check against that (since we don't know a priori what that will be), hence my using EXTTYPE here.

Following the discussions in [DM-28716](https://jira.lsstcorp.org/browse/DM-28716), the EXTNAME of

SimonKrughoff: OK

PR #245, opened 2021-06-23
DM-30829: Add persistence for WarpedPsf

kfindeisen:
This PR implements persistence for `WarpedPsf`. The adopted persisted form is merely the concatenation of the three subobjects currently making up the `WarpedPsf` implementation, as I don't see a more generic way to represent that state. Depends on lsst/afw#597.

### DIFF
--- a/tests/test_warpedPsf.cc
+++ b/tests/test_warpedPsf.cc
@@ -1,3 +1,26 @@
+/*
+ * This file is part of meas_algorithms.
+ *
+ * Developed for the LSST Data Management System.
+ * This product includes software developed by the LSST Project
+ * (https://www.lsst.org).
+ * See the COPYRIGHT file at the top-level directory of this distribution
+ * for details of code ownership.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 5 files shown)


tests/test_warpedPsf.cc
```diff
@@ -247,24 +309,23 @@ BOOST_AUTO_TEST_CASE(warpedPsf)
     unwarped_psf->evalABC(a, b, c, q);
 
     Eigen::Matrix2d m0;
-    m0 << a, b,
-          b, c;
+    m0 << a, b, b, c;
```
jmeyers314, line 312: I assume this is to make clang-format happy?  Is there a way to do that and keep it obvious that these are entries in a 2x2 matrix like before?

tests/test_warpedPsf.cc
```diff
@@ -247,24 +309,23 @@ BOOST_AUTO_TEST_CASE(warpedPsf)
...
 
     AffineTransform atr = lsst::afw::geom::linearizeTransform(*distortion->inverted(), p);
 
     Eigen::Matrix2d md;
-    md << atr.getLinear()[0], atr.getLinear()[2],
-          atr.getLinear()[1], atr.getLinear()[3];   // LinearTransform transposed index convention
+    md << atr.getLinear()[0], atr.getLinear()[2], atr.getLinear()[1],
```
jmeyers314, line 317: This seems like a regression in the formatting to me too.  Before it was clear that a 2x2 matrix was being processed.  Now that's less clear.

(Not trying to get into a format war here, but if there's a way to satisfy automated tools / standards and maintain useful information like "this is a 2x2 matrix", that'd be great.  If not, then carry on I guess.)

jmeyers314 (approved):
LGTM.  Just a couple formatting questions.

tests/test_warpedPsf.cc
kfindeisen: Sorry, I just blindly ran `clang-format`. Since I'm not familiar enough with Eigen to know if there's another way to write this, I'll just revert.

kfindeisen: Aha, there's a workaround: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#disabling-formatting-on-a-piece-of-code

PR #251, opened 2021-07-19
DM-30202: Add "online" accumulator-based mean stack consistent with afw.math.

erykoff:


tests/test_accumulatorMeanStack.py
```diff
@@ -0,0 +1,202 @@
...
+import lsst.afw.math as afwMath
+from lsst.meas.algorithms import AccumulatorMeanStack, stats_ctrl_to_threshold_dict
+
+
+class AccumulatorMeanStackTestCase(lsst.utils.tests.TestCase):
+    def make_test_images_to_coadd(self):
+        """Make test images to coadd."""
+        np.random.seed(12345)
+
+        n_image = 10
```
arunkannawadi, line 17: Perhaps indicate in a comment that `n_image` has to be at least 5 and the `image_size` has to be at least 150 pixels along each dimension.

tests/test_accumulatorMeanStack.py
```diff
@@ -0,0 +1,202 @@
...
+        masked_image_list = [exp.maskedImage for exp in exposures]
+
+        afw_masked_image = afwMath.statisticsStack(masked_image_list,
+                                                   stats_flags,
+                                                   stats_ctrl,
+                                                   weights,
+                                                   clipped,
+                                                   mask_map)
+
+        # Make the stack with the online accumulator
```
arunkannawadi, line 163: This comment should be two lines down.

arunkannawadi (approved):
I still have to wrap my head around the calculations going in `threshold_dict`, but nothing that should stop the merge. I left other minor suggestions inline.

python/lsst/meas/algorithms/accumulatorMeanStack.py
```diff
@@ -0,0 +1,184 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the LSST License Statement and
+# the GNU General Public License along with this program.  If not,
+# see <https://www.lsstcorp.org/LegalNotices/>.
+#
+import numpy as np
+
+
+__all__ = ['AccumulatorMeanStack', 'stats_ctrl_to_threshold_dict']
```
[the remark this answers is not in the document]
erykoff: So now that I reread the dev guide I knew https://developer.lsst.io/python/style.html#naming-conventions-for-science-pipelines "class names are CamelCase with leading uppercase", so the name of the class as `AccumulatorMeanStack` is correct.  However, I now also see that https://developer.lsst.io/python/style.html#a-python-source-file-name-should-be-camelcase-with-leading-lowercase-or-snake-case-and-end-in-py so the _file_ should be renamed to `accumulator_mean_stack.py`.

python/lsst/meas/algorithms/accumulatorMeanStack.py
```diff
@@ -0,0 +1,184 @@
...
+
+    return threshold_dict
+
+
+class AccumulatorMeanStack(object):
+    """Stack masked images.
+
+    Parameters
+    ----------
+    shape : `tuple`
```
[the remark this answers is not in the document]
erykoff: The reason I did it this way is that therefore the code isn't dependent on specific stack classes.  However, I'm still using `lsst.afw.image.maskedImage` objects as inputs, so it is dependent after all.  At the same time, the way this task is being used by descwl really prefers the shape as a tuple, so I'm going to leave it this way if that's okay.

tests/test_accumulatorMeanStack.py
erykoff: There's no such restriction, this is just arbitrary.

arunkannawadi: But you do explicitly use first 5 images to introduce artifacts by hand and if I later edit the test code to `n_image = 3`, the test would fail.

erykoff: Ah, I think I see the confusion.  The idea is to set either 1 bad pixel in the stack or half in the stack to test different scenarios with the thresholding.  And the number of images and the setting of the images in the tests are not independent.  I've added the following comment:
```
        # Choose an arbitrary number of images and image size.
        # Below we set bad pixels for either 1 or half these images,
        # at arbitrary locations.
```

PR #256, opened 2021-09-13
DM-30927: Set calexp WCS to None for failed astrometric fit

laurenam:


python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -101,7 +104,14 @@ def calculateThreshold(self, exposure, seed, sigma=None):
                 Additive factor to be applied to the background
                 level (`float`).
         """
-        # Make a catalog of sky objects
+        if exposureOrig.getWcs() is None:  # create a dummy WCS as needed by ForcedMeasurementTask
+            self.log.info("WCS for exposure is None.  Setting a dummy WCS for dynamic detection.")
+            exposure = copy.deepcopy(exposureOrig)  # don't wand to modify original exposure
```
erykoff, line 109: Is there a `clone` for exposure objects?  Or does that not do what we need here?

Also `wand` -> `want`.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -490,18 +490,55 @@ def apply(self, catalog):
...
+
+
 class ScienceSourceSelectorConfig(pexConfig.Config):
     """Configuration for selecting science sources"""
     doFluxLimit = pexConfig.Field(dtype=bool, default=False, doc="Apply flux limit?")
     doFlags = pexConfig.Field(dtype=bool, default=False, doc="Apply flag limitation?")
     doUnresolved = pexConfig.Field(dtype=bool, default=False, doc="Apply unresolved limitation?")
     doSignalToNoise = pexConfig.Field(dtype=bool, default=False, doc="Apply signal-to-noise limit?")
     doIsolated = pexConfig.Field(dtype=bool, default=False, doc="Apply isolated limitation?")
+    doRequireFiniteRaDec = pexConfig.Field(dtype=bool, default=False, doc="Apply sky coordinates limitation?")
```
erykoff, line 534: This doc should be `Apply finite sky coordinate check?` or I guess limitation is okay (since it's how the others are written, sigh) but it needs the `finite`.

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -490,18 +490,55 @@ def apply(self, catalog):
...
     fluxLimit = pexConfig.ConfigField(dtype=FluxLimit, doc="Flux limit to apply")
     flags = pexConfig.ConfigField(dtype=RequireFlags, doc="Flags to require")
     unresolved = pexConfig.ConfigField(dtype=RequireUnresolved, doc="Star/galaxy separation to apply")
     signalToNoise = pexConfig.ConfigField(dtype=SignalToNoiseLimit, doc="Signal-to-noise limit to apply")
     isolated = pexConfig.ConfigField(dtype=RequireIsolated, doc="Isolated criteria to apply")
+    requireFiniteRaDec = pexConfig.ConfigField(dtype=RequireFiniteRaDec,
+                                               doc="Sky coordinates criteria to apply")
```
erykoff, line 541: `Finite sky coordinate criteria to apply`

erykoff (approved):
Minor wording issues, otherwise looks good.

python/lsst/meas/algorithms/dynamicDetection.py
TallJimbo: I would not assume any such issues have been resolved.

But why do we need a deep copy anyway?  I don't see a reason for this ticket to introduce changes to pixel values or modifications to Exposure components (as opposed to replacing Exposure components with new ones or None), and if that's the case a shallow copy is all we want.  I'd do that by calling the Exposure constructor directly because I understand what that does, but I may be overly pessimistic about more Pythonic options like `copy.copy` or `clone` and it'd be even better to see if they do what we want (though if `clone` exists, I think it does deep copy at least *pixels* and again I'm skeptical that's desired here).

laurenam: I honestly can't recall why I felt a copy was necessary at all here (maybe just overly paranoid, as usual!), and the covid-fog is still pretty thick...

If you all think just using the original exposure for the dummy WCS setting then setting back to `None` just before the `return` is a-ok, I'll go with that!

As noted in the commit message, the context for this even triggering does not happen anywhere, so this is really just a failsafe in case anyone wants to use this task in another context.  I played around with testing this via:
```
from lsst.meas.algorithms import DynamicDetectionTask, DynamicDetectionConfig                                  
import lsst.geom as geom                                                                                       
                                                                                                               
root = "/sdf/data/rubin/u/laurenma/LSST/ci_hsc_gen3/DATA"                                                      
collection = "HSC/runs/ci_hsc"                                                                                 
camera = "HSC"                                                                                                 
butler = Butler(root, instrument=camera, collections=collection)                                               
                                                                                                               
config = DynamicDetectionConfig()                                                                              
config.doTempWideBackground = True                                                                             
config.background.undersampleStyle = 'REDUCE_INTERP_ORDER'                                                     
config.background.binSize = 4096                                                                               
config.background.useApprox = False                                                                            
config.reEstimateBackground = False                                                                            
config.isotropicGrow = True                                                                                    
config.thresholdType = "pixel_stdev"                                                                           
                                                                                                               
det = butler.get("deepCoadd_det", band="r", tract=0, patch=69)                                                 
schema = det.schema                                                                                            
table = afwTable.SourceTable.make(schema)                                                                      
exposure = butler.get("calexp", detector=0, visit=903344)                                                      
detection = DynamicDetectionTask(config=config, schema=schema)                                                 
detections = detection.run(table, exposure, expId=0)
```
Indeed, the results as far as I checked (catalogs of sky objects, `bgMedian`, `stdevMeas`, `medianError`, image plane after the fact), all looks identical with and without the copy, so I'm reasonably confident just operating directly on the exposure object is "safe".

PR #260, opened 2021-10-14
DM-30624: Document gen2 to gen3 refcat ingestion

leeskelvin:


doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -128,14 +128,54 @@ Monitor the log file in a new terminal with:
...
+
+    This is an example script that constructs conversion lookup tables for all converted FITS files in the `gaiadr2` and `panstarrsps1` directories, outputting to the `outdir` directory:
+
+    .. code-block:: python
+
+        import os
+        import glob
+        import astropy.table
+
+        refcat_dirs = [gaiadr2, panstarrsps1]
```
parejkoj, line 152: These need to be strings.

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -128,14 +128,54 @@ Monitor the log file in a new terminal with:
...
+        for refcat_dir in refcat_dirs:
+
+            outfile = f"{outdir}/{os.path.basename(refcat_dir)}.ecsv"
+            print(f"Saving to: {outfile}")
+
+            table = astropy.table.Table(names=("filename", "htm7"), dtype=("str", "int"))
+            files = glob.glob(f"{refcat_dir}/[0-9]*.fits")
+
+            for ii, file in enumerate(files):
+                print(f"{ii}/{len(files)} ({100*ii/len(files):0.1f}%)", end="\r")
```
parejkoj, line 166: I don't think you want a print here at all: it would print ~65k lines. I don't think it's necessary; how long does it take to run, anyway?

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -128,14 +128,54 @@ Monitor the log file in a new terminal with:
...
+        # 1) the FITS filename, and 2) the htm7 pixel index
+        for refcat_dir in refcat_dirs:
+
+            outfile = f"{outdir}/{os.path.basename(refcat_dir)}.ecsv"
+            print(f"Saving to: {outfile}")
+
+            table = astropy.table.Table(names=("filename", "htm7"), dtype=("str", "int"))
+            files = glob.glob(f"{refcat_dir}/[0-9]*.fits")
+
+            for ii, file in enumerate(files):
```
parejkoj, line 165: Single `i` for iterator, please?

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -128,14 +128,54 @@ Monitor the log file in a new terminal with:
...
+                print(f"{ii}/{len(files)} ({100*ii/len(files):0.1f}%)", end="\r")
+                # try/except to catch extra .fits files which may be in this dir
+                try:
```
parejkoj, line 168: This `try: except` worries me. The glob above should only catch index files; if it's getting other fits files (there should only be one: `schema.fits`), we should fix the glob.

If you really want an except here, at least print the filename that raised in the except clause.

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -128,14 +128,54 @@ Monitor the log file in a new terminal with:
...
+    .. code-block:: python
+
+        import os
+        import glob
+        import astropy.table
+
+        refcat_dirs = [gaiadr2, panstarrsps1]
+
+        # loop over each FITS file in all refcats
+        # note: this constructs a series of .ecsv files, each containing two columns:
```
parejkoj, line 155: It's not "a series of ecsv files", it's one per refcat.

Thinking on this: I think this script should have a `refcat="gaiadr2"` or whatever at the top, and only a single for loop below. This document is about converting/ingesting a single refcat, and we should keep it that way in this note.

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -128,14 +128,54 @@ Monitor the log file in a new terminal with:
...
 
 .. _lsst.meas.algorithms-refcat-ingest:
 
 4. Ingest the files into the butler
 ===================================
 
-When ``convertReferenceCatalog`` has finished, it will print the two commands you need to run to register the new refcat dataset type and ingest your converted output into it.
+When ``convertReferenceCatalog`` has finished, a new directory (named ``gaia-refcat/`` in the example above) will now exist containing all reference catalogs in the LSST format.
+
+.. note::
```
parejkoj, line 140: Looking at the generated text, the `note` breaks up the flow quite a lot, and is irrelevant for anyone following the docs with a fresh refcat. I wonder if this would be better as a `section 5: what to do with existing refcats`?

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
leeskelvin: They're meant to be objects, that the user has already assigned elsewhere. I hoped the intro text to this code block explained that. As this code is intended to be used on other machines, where the refcats may be saved in a non-standard location, I thought this was the safest bet.

If it would help, I could add some dummy lines above in the script along the lines of:

```
gaiadr2 = "/path/to/my/gaia/dir"
panstarrsps2 = "/path/to/my/panstarrs/dir"
```

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
leeskelvin: Good idea. I'll move it as you suggest.

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
leeskelvin: Following the clean-up as per your comments below, I've also removed this try/except block too. You're right in that it shouldn't be necessary, and the example reads more cleanly without it.

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -155,3 +158,40 @@ You can query the ``refcats`` collection to see whether your htm shards appear:
     butler query-datasets --collections refcats REPO
 
 For LSST staff using ``lsst-devl``, see the `Reference catalogs policy <https://developer.lsst.io/services/datasets.html#reference-catalogs>`_ in the Developer Guide for additional policy about adding reference catalogs to the common repo.
+
+.. _lsst.meas.algorithms-refcat-existing:
+
+5. What to do with existing reference catalogs?
+===============================================
+
+Already existing reference catalogs (for example, the PS1 or Gaia DR2 catalogs that were used in gen2 butlers) can also be ingested into the butler, negating the need to convert them to the LSST format yourself.
```
parejkoj, line 167: Maybe "... can be directly ingested into a gen3 repo as they are already in the LSST format."

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -155,3 +158,40 @@ You can query the ``refcats`` collection to see whether your htm shards appear:
...
+To ingest an already existing converted reference catalog, first create a suitable filename to htm7-index astropy lookup table, and then follow the steps above to `ingest the files into the butler <lsst.meas.algorithms-refcat-ingest>`.
```
parejkoj, line 168: The link at the end of this sentence doesn't work in the generated output. I think you need a `:ref:` in front.

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -155,3 +158,40 @@ You can query the ``refcats`` collection to see whether your htm shards appear:
...
+
+This is an example script that constructs a conversion lookup table for all converted FITS files in the `refcat_dir` directory, outputting to the `out_dir` directory:
```
parejkoj, line 170: "that constructs..." -> ```that creates an ``.ecsv`` file for the ``butler ingest-files`` command, from all of the HTM indexed files in a given directory. We use the existing Gaia DR2 catalog on lsst-devl in this example:```

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -155,3 +158,40 @@ You can query the ``refcats`` collection to see whether your htm shards appear:
...
+        # running status, overwriting each print statement as it proceeds
+        print(f"{i}/{len(files)} ({100*i/len(files):0.1f}%)", end="\r")
+
+        # extract file index; add row to table
+        file_index = int(os.path.basename(os.path.splitext(file)[0]))
+        table.add_row((file, file_index))
+
+    table.write(out_file)
+
+Once this script is complete, finalize the reference catalog ingestion by following the `file ingestion instructions above <lsst.meas.algorithms-refcat-ingest>`.
```
parejkoj, line 197: "finalize"->"finish", maybe?

And the final link doesn't work in the output HTML. I think you need a `:ref:` before these.

https://developer.lsst.io/restructuredtext/style.html#linking

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -155,3 +158,40 @@ You can query the ``refcats`` collection to see whether your htm shards appear:
...
+
+This is an example script that constructs a conversion lookup table for all converted FITS files in the `refcat_dir` directory, outputting to the `out_dir` directory:
+
+.. code-block:: python
+
+    import os
+    import glob
+    import astropy.table
+
+    refcat_dir = "/path/to/my/refcat/directory"
```
parejkoj, line 178: In my comment above, I suggested using `/datasets/refcats/htm/v1/gaia_dr2_20200414` here, to make it clear what exactly the this path should look like for an existing refcat. The rest of this example refers to lsst-devl (e.g. in the GaiaSource path, and the number of cores to use in the convert), so I don't think it hurts to do the same here.

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
```diff
@@ -155,3 +158,40 @@ You can query the ``refcats`` collection to see whether your htm shards appear:
...
+5. What to do with existing reference catalogs?
+===============================================
+
+Already existing reference catalogs (for example, the PS1 or Gaia DR2 catalogs that were used in gen2 butlers) can also be ingested into the butler, negating the need to convert them to the LSST format yourself.
+To ingest an already existing converted reference catalog, first create a suitable filename to htm7-index astropy lookup table, and then follow the steps above to `ingest the files into the butler <lsst.meas.algorithms-refcat-ingest>`.
+
+This is an example script that constructs a conversion lookup table for all converted FITS files in the `refcat_dir` directory, outputting to the `out_dir` directory:
+
+.. code-block:: python
+
```
parejkoj, line 173: Lets put a docstring at the top:
```
"""Generate an astropy-readable .ecsv files for `butler ingest-files`, to ingest an existing gen2 refcat.
"""
```

doc/lsst.meas.algorithms/creating-a-reference-catalog.rst
leeskelvin: `This is an example script that creates an ``.ecsv`` lookup table for the ``butler ingest-files`` command, from all of the HTM indexed files in a given directory (`refcat_dir` here). We use the existing Gaia DR2 catalog on lsst-devl in this example:`

leeskelvin:
Thanks John. I think I've addressed everything above. I hope that this now reads well; let me know.

parejkoj (approved):
Thanks, this looks good. The links work in the built version.

PR #265, opened 2021-10-28
DM-32355: Develop function and unit test to remove Poisson contribution from sources in variance plane

MorganSchmitz:


beckermr:
What is the status of this one?

MorganSchmitz:
The code itself is ready (it's just a very simple free function)
Iirc, I never got it to review because I ran into trouble trying to make unit tests for it, iirc it had to do with creating an Exposure with several Amplifiers (see [this slack thread](https://lsstc.slack.com/archives/C013QRP608J/p1635518144201700))

and then I guess it completely slipped my mind, sorry

python/lsst/meas/algorithms/variance_plane.py
```diff
@@ -0,0 +1,79 @@
...
+            amp_im_arr = exposure[amp_bbox].image.array
+            amp_var_arr = variance_plane[amp_bbox].array
+            good = (amp_var_arr != 0) & np.isfinite(amp_var_arr) & np.isfinite(amp_im_arr)
+            fit = np.polyfit(amp_im_arr[good], amp_var_arr[good], deg=1)
+            # fit is [1/gain, sky_var]
+            gain = 1./fit[0]
+            variance_plane[amp_bbox].array[good] -= amp_im_arr[good]/gain
+    else:
+        im_arr = exposure.image.array
+        variance_plane.array -= im_arr/gain
```
arunkannawadi, line 78: These calculations should still happen for each amplifiers separately if `average_accross_amps = False`, right?

tests/test_variance_plane.py
```diff
@@ -0,0 +1,917 @@
...
+from lsst.meas.algorithms import remove_signal_from_variance
+from lsst.utils.tests import methodParametersProduct
+from matplotlib.legend_handler import HandlerTuple
+from matplotlib.patheffects import withStroke
+from matplotlib.ticker import FixedLocator, FuncFormatter
+
+# Set to True to save the plot of the variance plane before and after
+# correction for a representative test case.
+SAVE_PLOT = True
+
```
enourbakhsh, line 41: I've left the plotting parts in for now, knowing we'll likely remove them before finalizing. I thought they might be useful for the review.

arunkannawadi: Just set it to `False` before merging. It's useful diagnostic to have when things go wrong but not worth saving a copy for every run.

TallJimbo (approved):
Looks good!

In addition to the line comments, I'd like to make sure this is tested to robustly do nothing (or nearly nothing) when given an image with no signal, rather than (say) subtracting arbitrary constants.

I'm also interested in the use case of doing this on coadds, if that's even possible - I'd love for someone to work through the algebra to see if:
- there is an "effective gain" on the coadd that is a weighted sum of images with different gains;
- if not, what the error is when running this algorithm on such a coadd if (say) those gains are drawn from a Gaussian distribution with some mean and standard deviation.

python/lsst/meas/algorithms/variance_plane.py
enourbakhsh: Yes, if I got your question right!

enourbakhsh:
> Looks good!
> 
> In addition to the line comments, I'd like to make sure this is tested to robustly do nothing (or nearly nothing) when given an image with no signal, rather than (say) subtracting arbitrary constants.
> 
> I'm also interested in the use case of doing this on coadds, if that's even possible - I'd love for someone to work through the algebra to see if:
> 
> * there is an "effective gain" on the coadd that is a weighted sum of images with different gains;
> * if not, what the error is when running this algorithm on such a coadd if (say) those gains are drawn from a Gaussian distribution with some mean and standard deviation.

I moved the unit test to `ip_isr` to avoid circular dependency. As requested, I added a test for when the image given has no signal to ensure nearly the same variance plane. Regarding its applicability to coadds, I've opened a new ticket linked to this one.

PR #266, opened 2021-11-22
DM-32624: Override source selection criteria in SFM calibration stages for LSSTCam-imSim

laurenam:


python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -64,10 +64,13 @@ class ObjectSizeStarSelectorConfig(BaseSourceSelectorTask.ConfigClass):
         dtype=bool,
         default=False,
     )
+    # Note that the signalToNoiseMin value should take into consideration the
+    # detection thresholds (set, e.g. in the characterizeImage setDefaults
+    # function for the measurePsf stage) for the catalog of interest.
```
yalsayyad, line 69: Another option to make your note more readily accessible would be to add a second sentence to the docstring:
"Should be greater or equal to the detection threshold SNR used, e.g., 50 for characterizeImage"

Unrelated random idea. Our config fields would benefit from a `defaultRationale` field. One day....

yalsayyad (approved):

python/lsst/meas/algorithms/objectSizeStarSelector.py
laurenam: I hesitated suggesting that the threshold for footprint detection directly maps to the source S/N threshold.  They seem to match up at about 50 here, but I'm not banking on that to be true, but rather had the empirical evidence from the plots.  I don't have good intuition for this, but is it possible the mapping could change (from one-to-one) as a function of the footprint detection threshold?  I can update the docstring, but am inclined to keep the vague "take into consideration" guideline to imply that a move to make changes should involve some "thought".

PR #267, opened 2021-11-23
DM-32683: Print refcat name in log when loading

parejkoj:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -958,7 +961,8 @@ def loadPixelBox(self, bbox, wcs, filterName=None, photoCalib=None, epoch=None):
         circle = self._calculateCircle(bbox, wcs)
 
         # find objects in circle
-        self.log.info("Loading reference objects using center %s and radius %s deg",
+        self.log.info("Loading reference objects from %s using center %s and radius %s deg",
+                      self.config.ref_dataset_name,
                       circle.coord, circle.radius.asDegrees())
```
leeskelvin, line 966: do these need to be on a new line here?

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -958,7 +961,8 @@ def loadPixelBox(self, bbox, wcs, filterName=None, photoCalib=None, epoch=None):
         circle = self._calculateCircle(bbox, wcs)
 
         # find objects in circle
-        self.log.info("Loading reference objects using center %s and radius %s deg",
+        self.log.info("Loading reference objects from %s using center %s and radius %s deg",
+                      self.config.ref_dataset_name,
```
leeskelvin, line 965: Does this need to have a new line at the end here? I think the args will still fit and be within 110 characters.

tests/ingestIndexTestBase.py
```diff
@@ -224,6 +227,10 @@ def setUpClass(cls):
     def setUp(self):
         self.repoPath = tempfile.TemporaryDirectory()  # cleaned up automatically when test ends
         self.butler = self.makeTemporaryRepo(self.repoPath.name, self.depth)
+        self.logger = lsst.log.Log.getLogger('lsst.ReferenceObjectLoader')
```
timj, line 230: Why is this an LSST logger? I've been switching things away from that. Assert logs will work without the forwarding because all task loggers are now python loggers.  What am I missing?

parejkoj: I did it that way it because I didn't know we could just use `logging.getLogger` for this kind of thing. I've switched it.

PR #271, opened 2021-12-07
DM-32662: Fix tests that assign to Angle columns.

TallJimbo:
Changes in afw make this an error by marking Angle column arrays as read-only via NumPy flags.  But those columns have never been views, so this code was *silently* broken before - it seems the tests didn't actually care what coordinate values were used.

### DIFF
--- a/tests/test_cr.py
+++ b/tests/test_cr.py
@@ -27,15 +27,15 @@ import unittest
 import lsst.geom
 import lsst.afw.image as afwImage
 import lsst.afw.math as afwMath
-import lsst.log.utils as logUtils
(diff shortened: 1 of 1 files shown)


natelust:
I think this all looks reasonable

natelust (approved):

TallJimbo:
Sorry for the confusion - I added these changes in response to some Jenkins failures, but @erykoff later informed me that this was actually a regression in `afw`, and the fix needed to go there.  So I've reset this branch to `main` (so the old one won't get picked up by Jenkins) and will delete it after merge.  And there are a pair of new commits (small ones) on the afw branch instead.

PR #272, opened 2021-12-16
Remove defaultFilter in LoadReferenceObjectsConfig

parejkoj:


kherner (approved):
The code itself looks good. I wonder if there are any other pieces of code that haven't been updated to reflect the new behavior (i.e. they don't call certain functions with a filter name because they expect the None default to be implicit). That may be beyond the scope of this ticket though.

parejkoj:
I searched through the stack for places that called the relevant code, and didn't find anything outside of this package. A full ci run didn't catch anything either. Nothing has set `defaultFilter` for years, so I think we're safe.

PR #274, opened 2021-12-28
DM-27843: Add anyFilterMapsToThis support to ReferenceObjectLoader

parejkoj:
As part of getting the gen3 code to support the more complicated filter mapping logic that the gen2 code did, I also refactored several methods into the common base class. Gen2 is going away soon, but we also want to keep as much of the gen2 test coverage as we can: this should help with that.

### DIFF
--- a/python/lsst/meas/algorithms/loadReferenceObjects.py
+++ b/python/lsst/meas/algorithms/loadReferenceObjects.py
@@ -301,33 +301,165 @@ class ReferenceObjectLoaderBase:
 
         applyProperMotionsImpl(self.log, catalog, epoch)
 
+    @staticmethod
+    def makeMinimalSchema(filterNameList, *, addCentroid=False,
+                          addIsPhotometric=False, addIsResolved=False,
+                          addIsVariable=False, coordErrDim=2,
+                          addProperMotion=False, properMotionErrDim=2,
+                          addParallax=False):
+        """Make a standard schema for reference object catalogs.
 
-class ReferenceObjectLoader(ReferenceObjectLoaderBase):
-    """This class facilitates loading reference catalogs with gen 3 middleware.
+        Parameters
+        ----------
+        filterNameList : `list` of `str`
+            List of filter names. Used to create <filterName>_flux fields.
+        addIsPhotometric : `bool`
+            If True then add field "photometric".
+        addIsResolved : `bool`
+            If True then add field "resolved".
+        addIsVariable : `bool`
+            If True then add field "variable".
+        coordErrDim : `int`
+            Number of coord error fields; must be one of 0, 2, 3:
 
-    The QuantumGraph generation will create a list of datasets that may
-    possibly overlap a given region. These datasets are then used to construct
-    and instance of this class. The class instance should then be passed into
-    a task which needs reference catalogs. These tasks should then determine
-    the exact region of the sky reference catalogs will be loaded for, and
-    call a corresponding method to load the reference objects.
+            - If 2 or 3: add fields "coord_raErr" and "coord_decErr".
+            - If 3: also add field "coord_radecErr".
+        addProperMotion : `bool`
+            If True add fields "epoch", "pm_ra", "pm_dec" and "pm_flag".
+        properMotionErrDim : `int`
+            Number of proper motion error fields; must be one of 0, 2, 3;
+            ignored if addProperMotion false:
+            - If 2 or 3: add fields "pm_raErr" and "pm_decErr".
+            - If 3: also add field "pm_radecErr".
+        addParallax : `bool`
+            If True add fields "epoch", "parallax", "parallaxErr"
+            and "parallax_flag".
 
-    Parameters
-    ----------
-    dataIds : iterable of `lsst.daf.butler.DataCoordinate`
-        An iterable object of data IDs that point to reference catalogs
-        in a gen 3 repository.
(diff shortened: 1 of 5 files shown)


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -461,6 +461,118 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+        with additional columns defined from the remaining function arguments.
+
+        Parameters
+        ----------
+        refCat : `lsst.afw.table.SimpleCatalog`
+            Reference catalog to map to new catalog
+        anyFilterMapsToThis : `str`
+            Always use this reference catalog filter.
+            Mutually exclusive with `filterMap`
+        filterMap : `dict` [`str`,`str`]
```
TallJimbo, line 477:
```suggestion
        filterMap : `dict` [`str`,`str`], optional
```

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -461,6 +461,118 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+            Mapping of camera filter name: reference catalog filter name.
+        centroids : `bool`
```
TallJimbo, line 479:
```suggestion
        centroids : `bool`, optional
```

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -461,6 +461,118 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+        or sets `anyFilterMapsToThis` in the schema.
+
+        Parameters
+        ----------
+        schema : `lsst.afw.table.Schema`
+            Schema for reference catalog.
+        anyFilterMapsToThis : `str`
+            Always use this reference catalog filter.
+            Mutually exclusive with `filterMap`
+        filterMap : `dict` [`str`,`str`]
```
TallJimbo, line 527:
```suggestion
        filterMap : `dict` [`str`,`str`], optional
```

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -461,6 +461,118 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+        For each camFilter: refFilter in filterMap, adds these aliases:
+            <camFilter>_camFlux:      <refFilter>_flux
+            <camFilter>_camFluxErr: <refFilter>_fluxErr, if the latter exists
+        or sets `anyFilterMapsToThis` in the schema.
+
+        Parameters
+        ----------
+        schema : `lsst.afw.table.Schema`
+            Schema for reference catalog.
+        anyFilterMapsToThis : `str`
```
TallJimbo, line 524:
```suggestion
        anyFilterMapsToThis : `str`, optional
```

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -461,6 +461,118 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+            Always use this reference catalog filter.
+            Mutually exclusive with `filterMap`
+        filterMap : `dict` [`str`,`str`]
+            Mapping of camera filter name: reference catalog filter name.
+            Mutually exclusive with `anyFilterMapsToThis`
+
+        Raises
+        ------
+        RuntimeError
+            If any required reference flux field is missing from the schema.
```
TallJimbo, line 534:
```suggestion
            Raised if any required reference flux field is missing from the schema.
```

TallJimbo (approved):
I reviewed commit-by-commit, and some code moved around, so it's possible the locations some changes should be made have moved.  But all comments are minor, and (I think) entirely about docs.

Overall, looks like a very nice cleanup.  I _think_ this doesn't actually touch the logic used to do the padding when actually loading refcat rows in regions, which is good - that code is _super_ fragile and not particularly well tested in CI, IIRC, and there are subtle differences between how Gen2 and Gen3 work that I think are important.  Someday we should try to fix that, of course, but for now I just want to make sure we avoid letting that problem torpedo unrelated "nearby" tickets like this one.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -295,35 +303,31 @@ def applyProperMotions(self, catalog, epoch):
...
-        config : `lsst.pex.config.configurableField`
-            Configuration for the loader.
-        log : `lsst.log.Log`, `logging.Logger` or `None`, optional
-            Logger object used to write out messages. If `None` a default
-            logger will be used.
-        """
+    Parameters
+    ----------
+    dataIds : iterable of `lsst.daf.butler.DataIds`
+        An iterable object of DataSetRefs which point to reference catalogs
```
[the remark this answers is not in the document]
parejkoj: I was hoping not to modify the logic of the gen3 loader code itself: this `__init__` gets called in potentially a lot of places (anything that uses a refcat and has a separate `runQuantum`, like jointcal and fgcmcal), so although that's probably a good cleanup, I think it needs its own ticket. It's also an API change, so might need an RFC, even if we all agree on it.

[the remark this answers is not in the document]
parejkoj: I filed this as https://jira.lsstcorp.org/browse/DM-33066

PR #278, opened 2022-04-19
DM-34484: Update tests to remove dependency on obs_test.

erykoff:


python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,116 @@ def makeDefectList():
...
+    convert : `bool`
+        Convert from old refcat format?
+    """
+    def __init__(self, filenames, config=None, convert=False):
+        self._cat = self._loadFiles(filenames, convert=convert)
+        self.config = config
+        if self.config is None:
+            self.config = LoadReferenceObjectsConfig()
+
+    def loadSkyCircle(self, ctrCoord, radius, filterName, epoch=None):
```
timj, line 210: Should we mention that the base class returns a SourceCatalog but this version returns a Struct? Am I reading the wrong documentation?

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,116 @@ def makeDefectList():
...
+                radius.asDegrees()
+            )
+        sel = np.zeros(len(self._cat), dtype=bool)
+        sel[idx[0]] = True
+        cat = self._cat[sel].copy(deep=True)
+
+        return Struct(refCat=cat, fluxField=f'{filterName}_flux')
+
+    def loadRegion(self, region, filterName, filtFunc=None, epoch=None):
+        return Struct(refCat=self._cat, fluxField=f'{filterName}_flux')
```
timj, line 224: This is also meant to be a SourceCatalog. I'm clearly missing something important.

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,116 @@ def makeDefectList():
...
+        cat['centroid_y'] = y
+        cat['hasCentroid'] = np.ones(len(cat), dtype=bool)
+        paddedBbox = lsst.geom.Box2D(bbox)
+        paddedBbox.grow(self.config.pixelMargin)
+        innerSkyRegion, _, _, _ = self._makeBoxRegion(paddedBbox, wcs, bboxToSpherePadding)
+        included = innerSkyRegion.contains(lon=cat['coord_ra'], lat=cat['coord_dec'])
+
+        cat = cat[included].copy(deep=True)
+
+        return Struct(refCat=cat, fluxField=f'{filterName}_flux')
```
timj, line 240: SimpleCatalog?

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,116 @@ def makeDefectList():
...
+
+    def _loadFiles(self, filenames, convert=False):
```
timj, line 242: Document the return value.

tests/test_refObjLoader.py
```diff
@@ -0,0 +1,289 @@
...
+from lsst.meas.algorithms import (ConvertReferenceCatalogTask, ReferenceObjectLoader)
+import lsst.utils
+import lsst.geom
+
+import ingestIndexTestBase
+
+
+class TestRefObjLoader(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
+                       lsst.utils.tests.TestCase):
+    def setUp(self):
```
timj, line 42: It's a shame that this gets run for every test when seemingly it's a readonly butler that is created and the tests aren't multi-threaded (at least I don't see any of the tests writing to a butler since the butler is not stored anywhere apart from the deferred dataset refs). Otherwise there is a lot of overhead creating a butler and populating it and then doing it all again. Can you see if setUpClass works?

timj (approved):
Looks great. I have one performance comment in the tests and confusion about the return values of the mock not matching (I think) the base class.

python/lsst/meas/algorithms/testUtils.py
erykoff: You are reading the documentation correctly, but the documentation was wrong. I have now fixed it on this ticket.

tests/test_refObjLoader.py
erykoff: Running all the tests here only takes 10s or so, even with the overhead.  I was having trouble with setupClass following the the same format of the other test this was based on.  There seemed to be some sort of filename clashes when running with xdist. The performance didn't seem like a problem and I didn't want to spend too much time trying to optimize.

erykoff: The problem might have been before I fixed all the temporary directory calls.  But once I got it working I didn't want to poke it any further...

timj: This change works for me:

```diff
diff --git a/tests/test_refObjLoader.py b/tests/test_refObjLoader.py
index 1941b559..4519d4e2 100644
--- a/tests/test_refObjLoader.py
+++ b/tests/test_refObjLoader.py
@@ -39,15 +39,18 @@ import ingestIndexTestBase
 
 class TestRefObjLoader(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
                        lsst.utils.tests.TestCase):
-    def setUp(self):
+
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
         np.random.seed(10)
 
         # Generate a second catalog, with different ids
         inTempDir = tempfile.TemporaryDirectory()
         inPath = inTempDir.name
-        skyCatalogFile, _, skyCatalog = self.makeSkyCatalog(inPath, idStart=25, seed=123)
+        skyCatalogFile, _, skyCatalog = cls.makeSkyCatalog(inPath, idStart=25, seed=123)
 
-        self.skyCatalog = skyCatalog
+        cls.skyCatalog = skyCatalog
 
         # override some field names, and use multiple cores
         config = ingestIndexTestBase.makeConvertConfig(withRaDecErr=True, withMagErr=True,
@@ -59,8 +62,8 @@ class TestRefObjLoader(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
         config.file_reader.format = 'ascii.commented_header'
         config.n_processes = 1
         config.id_name = 'id'  # Use the ids from the generated catalogs
-        self.repoTempDir = tempfile.TemporaryDirectory()
-        repoPath = self.repoTempDir.name
+        cls.repoTempDir = tempfile.TemporaryDirectory()
+        repoPath = cls.repoTempDir.name
 
         # Convert the input data files to our HTM indexed format.
         dataTempDir = tempfile.TemporaryDirectory()
@@ -69,7 +72,7 @@ class TestRefObjLoader(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
         converter.run([skyCatalogFile])
 
         # Make a temporary butler to ingest them into.
-        butler = self.makeTemporaryRepo(repoPath, config.dataset_config.indexer.active.depth)
+        butler = cls.makeTemporaryRepo(repoPath, config.dataset_config.indexer.active.depth)
         dimensions = [f"htm{depth}"]
         datasetType = DatasetType(config.dataset_config.ref_dataset_name,
                                   dimensions,
@@ -95,8 +98,8 @@ class TestRefObjLoader(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
         for dataRef in datasetRefs:
             handles.append(DeferredDatasetHandle(butler=butler, ref=dataRef, parameters=None))
 
-        self.datasetRefs = datasetRefs
-        self.handles = handles
+        cls.datasetRefs = datasetRefs
+        cls.handles = handles
 
         inTempDir.cleanup()
         dataTempDir.cleanup()
@@ -270,10 +273,12 @@ class TestRefObjLoader(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
         )
         self.assertEqual(result.fluxField, 'a_flux')
 
-    def tearDown(self):
+    @classmethod
+    def tearDownClass(cls):
+        super().tearDownClass()
         # Clear this to clean up the temporary directory.
-        self.repoTempDir.cleanup()
-        del self.repoTempDir
+        cls.repoTempDir.cleanup()
+        del cls.repoTempDir
 
 
 class TestMemory(lsst.utils.tests.MemoryTestCase):
```

and takes the non-multiprocessing testing from 16 seconds to 6 seconds and still works with xdist.

tests/test_htmIndex.py
```diff
@@ -1,372 +0,0 @@
...
-        config.requireProperMotion = False
-        config.anyFilterMapsToThis = "g"  # to use a catalog not made for obs_test
-        loader = LoadIndexedReferenceObjectsTask(butler=self.testButler, config=config)
-        with unittest.mock.patch.object(self.testButler, 'get', return_value=refcatData):
-            with self.assertLogs("lsst.LoadIndexedReferenceObjectsTask", level="WARNING") as cm:
-                loader.loadSkyCircle(center, self.searchRadius, "g", epoch=epoch)
-            warnLog1 = "Reference catalog pm_ra field is not an Angle; cannot apply proper motion."
-            self.assertEqual(cm.records[0].message, warnLog1)
-
-    def testLoadVersion0(self):
```
parejkoj, line 333: We've lost these tests in the new code. If the gen3 code doesn't support version0 refcats (I'd hope it does, but I'm not sure), we either need to fix it or RFC that lack of support.

Looking at lsst-devl, it looks like all the v0 refcats got converted to v1, so it's really a question of user support for the older ones. I'd support an RFC removing that support, I think.

tests/test_refObjLoader.py
```diff
@@ -0,0 +1,294 @@
...
+            filterName='a'
+        ).refCat
+
+        # Zero epoch change --> no proper motion correction (except minor numerical effects)
+        cat_pm = loader.loadSkyCircle(
+            center,
+            30.0*lsst.geom.degrees,
+            filterName='a'
+        ).refCat
+        loader.applyProperMotions(cat_pm, self.epoch)
```
parejkoj, line 211: This is unnecessary if you pass `epoch` to `loadSkyCircle`, and that's the way it is used in practice.

tests/test_refObjLoader.py
```diff
@@ -0,0 +1,294 @@
...
+            bbox,
+            wcs,
+            'a',
+            bboxToSpherePadding=0,
+            epoch=astropy.time.Time(30000, format='mjd', scale='tai')).refCat
+
+        self.assertFloatsNotEqual(cat['coord_ra'], catWithEpoch['coord_ra'], rtol=1.0e-4)
+        self.assertFloatsNotEqual(cat['coord_dec'], catWithEpoch['coord_dec'], rtol=1.0e-4)
+
+    def test_filterMap(self):
```
parejkoj, line 175: I think I'd rather this test didn't exist, and we just relied on the ones in `test_loadReferenceObjects.py`, but I don't know if this one provides any specific coverage that that doesn't.

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,121 @@ def makeDefectList():
...
+            self.config = LoadReferenceObjectsConfig()
+
+    def loadSkyCircle(self, ctrCoord, radius, filterName, epoch=None):
+        with Matcher(ctrCoord.getLongitude().asDegrees(), ctrCoord.getLatitude().asDegrees()) as matcher:
+            idx = matcher.query_radius(
+                np.rad2deg(self._cat['coord_ra']),
+                np.rad2deg(self._cat['coord_dec']),
+                radius.asDegrees()
+            )
+        sel = np.zeros(len(self._cat), dtype=bool)
```
parejkoj, line 217: `sel`?

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,121 @@ def makeDefectList():
                   ]
 
     return defectList
+
+
+class MockLoadReferenceObjects(ReferenceObjectLoader):
+    """A simple mock of LoadReferenceObjects for tests.
```
parejkoj, line 190: I'm not sure this is a "simple" mock: it seems like there's a lot going on in here. You've effectively re-implemented most of the important parts of `ReferenceObjectLoader` in here and I'm concerned that this hides some of the things we want to test. I wonder if we could get by with some `unittest.mock`s of dataId/datasetRef instead? Or even have a very minimal butler live in `meas_algorithms/tests/data`?

tests/test_refObjLoader.py
```diff
@@ -0,0 +1,294 @@
...
+import lsst.geom
+
+import ingestIndexTestBase
+
+
+class TestRefObjLoader(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
+                       lsst.utils.tests.TestCase):
+
+    @classmethod
+    def setUpClass(cls):
```
parejkoj, line 44: This setUpClass is very long. Is all of this really necessary, and/or could it be split out into free functions?

parejkoj (changes requested):
Thanks for fixing those docstrings!

I'm concerned about `MockLoadReferenceObjects` looking rather complicated. I think we should be able to pretty easily create a minimal butler or use unittest.mocks in e.g. `meas_astrom`.

I fear that just looking at coverage HTML isn't enough for this: the refcat code is not well isolated, so measured coverage might not change but we might actually lose out on things being covered (like the version0 tests). There are also multiple different files that test refcat-related things, in redundant and overlapping ways.

python/lsst/meas/algorithms/testUtils.py
erykoff: Renamed to `selected`.

tests/test_refObjLoader.py
erykoff: I'm copying over the tests that were there before in gen2-only land.  But the slight difference is that this is testing the return values of `loadSkyCircle` which the tests in `test_loadReferenceObjects.py` are not (because it has a dummy `loadSkyCircle`).  The second test (that the values are equal) is I guess redundant with the schema mapping tests, but it's not expensive.

erykoff: In particular, this test hits https://github.com/lsst/meas_algorithms/blob/f36ae6ace101d9874c6f103cd9182f19e188b404/python/lsst/meas/algorithms/loadReferenceObjects.py#L489

tests/test_refObjLoader.py
```diff
@@ -0,0 +1,294 @@
...
+        # Test that we require an epoch set.
+        msg = 'requireProperMotion=True but epoch not provided to loader'
+        with self.assertRaisesRegex(RuntimeError, msg):
+            _ = loader.loadSkyCircle(
+                center,
+                30.0*lsst.geom.degrees,
+                filterName='a'
+            ).refCat
+
+        # Test that anyFilterMapsToThis works.
```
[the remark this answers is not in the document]
erykoff: Ah yes, I have confirmed that it is.  I've removed these redundant tests (anyFilter and below), but left the config test in since that is otherwise not covered (hitting https://github.com/lsst/meas_algorithms/blob/f36ae6ace101d9874c6f103cd9182f19e188b404/python/lsst/meas/algorithms/loadReferenceObjects.py#L281 )

tests/ingestIndexTestBase.py
```diff
@@ -193,7 +189,8 @@ def tearDownClass(cls):
 
     @classmethod
     def setUpClass(cls):
-        cls.outPath = tempfile.mkdtemp()
+        cls.outDir = tempfile.TemporaryDirectory()
```
timj, line 192: We try to avoid using this in tests because it can lead to test failures over NFS caused by TemporaryDirectory not using the ignore_errors flag in shutil when it automatically runs the cleanup (because of nfs lock files hanging around). You might get away with it.

erykoff: Oh no!  Okay, we had a mix before and I thought that `mkdtemp` was old and `TemporaryDirectory` was new.  I can move everything over to `mkdtemp`.  The problem I was trying to solve was that the temporary directories aren't being cleaned up.  (These _might_ be fine because they all go in `/tmp` and who mounts `/tmp` over nfs?)

TallJimbo: Other people may be excited about fancy case syntax, but [`ignore_cleanup_errors`](https://docs.python.org/3/library/tempfile.html?highlight=tempfile#tempfile.TemporaryDirectory) is what I'm really looking forward to.

timj: Ooh. They've finally fixed it to allow you to say you don't care. If only someone would write a 3.10 migration RFC.

timj: Hmm. Don't we have `lsst.utils.tests.temporaryDirectory` that tries to fix this?

erykoff: Question: should I move this all back to mkdtemp and add the robust cleanup now, or let this slide until we get to 3.10?  I think it's fine because `/tmp` shouldn't be nfs mounted, but that doesn't mean it can't be nfs mounted.

TallJimbo: :+1: to `lsst.utils.tests.temporaryDirectory` as the solution for now.

erykoff: Hmmm.  That's a context manager so does it work against setup/teardown which we need here?

timj: 🤔 You are probably right that /tmp is not going to mess us up. I think the problem we had before was from people creating temp directories within the `tests/` directory (which is still common).

parejkoj: I don't see any notes about temporary directory usage in the dev guide, and I wasn't aware of that problem. Should we add a section about that to the python testing guide?

tests/test_convertReferenceCatalog.py
```diff
@@ -55,11 +56,14 @@ def test_main_args(self):
             # Test with sets because the glob can come out in any order.
             self.assertEqual(set(run.call_args.args[0]), set(self.expected_files))
 
+        outdir.cleanup()
```
parejkoj, line 59: Since you're using `TemporaryDirectory`, you don't need to call `cleanup`: it should be cleaned up automatically when it goes out of scope.

tests/test_loadReferenceObjects.py
```diff
@@ -235,6 +237,95 @@ def make_catalog():
         self.assertIsNone(newRefCat)
 
 
+class ConvertReferenceCatalogConfigValidateTestCase(lsst.utils.tests.TestCase):
+    """Test validation of IngestIndexReferenceConfig."""
```
parejkoj, line 241: IngestIndexReferenceConfig -> ConvertReferenceCatalogConfig

tests/test_referenceObjectLoader.py
```diff
@@ -0,0 +1,327 @@
...
+import ingestIndexTestBase
+
+
+class ReferenceObjectLoaderTestCase(ingestIndexTestBase.ConvertReferenceCatalogTestBase,
+                                    lsst.utils.tests.TestCase):
+    """Test case for ReferenceObjectLoader."""
+    @classmethod
+    def setUpClass(cls):
+        super().setUpClass()
+        np.random.seed(10)
```
parejkoj, line 50: No need to seed here: `makeSkyCatalog` does its own seeding.

parejkoj (commented):
A handful of new comments, but I think this is pretty close now. I filed an RFC about removing support for version0 refcats, but at least we know they should work in gen3 now.

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,122 @@ def makeDefectList():
...
+        def datasetType(self):
+            return self.MockDatasetType(self._name)
+
+    @property
+    def ref(self):
+        return self.MockRef(self._name)
+
+
+class MockReferenceObjectLoaderFromFiles(ReferenceObjectLoader):
+    """A simple mock of ReferenceObjectLoader to use a set of files.
```
parejkoj, line 250: Oh, let'ss expand this docstring to something like: "A ReferenceObjectLoader that doesn't know about regions on the sky or the butler, and takes the list of files on disk to mock dataIds."

erykoff:
```
    This mock ReferenceObjectLoader uses a set of files on disk to create
    mock dataIds and data reference handles that can be accessed
    without a butler. The files must be afw catalog files in the reference
    catalog format, sharded with HTM pixelization.
```

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -178,3 +184,122 @@ def makeDefectList():
...
+    filenames : `list` [`str`]
+        Names of files to use.
+    config : `lsst.meas.astrom.LoadReferenceObjectsConfig`, optional
+        Configuration object if necessary to override defaults.
+    htmLevel : `int`, optional
+        HTM level to use for the loader.
+    """
+    def __init__(self, filenames, name='cal_ref_cat', config=None, htmLevel=4):
+        if config is None:
+            config = LoadReferenceObjectsConfig()
```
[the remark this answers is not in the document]
erykoff: It's an easy fix, I can just add it on this ticket.

tests/test_convertReferenceCatalog.py
erykoff: You're right.  In some cases I see warnings, but I guess that's when it's not fully local?

PR #279, opened 2022-04-22
DM-31811: Add Detection SNR to DIASource catalogs

parejkoj:


python/lsst/meas/algorithms/detection.py
```diff
@@ -697,14 +701,75 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
...
+
             self.display(exposure, results, middle)
 
         return results
 
+    def setPeakSignificance(self, exposure, footprints, threshold, negative=False):
+        """Set the significance of each detected peak to the pixel value divided
+        by the appropriate standard-deviation for ``config.thresholdType``.
+
+        Only sets significance for "stdev" and "pixel_stdev" thresholdTypes.
```
TallJimbo, line 724: I feel like I should know the answer given our previous conversation, but why?

(And a reader other than me would probably have the same question.)

TallJimbo (approved):

python/lsst/meas/algorithms/detection.py
parejkoj: I'm not sure how to define a significance value for the `variance` thresholdType (and I don't think we use that anywhere); would it be the ratio of the variance, or of the stddev?

Similar question for `thresholdType=value`: if we just took `peak['peakValue'] / thresholdValue`, that would mean a detection has significance >=1, which could be confusing.

I figured I'd err on the side of doing the calculation for the obvious cases where there was a clear meaning, and leaving the others that I don't think we actually use anywhere.

parejkoj:
A question here: I originally wrote the doc value for the `significance` field as "Value of threshold image at peak position", but there is no such thing as a "threshold image" in `SourceDetectTask`; there may be an implicit one, but I feel that mentioning it like this is misleading. Since I'm only computing it for the two versions of stddev `thresholdType`, I'm thinking to change that to be `doc="Ratio of peak value to configured standard deviation"`. Thoughts?

python/lsst/meas/algorithms/detection.py
TallJimbo: That all sounds reasonable - just put something like that (even just a very brief summary) in the docs.

PR #283, opened 2022-06-06
DM-35105 Force (0, 0) for center of warped PSF images.

TallJimbo:
Previous logic forced the image size to be odd, but picked the center of the kernel image bbox based on center of the transformed kernel image bbox.  But the output kernel image center is an input: (0, 0).

This manifested as even-sizes images later in CoaddPsf; the union of two odd-sized boxes with the same dimensions but offset center is an even-sized box.

### DIFF
--- a/src/WarpedPsf.cc
+++ b/src/WarpedPsf.cc
@@ -89,11 +89,11 @@ geom::Box2I computeBBoxFromTransform(geom::Box2I const bbox, geom::AffineTransfo
     }
 
     // We want to guarantee that the output bbox has odd dimensions, so instead
-    // of using the Box2I converting constructor directly, we round the center
-    // point of the floating point box and dilate by its half-dimensions.
+    // of using the Box2I converting constructor directly, we start with (0, 0)
+    // and dilate by the floating-point box's half-dimensions.
     geom::Extent2I out_half_dims = geom::floor(0.5*out_box_fp.getDimensions());
     geom::Box2I out_box;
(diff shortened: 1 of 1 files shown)


src/WarpedPsf.cc
```diff
@@ -89,11 +89,11 @@ geom::Box2I computeBBoxFromTransform(geom::Box2I const bbox, geom::AffineTransfo
...
     // We want to guarantee that the output bbox has odd dimensions, so instead
-    // of using the Box2I converting constructor directly, we round the center
-    // point of the floating point box and dilate by its half-dimensions.
+    // of using the Box2I converting constructor directly, we start with (0, 0)
+    // and dilate by the floating-point box's half-dimensions.
     geom::Extent2I out_half_dims = geom::floor(0.5*out_box_fp.getDimensions());
     geom::Box2I out_box;
-    geom::Point2I out_center(out_box_fp.getCenter());
+    geom::Point2I out_center(0, 0);
     out_box.include(out_center);
```
fred3m, line 97: Any reason that you decided to define `out_box` this way instead of using `out_box = Box2I(Point2I(0,0), Extent2I(1,1))`?

fred3m (approved):
This is fine, I just had a quick question on style that is out of curiosity but doesn't require any changes.

src/WarpedPsf.cc
TallJimbo: Not really; it just fit my mental model of what I was doing better.

PR #284, opened 2022-06-06
DM-34698: Increase default kernelSize of PSF candidates to 25

arunkannawadi:


parejkoj (commented):
Please make the PR title include the ticket number.

PR #285, opened 2022-07-07
DM-34792: Deprecate version 0 refcat loading code

parejkoj:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -74,15 +78,23 @@ def getFormatVersionFromRefCat(refCat):
         Format verison integer. Returns `0` if the catalog has no metadata
         or the metadata does not include a "REFCAT_FORMAT_VERSION" key.
     """
+    # TODO DM-34793: change to "Version 0 refcats are no longer supported: refcat fluxes must have nJy units."
+    # TODO DM-34793: and raise an exception instead of returning 0.
+    deprecation_msg = "Support for version 0 refcats (pre-nJy fluxes) will be removed after v25."
```
erykoff, line 83: I don't like the phrase "pre-nJy fluxes".  Why not "Jy fluxes"?

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -74,15 +78,23 @@ def getFormatVersionFromRefCat(refCat):
...
         return 0
     try:
         return md.getScalar("REFCAT_FORMAT_VERSION")
     except KeyError:
+        warnings.warn(deprecation_msg)
         return 0
 
 
+# TODO DM-34793: remove this function
+@deprecated(reason="Support for version 0 refcats (pre-nJy fluxes) will be removed after v25.",
```
erykoff, line 96: Same as above.

erykoff (approved):

python/lsst/meas/algorithms/loadReferenceObjects.py
parejkoj: There was no actual flux definition in version 0: I had one or two examples with fluxes in Jy, but I think at least one very old catalog used something else. The conversion code all assumes Jy, but that wasn't a defined part of the catalogs.

PR #286, opened 2022-08-02
DM-35671: Remove gen2 reference loaders.

erykoff:
This also consolidates code from the no-longer-necessary `ReferenceObjectLoaderBase` base class.  It does not currently update naming conventions.


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -756,8 +755,8 @@ def joinMatchListWithCatalog(self, matchCat, sourceCat):
         return joinMatchListWithCatalogImpl(self, matchCat, sourceCat)
 
 
-class ReferenceObjectLoader(ReferenceObjectLoaderBase):
-    """This class facilitates loading reference catalogs with gen 3 middleware.
+class ReferenceObjectLoader:
+    """This class facilitates loading reference catalogs.
```
parejkoj, line 759: Oh, can you cleanup the paragraph below, too? e.g. `and instance`

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -777,14 +776,515 @@ class ReferenceObjectLoader(ReferenceObjectLoaderBase):
         Logger object used to write out messages. If `None` a default
         logger will be used.
     """
+    ConfigClass = LoadReferenceObjectsConfig
+
     def __init__(self, dataIds, refCats, log=None, config=None, **kwargs):
```
parejkoj, line 781: Can we get rid of the `**kwargs`, since they're not used?

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
```diff
@@ -23,134 +23,27 @@
...
+@deprecated(reason=("This config is no longer used; it will be removed after v24. "
+                    "Please use LoadReferenceObjectsConfig instead."),
+            version="v24.0", category=FutureWarning)
 class LoadIndexedReferenceObjectsConfig(LoadReferenceObjectsConfig):
     ref_dataset_name = pexConfig.Field(
         dtype=str,
         default='cal_ref_cat',
-        doc='Name of the ingested reference dataset'
+        doc='Name of the ingested reference dataset',
+        deprecated='This field is no longer used. It will be removed after v24.',
```
parejkoj, line 39: How did I miss that deprecation?

tests/test_loadReferenceObjects.py
```diff
@@ -26,24 +26,24 @@
...
 import lsst.pex.config
 import lsst.utils.tests
 
 from ingestIndexTestBase import makeConvertConfig
 
 
-class TrivialLoader(ReferenceObjectLoaderBase):
-    """Minimal subclass of ReferenceObjectLoaderBase to allow instantiation
+class TrivialLoader(ReferenceObjectLoader):
+    """Minimal subclass of ReferenceObjectLoader to allow instantiation
```
parejkoj, line 38: You can just delete `TrivialLoader`, since ReferenceObjectLoader is no longer an ABC.

parejkoj (changes requested):
Don't remove the `LoadReferenceObjectsTask` rst file: it contains useful information about the catalog schema. We do still have [DM-11044](https://jira.lsstcorp.org/browse/DM-11044) about finding a better permanent home for these. It's no longer a `Task`, so I think it would go int the generic module topic type:

https://developer.lsst.io/stack/generic-guide-topic-type.html

The way loadReferenceObjects.py was refactored makes it look like you added a bunch of code, but I think it was just shuffled around in the file. Is this just git doing a wonky diff?

I'd like to move all of `loadReferenceObjects.py` into `referenceObjectLoader.py` (and similarly for the config class) to try to get the names consistent. Maybe that can't be done on this ticket, but it'd be nice to try to start moving that direction.

erykoff:
Oh wow, me moving the code around seems to have really confused the github change display.  Sorry about that, but I think that's beyond the scope of this ticket.
Re: moving code into the same file, I'll see what more cleanup I can do.  But in terms of rectifying names I think that's better to leave for a separate ticket (and a separate diff! 😂 ).

python/lsst/meas/algorithms/loadIndexedReferenceObjects.py
erykoff: I can't answer that one!

python/lsst/meas/algorithms/loadReferenceObjects.py
erykoff: I'm worried that could break some existing code (if somebody was erroneously passing a `butler=` for example).  The problem was that usage was incorrect but never warned.  So what I'm going to do is check if `**kwargs` is non-empty and throw up a deprecation warning.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -756,35 +755,539 @@ def joinMatchListWithCatalog(self, matchCat, sourceCat):
...
+        Handles to load refCats on demand.
     log : `lsst.log.Log`, `logging.Logger` or `None`, optional
         Logger object used to write out messages. If `None` a default
         logger will be used.
     """
+    ConfigClass = LoadReferenceObjectsConfig
+
     def __init__(self, dataIds, refCats, log=None, config=None, **kwargs):
+        if kwargs:
+            warnings.warn("Instantiating ReferenceObjectLoader with additional kwargs is deprecated "
```
parejkoj, line 782: Oh, please file a ticket about that removal.

tests/test_loadReferenceObjects.py
```diff
@@ -26,30 +26,22 @@
...
-    """
-
-    def loadSkyCircle(self, ctrCoord, radius, filterName):
-        pass
-
-
-class ReferenceObjectLoaderBaseTestCase(lsst.utils.tests.TestCase):
-    """Test case for ReferenceObjectLoaderBase abstract base class.
+class ReferenceObjectLoaderTestCase(lsst.utils.tests.TestCase):
+    """Test case for ReferenceObjectLoader.
```
parejkoj, line 38: This line isn't necessary: that's obvious from the name. The sentence below probably could go too, since there's no more ABC. Maybe rename or put in a sentence about how this tests the more generic parts of it, not the actual catalog loading.

parejkoj (approved):
Two more small comments

I think you can rename `tests/nopytest_ingestIndexRef...` and the related `ingestIndexTestBase` to convert. Or we could save that for DM-31698 (if so, please note as such on that ticket). And `test_referenceObjectLoader` and it's class could be renamed to match?

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -756,35 +755,539 @@ def joinMatchListWithCatalog(self, matchCat, sourceCat):
...
+                          "and will be removed after v25.0", FutureWarning)
```
timj, line 783: Please use `stacklevel=2` here so that the warning appears to come from the caller code -- it makes it easier to find which of their lines had the errant kwarg. Generally stacklevel is always going to be useful.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -233,6 +232,13 @@ class LoadReferenceObjectsConfig(pexConfig.Config):
         dtype=bool,
         default=False,
     )
+    cal_ref_cat = pexConfig.Field(
```
parejkoj, line 235: Why is this called `cal_ref_cat`? That's not the name of any config field. The config field name in `LoadIndexedReferenceObjectsConfig` is `ref_dataset_name`, with a default value of `"cal_ref_cat"`. We want no default value here, certainly.

I'm assuming the rationale of this field is to be able to read in previously written configs? Maybe the NCSA->USDF transition will help with that, by pruning out the oldest runs, but we've always had a danger with removing config fields because old runs stick around a long time. I don't know how many cases it might actually come up though, and I think your approach here isn't terrible? Especially as long as this field doesn't get written when it's defaulted (does it?).

erykoff: This was a really dumb typo.

parejkoj: I just checked, and defaulted+deprecated fields don't get written, so that's fine.

PR #288, opened 2022-08-05
DM-971: fix errors now that tasks always validate on init

parejkoj:


tests/test_measure.py
```diff
@@ -351,7 +351,7 @@ def testPsfFlux(self):
         schema.addField("centroid_x", type=float)
         schema.addField("centroid_y", type=float)
         schema.addField("centroid_flag", type='Flag')
-        sfm_config = measBase.SingleFrameMeasurementConfig()
+        sfm_config = measBase.SingleFrameMeasurementConfig(ignoreSlotPluginChecks=True)
```
erykoff, line 354: This will need to be wrapped to ignore/check warnings if the meas_base change is done.

erykoff (approved):
Small change related to meas_base.

PR #290, opened 2022-08-13
DM-35871: Add refcat name arg to ReferenceObjectLoader init

parejkoj:


erykoff:
On this one, I would prefer if the name were an optional keyword argument, since (a) this code will still work fine if the name is `None`; (b) and changing this to a required positional argument is a breaking API change (even if this object is very rarely instantiated outside of pipeline code, it may be.)

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -264,13 +264,18 @@ class ReferenceObjectLoader:
         An iterable object of data IDs that point to reference catalogs.
     refCats : iterable of `lsst.daf.butler.DeferredDatasetHandle`
         Handles to load refCats on demand.
+    name : `str`
```
erykoff, line 267: Should add `, optional`

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -264,13 +264,18 @@ class ReferenceObjectLoader:
...
+        The name of the refcat that this object will load. This name is used
+        for applying colorterms, for example.
+    config : `lsst.pex.config.Config`
```
erykoff, line 270: Doesn't this need to be an instance of `LoadReferenceObjectsConfig` (or at least a subclass)?

python/lsst/meas/algorithms/testUtils.py
```diff
@@ -266,7 +266,7 @@ class MockReferenceObjectLoaderFromFiles(ReferenceObjectLoader):
     def __init__(self, filenames, name='cal_ref_cat', config=None, htmLevel=4):
         dataIds, refCats = self._createDataIdsAndRefcats(filenames, htmLevel, name)
 
-        super().__init__(dataIds, refCats, config=config)
+        super().__init__(dataIds, refCats, name, config=config)
```
erykoff, line 269: This should be `name=name`, and perhaps all of them by keyword.

tests/test_loadReferenceObjects.py
```diff
@@ -115,7 +115,7 @@ def testFilterAliasMap(self):
         for filterMap in ({}, {"camr": "r"}):
             config = ReferenceObjectLoader.ConfigClass()
             config.filterMap = filterMap
-            loader = ReferenceObjectLoader(None, None, config=config)
+            loader = ReferenceObjectLoader(None, None, None, config=config)
```
erykoff, line 118: This should be `name=None` or not set (default).

tests/test_loadReferenceObjects.py
```diff
@@ -155,7 +155,7 @@ def testAnyFilterMapsToThisAlias(self):
         # test anyFilterMapsToThis
         config = ReferenceObjectLoader.ConfigClass()
         config.anyFilterMapsToThis = "gg"
-        loader = ReferenceObjectLoader(None, None, config=config)
+        loader = ReferenceObjectLoader(None, None, None, config=config)
```
erykoff, line 158: Same here.

erykoff (approved):
Some minor changes to docstrings and usage requested.  I've clicked Approve anyway since these are all small.

python/lsst/meas/algorithms/loadReferenceObjects.py
parejkoj: Good catch, yes!

python/lsst/meas/algorithms/loadReferenceObjects.py
parejkoj: I'd rather we treated it as defacto required, since knowing what a refobject loader actually has loaded is very useful, but I guess we probably can't make it "silently optional"?

PR #292, opened 2022-08-31
DM-34253: Enable PSF padding by default in computeApertureFlux

leeskelvin:


fred3m (approved):
This looks good. I wasn't aware of being able to update an image using `assign`, that's a neat trick! I think I'll update the multi-band version in meas_extensions_scarlet at some point.

PR #293, opened 2022-08-31
DM-28329: Standardize on jyear and remove "EPOCH" header keys

parejkoj:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -715,9 +715,11 @@ def getMetadataCircle(coord, radius, filterName, epoch=None):
         md.add('RA', coord.getRa().asDegrees(), 'field center in degrees')
         md.add('DEC', coord.getDec().asDegrees(), 'field center in degrees')
         md.add('RADIUS', radius.asDegrees(), 'field radius in degrees, minimum')
-        md.add('SMATCHV', 1, 'SourceMatchVector version number')
+        # Version 2 uses EQUINOX for the TAI Julian Epoch year.
+        md.add('SMATCHV', 2, 'SourceMatchVector version number')
         md.add('FILTER', filterName, 'filter name for photometric data')
-        md.add('EPOCH', "NONE" if epoch is None else epoch.mjd, 'Epoch (TAI MJD) for catalog')
+        md.add('EQUINOX', "NONE" if epoch is None else epoch.jyear,
```
timj, line 721: I think the standard now requires us to use `JEPOCH` here or `MJD-OBS`

timj (approved):

python/lsst/meas/algorithms/loadReferenceObjects.py
parejkoj: I've updated this to JEPOCH, cleaned up some docs, and added a trivial test of `getMetadataCircle`. Please have another look.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -715,14 +716,19 @@ def getMetadataCircle(coord, radius, filterName, epoch=None):
...
         md.add('RADIUS', radius.asDegrees(), 'field radius in degrees, minimum')
-        md.add('SMATCHV', 1, 'SourceMatchVector version number')
-        md.add('FILTER', filterName, 'filter name for photometric data')
-        md.add('EPOCH', "NONE" if epoch is None else epoch.mjd, 'Epoch (TAI MJD) for catalog')
+        # Version 1: Initial version
+        # Version 2: JEPOCH for TAI Julian Epoch year of PM/parallax correction
+        md.add('SMATCHV', 2, 'SourceMatchVector version number')
+        md.add('FILTER', filterName, 'camera filter name for photometric data')
+        md.add('TIMESYS', "TAI", "time scale of time keywords")
+        md.add('JEPOCH', "NONE" if epoch is None else epoch.tai.jyear,
```
timj, line 724: FITS is going to really want JEPOCH to always be a floating point number. Can you please use `None` here instead? That would result in JEPOCH being an undefined value in the header which sounds like what you want.

timj (approved):
One comment about the use of a string to mean undefined that I think needs fixing.

PR #296, opened 2022-10-04
DM-36071: Deprecate kernelSize* parameters in PSF determiner tasks

arunkannawadi:


python/lsst/meas/algorithms/psfDeterminer.py
```diff
@@ -34,22 +34,51 @@ class BasePsfDeterminerConfig(pexConfig.Config):
...
     kernelSizeMax = pexConfig.Field(
-        doc="Maximum radius of the kernel",
+        doc="Maximum radius of the kernel. Relevant only if kernelSize < 15.",
         dtype=int,
         default=45,
+        deprecated="This field is deprecated and will be removed after v25.",
     )
 
+    def validate(self):
+        # TODO: DM-36311: This entire method may be removed after v25.
```
TallJimbo, line 69: Might be better to be specific about which lines are to be removed, on the off chance something else gets added to this method before that ticket happens.

python/lsst/meas/algorithms/psfDeterminer.py
```diff
@@ -34,22 +34,51 @@ class BasePsfDeterminerConfig(pexConfig.Config):
 
     This is fairly sparse; more fields can be moved here once it is clear they are universal.
     """
+    stampSize = pexConfig.Field(
```
TallJimbo, line 37: I see you used the new type annotation syntax elsewhere on this PR, but not here.  Any reason why?  (I don't have a preference, except for consistency.)

TallJimbo (approved):
Looks good, though I suspect it might be better to squash some commits, just given the number of commits to the number of actual lines changed overall.

If you don't squash, you've got one commit message with a typo, in which the actual word "typo" should be "type".

python/lsst/meas/algorithms/psfDeterminer.py
arunkannawadi: I did that in one file, where there were only two config fields. I'm happy to change it in the files I'm touching.

arunkannawadi:
Squashed some commits (reduced 19 to 10) and added one more.

PR #299, opened 2022-11-10
DM-25132: Add position covariances to reference catalog converter

cmsaunders:
This adds support for position covariance parameters in the reference catalog converter. It also adds a function for Gaia specifically to set the position covariance.


tests/test_referenceObjectLoader.py
```diff
@@ -481,6 +534,97 @@ def test_requireProperMotion(self):
...
+                                           name="testrefcat")
+
+    def test_properMotion(self):
+        """Test proper motion correction."""
+        """loaderConfig = ReferenceObjectLoader.ConfigClass()
+        loaderConfig.filterMap = {'aprime': 'a'}
+        loader = ReferenceObjectLoader([dataRef.dataId for dataRef in self.datasetRefs],
+                                       self.handles,
+                                       name="testrefcat",
+                                       config=loaderConfig)"""
```
parejkoj, line 585: I see why you had to override `setUp` (to make a v1-style refcat), but why did you have to override this method? It should "just work" as-is, I'd think.

Also, you have a commented-out block here.

cmsaunders: Thanks for catching the commented-out block. I had actually decided I didn't want to inherit from `ReferenceObjectLoaderLoadTests`, because the other tests don't need to be rerun for version one, but then I didn't change the class definition.

kfindeisen (changes requested):
This is an ambitious PR, but it feels like coherence has suffered as a result -- the code, documentation, and unit tests don't quite line up. (And some of my comments are therefore mutually exclusive, depending on which way you decide to go.)

The two issues I'd like to see addressed before I sign off on this are:
1. The API seems to require a lot of familiarity with the code to be usable: `ConvertRefcatManager` seems to have hidden assumptions about a specific subset of the Gaia catalog built in, and there are a lot of interlocking preconditions that will surprise users (and, in the case of `config.coordinate_covariance`, are neither enforced nor checked at runtime).
2. While described as a deprecation, the new API will break existing calls, particularly those that expect `properMotionErrDim` to guarantee that error columns are present in the output catalog (not to mention _any_ clients of the old astrometric error columns). Is there an RFC that agrees to just break everything without a deprecation period (I couldn't find one), and if so why are you talking about deprecation at all?

Other than that, most of my comments are style nitpicks. Thanks for taking this on!

doc/lsst.meas.algorithms/reference-catalog-schema.rst
```diff
@@ -30,21 +31,22 @@ Catalog Schema
...
+    - ``coord_ra_dec_Cov``: covariance between ``coord_ra`` and ``coord_dec`` (rad^2)
     - ``epoch``: date of observation as TAI MJD (day)
     - ``pm_ra``: proper motion along the direction of right ascension (rad/year, an Angle) = dra/dt * cos(dec)
     - ``pm_dec``: proper motion along the direction of declination (rad/year, and Angle)
-    - ``pm_raErr``: uncertainty in `pm_ra` (rad/year)
-    - ``pm_decErr``: uncertainty in `pm_dec` (rad/year)
-    - ``pm_ra_dec_Cov``: covariance between pm_ra and pm_dec (rad2/year2)
+    - ``coord_pm_raErr``: uncertainty in `pm_ra` (rad/year)
+    - ``coord_pm_decErr``: uncertainty in `pm_dec` (rad/year)
+    - ``coord_pm_ra_pm_dec_Cov``: covariance between pm_ra and pm_dec (rad^2/year^2)
```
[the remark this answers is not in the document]
parejkoj: This is defined by the afw `CovarianceMatrixKey` that is used to store the errors and covariances. The change here aligns the names with that.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -369,30 +370,35 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+            Deprecated (see DM-?????): Use ``fullPositionCovariance`` when proper motions
+            are provided, as they will be correlated with the other position
+            paramters.
         addParallax : `bool`
             If True add fields "epoch", "parallax", "parallaxErr"
             and "parallax_flag".
+        fullPositionCovariance : `bool`
+            Include a full coordinate+PM+parallax covariance matrix. Requires
+            that `addProperMotion` and `addParallax` are set.
+            Ignores the `coordErrDim` and  `properMotionErrDim` parameters.
```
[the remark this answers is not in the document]
parejkoj: You're probably right that `addProperMotion` and `addParallax` should go away, too.

Our logic here is that either you have PM+parallax+all covariance (Gaia), or you don't care about any position errors beyond ra/dec (anything non-Gaia). Gaia is the only relevant astrometric catalog (until we make our own, and we won't need this code for that), and whether or not a given gaia star has the full 5-parameter covariance is irrelevant to whether the refcat should have it (since it needs to be able to hold all the covariances).

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -369,30 +370,35 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
-        -----
-        Reference catalogs support additional covariances, such as
-        covariance between RA and proper motion in declination,
-        that are not supported by this method, but can be added after
-        calling this method.
         """
+        if properMotionErrDim is not None:
+            warnings.warn("properMotionErrDim is deprecated and has no effect; "
+                          "use `fullPositionCovariance` to include all position errors in the schema. "
+                          "It will be removed after release 25.0.", category=FutureWarning)
```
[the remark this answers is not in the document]
parejkoj: Because there's no use for the current system: these hooks were in place to allow someone to put those error fields in the refcat, but we didn't actually have any code that used the `=3` case. Trying to maintain the above functionality (which uses different names for the fields) along with the new full covariance seemed like excessive work for no gain. The only relevant astrometric catalog is Gaia, and until this ticket we couldn't actually load the relevant error fields for Gaia; I'm positive that removing that old functionality breaks nobody's code.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -1213,10 +1211,11 @@ def applyProperMotionsImpl(log, catalog, epoch):
         record.set(coordKey,
                    record.get(coordKey).offset(bearing=bearingRad*geom.radians,
                                                amount=amountRad*geom.radians))
+    # TODO: this needs to incorporate the full covariance!
```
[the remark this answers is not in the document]
parejkoj: This should say `TODO DM-36979:`, since that's the ticket for that work. We wanted to put that off so that we could get this into review and make the catalog, then work on implementing the math and tests later.

tests/convertReferenceCatalogTestBase.py
```diff
@@ -41,13 +41,16 @@ def make_coord(ra, dec):
...
+def makeConvertConfig(withMagErr=False, withRaDecErr=False, withPm=False,
                       withParallax=False):
     """Make a config for ConvertReferenceCatalogTask
 
     This is primarily intended to simplify tests of config validation,
     so fields that are not validated are not set.
     However, it can calso be used to reduce boilerplate in other tests.
+
+    NOTE: if you set `withPm`, you probably need to manually set
+    `config.coordinate_covariance=True` to get a useable config.
```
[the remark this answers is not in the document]
parejkoj: This is just here for helping write tests of the configs, so edge cases are a major point. Given your suggestion about removing the pm/parallax kwargs to `makeMinimalSchema`, we probably can simplify all this quite a lot.

doc/lsst.meas.algorithms/reference-catalog-schema.rst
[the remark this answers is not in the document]
kfindeisen: That doesn't actually answer my question. Are the users of the catalogs required to know how you've coded it? Is there a reason that, for example, `pm_dec` is not `coord_pm_dec`?

python/lsst/meas/algorithms/loadReferenceObjects.py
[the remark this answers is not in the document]
kfindeisen: My point is that you're adding lots of API and maintenance complexity for the sole purpose of reducing functionality compared to if you just did things the simple way.

But even if you are starting from a Gaia catalog (and this is supposed to be the _generic_ converter, not the Gaia specialization), why should users be required to translate parallax in order to get covariances for the other parameters? I can see why LSST needs proper motions, but parallax seems irrelevant to data processing.

doc/lsst.meas.algorithms/reference-catalog-schema.rst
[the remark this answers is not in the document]
cmsaunders: I decided to not use `afw`s `CovarianceMatrixKey`, so there is no longer a change in the naming convention.

python/lsst/meas/algorithms/convertReferenceCatalog.py
```diff
@@ -271,6 +270,16 @@ def assertAllOrNone(*names):
...
 
+        # Need all the error field names set if we are including covariances.
+        if self.coordinate_covariance:
+            assertAllOrNone("coordinate_covariance",
+                            "ra_err_name", "dec_err_name", "coord_err_unit",
+                            "epoch_name", "epoch_format", "epoch_scale",
+                            "pm_ra_name", "pm_dec_name",
+                            "pm_ra_err_name", "pm_dec_err_name",
+                            "parallax_name", "parallax_err_name"
+                            )
```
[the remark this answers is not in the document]
cmsaunders: We already know `coordinate_covariance` is True, so there is no way to have none of them true. I agree that is not obvious, so I added a comment.

tests/test_referenceObjectLoader.py
```diff
@@ -105,40 +101,58 @@ def testMakeMinimalSchema(self):
...
-                    self.assertEqual("parallax" in refSchema, addParallax)
-                    self.assertEqual("parallaxErr" in refSchema, addParallax)
-                    self.assertEqual("parallax_flag" in refSchema, addParallax)
+                    # We test setting thse these all together, below.
+                    self.assertNotIn("epoch", refSchema)
+                    self.assertNotIn("pm_ra", refSchema)
+                    self.assertNotIn("pm_dec", refSchema)
+                    self.assertNotIn("pm_flag", refSchema)
+                    self.assertNotIn("parallax", refSchema)
+                    self.assertNotIn("parallax_flag", refSchema)
```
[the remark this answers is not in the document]
cmsaunders: These should not be in the schema because the default for `addParallax` and `addProperMotion` is False, but I can remove this check if it seems confusing.

cmsaunders:
I've implemented the changes proposed in my last comment, and I think the result is significantly clearer. Please disregard the status of the commits. I made atomized commits to facilitate easier squashing, pending approval of the changes.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -446,7 +450,7 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
                 prefix="coord",
                 names=["ra", "dec"],
                 units=["rad", "rad"],
-                diagonalOnly=(coordErrDim == 2),
+                diagonalOnly=((coordErrDim == 2) or (fullCoordinateCovariance)),
```
kfindeisen, line 453: Doesn't this produce `coord_ra_dec_Cov`, not `coord_ra_coord_dec_Cov` as advertised? In fact, it seems that if `coordErrDim > 0` and `fullCovarianceMatrix`, then you get *both* columns.

Unfortunately, I don't see an easy fix short of modifying `addFields` to support empty prefixes; the code will inject a leading underscore if you just pass `prefix=""`.

kfindeisen (changes requested):
I'm sorry, but I still can't approve these changes. While the flag simplification could potentially make the documented API less confusing for users, the fact that the code doesn't match the documentation makes it impossible for me to tell what the intended behavior was (all three interpretations seem a bit odd). I think better test coverage would help.

I haven't seen anything addressing my other objection, that this PR is making breaking changes to both the API and the catalog schema and calling it a deprecation. Are you authorized to break everything (in which case no deprecation is needed), or are you trying to create a transition period (in which case backwards-compatibility needs to be preserved)? The current approach gives you the worst of both worlds.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -504,6 +505,24 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
                 type="Flag",
                 doc="Set if parallax or parallax error is bad",
             )
+            if fullCoordinateCovariance:
+                if not (addProperMotion and addParallax):
+                    warnings.warn("Adding proper motion and parallax covariances without proper motion and "
+                                  "parallax parameters and errors.")
```
parejkoj, line 511: This seems like a bad idea; why would we ever want someone to be able to do this?

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -504,6 +505,24 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+                fields = ["coord_ra", "coord_dec", "pm_ra", "pm_dec", "parallax"]
+                units = ["rad", "rad", "rad/year", "rad/year", "rad"]
+                for i in range(5):
+                    i_field = fields[i]
+                    i_unit = units[i]
```
parejkoj, line 516: I don't understand how this nested for-loop is an improvement over using the CovarianceMatrixXKey?

parejkoj:
I'm happy to file an RFC if one of the objections is that this is an API-breaking change for the refcat converter; I expect that RFC to have no objections at all, because the code dealing with position errors had no users other than me.

cmsaunders:
Thank you for the second look at the PR. I'll try to address the major points from your review now, and will go through the in-line comments in the next day or so:
- I did miss addressing the deprecation of `properMotionErrDim` issue in the changes to the PR, so I apologize for that. As @parejkoj commented, it is highly unlikely that anyone else is using this and would object to removing it. We can do an RFC, and if it is an option to completely remove it without going through the deprecation process, that is obviously preferable. If not, I can put back the original behavior. 
-  Since the recent commits, the changes in the reference catalog schema have been reduced to just the names of the coord_ra/coord_dec and pm_ra/pm_dec covariance, neither of which were actually present in any reference catalogs used in our processing. I don't know if an RFC is appropriate for changing the names, but I don't think there is any way to keep the names the same and still have overall naming consistency.
- I do see some places where the documentation is inconsistent and can be improved, but to avoid endless going back and forth over language and the general approach to the ticket, maybe we can find a time to meet and sort out an acceptable solution.

kfindeisen:
> I'm happy to file an RFC if one of the objections is that this is an API-breaking change for the refcat converter; I expect that RFC to have no objections at all, because the code dealing with position errors had no users other than me.

> I did miss addressing the deprecation of `properMotionErrDim` issue in the changes to the PR, so I apologize for that. As @parejkoj commented, it is highly unlikely that anyone else is using this and would object to removing it. We can do an RFC, and if it is an option to completely remove it without going through the deprecation process, that is obviously preferable. If not, I can put back the original behavior.
> Since the recent commits, the changes in the reference catalog schema have been reduced to just the names of the coord_ra/coord_dec and pm_ra/pm_dec covariance, neither of which were actually present in any reference catalogs used in our processing. I don't know if an RFC is appropriate for changing the names, but I don't think there is any way to keep the names the same and still have overall naming consistency.

I think a new RFC would be appropriate but won't insist on it, since that would add at least a week's delay to getting this merged. The main reason I mentioned it was that I assumed there already was one (you need to RFC deprecations, after all) and I just couldn't find it.

For me, the core issue here is _choosing_ whether you are breaking things now or following the deprecation process, and committing to that choice. If you don't think anybody would be affected by just removing the old API, then I'm willing to take your word for it.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -370,16 +370,17 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
 
             - If 2 or 3: add fields "coord_raErr" and "coord_decErr".
             - If 3: also add field "coord_radecErr".
-
-            Ignored if ``fullCoordinateCovariance=True``.
+                Ignored if ``fullCoordinateCovariance=True``.
```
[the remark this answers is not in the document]
cmsaunders: No, this line has an extra indent because it only applies to "If 3...". 
I don't think your second comment is correct. The only column that is in common is the `coord_ra`/`coord_dec` covariance. If `fullCoordinateCovariance` is True, then the `diagonalOnly` option at line 454 will be True, so the covariance will not be added there. You are correct that the two paths to adding ra/dec covariance end up with different names, but I don't see a way to avoid this.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -498,11 +495,32 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
...
+                type="Angle",
+                doc="uncertainty in parallax",
+                units="rad",
+            )
             schema.addField(
                 field="parallax_flag",
                 type="Flag",
                 doc="Set if parallax or parallax error is bad",
             )
+            if fullCoordinateCovariance:
```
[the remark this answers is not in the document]
cmsaunders: Ok, there were a couple errors on my part here. First, the indentation of this block of code was wrong. It should not have been inside the `if addParallax` block. Second, the documentation for `fullCoordinateCovariance` wasn't fixed to reflect the refactoring. (It now reads "Include the off-diagonal elements of the full coordinate+PM+parallax covariance matrix. The diagonal elements are provided by setting `coordErrDim`, `addProperMotion`, and `addParallax`.")
None of the three fields require each other now. The second of your bullet points above is now the only one that is true.

python/lsst/meas/algorithms/loadReferenceObjects.py
cmsaunders: It's correct that this produces `coord_ra_dec_Cov`, not `coord_ra_coord_dec_Cov` (which is the way `fullCoordinateCovariance` names the field). I think the only solution would be to just have a separate `addFields` command to add the covariance. We can discuss during the meeting today.

python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -507,8 +507,11 @@ def makeMinimalSchema(filterNameList, *, addCentroid=False,
                 doc="Set if parallax or parallax error is bad",
             )
             if fullCoordinateCovariance:
+                if not (addProperMotion and addParallax):
```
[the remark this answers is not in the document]
cmsaunders: Fixed this, so this is not in the `if addParallax` branch anymore.

python/lsst/meas/algorithms/loadReferenceObjects.py
cmsaunders: I agree that no one will want to do this, but I don't think it's worth forcing a failure if they do, hence the warning.

python/lsst/meas/algorithms/loadReferenceObjects.py
cmsaunders: The difference is that the resulting names match the existing schema much more closely, so we don't have to add hundreds of lines elsewhere.

tests/test_referenceObjectLoader.py
```diff
@@ -296,7 +296,7 @@ def setUpClass(cls):
 
         # override some field names.
         config = convertReferenceCatalogTestBase.makeConvertConfig(withRaDecErr=True, withMagErr=True,
-                                                                   withPm=True, withParallax=True)
+                                                                   withPm=True)
```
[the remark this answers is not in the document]
cmsaunders: This is a consequence of changing `ReferenceObjectLoader.makeMinimalSchema` so that `addProperMotion` adds the proper motion parameters **and** their errors. I removed the argument `withPmErr` from `convertReferenceCatalogTestBase.makeConvertConfig` because the proper motion errors are no longer added separately (so `withPm` adds both), so it is also removed here. I can include this in the commit message when I rebase and squash the messages.

tests/test_referenceObjectLoader.py
```diff
@@ -107,7 +107,7 @@ def testMakeMinimalSchema(self):
                     self.assertEqual("photometric" in refSchema, addIsPhotometric)
                     self.assertEqual("coord_raErr" in refSchema, coordErrDim > 0)
                     self.assertEqual("coord_decErr" in refSchema, coordErrDim > 0)
-                    self.assertEqual("coord_ra_dec_Cov" in refSchema, coordErrDim == 3)
+                    self.assertEqual("coord_ra_coord_dec_Cov" in refSchema, coordErrDim == 3)
```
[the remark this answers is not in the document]
cmsaunders: Currently yes (though no, they can't both be set). I addressed this in another comment, we can discuss options.

tests/test_referenceObjectLoader.py
```diff
@@ -162,8 +152,8 @@ def testMakeMinimalSchema_fullCovariance(self):
                                                          coordErrDim=-1)
         # Again, just test a sample of the fields.
         self.assertIn("coord_raErr", schema)
-        self.assertIn("coord_ra_dec_Cov", schema)
-        self.assertIn("coord_pm_ra_parallax_Cov", schema)
+        self.assertIn("coord_ra_coord_dec_Cov", schema)
+        self.assertIn("pm_ra_parallax_Cov", schema)
```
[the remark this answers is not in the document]
cmsaunders: It looks like this comment was not made on the latest version of the code...

tests/test_convertReferenceCatalog.py
```diff
@@ -207,12 +207,6 @@ def testValidateCovariance(self):
                 with self.assertRaises(ValueError, msg=name):
                     config.validate()
 
-        with self.subTest(name="manager"):
-            config = makeConvertConfig(withRaDecErr=True, withParallax=True, withPm=True)
-            config.coordinate_covariance = True
-            with self.assertRaises(FieldValidationError):
-                config.validate()
```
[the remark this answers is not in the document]
cmsaunders: This is not the current version of the code. Is this comment supposed to go with the lines below this that actually do a conversion?

tests/test_referenceObjectLoader.py
```diff
@@ -139,21 +135,10 @@ def testMakeMinimalSchema_fullCovariance(self):
...
-        # coordErrDim is completely ignored for fullCoordinateCovariance=True
-        schema = ReferenceObjectLoader.makeMinimalSchema(filterNameList=["r"],
-                                                         fullCoordinateCovariance=True,
-                                                         addParallax=True,
-                                                         addProperMotion=True,
-                                                         coordErrDim=-1)
-        # Again, just test a sample of the fields.
-        self.assertIn("coord_raErr", schema)
-        self.assertIn("coord_ra_coord_dec_Cov", schema)
-        self.assertIn("pm_ra_parallax_Cov", schema)
```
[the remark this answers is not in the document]
cmsaunders: Because this was to test that `coordErrDim` was being ignored, which is no longer the case. Line 154 would have to be `self.assertNotIn("coord_raErr", schema)` to pass now.

python/lsst/meas/algorithms/convertReferenceCatalog.py
```diff
@@ -202,9 +202,7 @@ class ConvertReferenceCatalogConfig(pexConfig.Config):
     coordinate_covariance = pexConfig.Field(
         dtype=bool,
         doc="Include covariances between sky coordinates, proper motion, and parallax in the schema. "
-            "If true, a custom `ConvertRefcatManager` class must exist to compute the output covariances.",
-        # TODO: We need to check that the manager config is overridden because the default convert manager
-        # will not work for this.
+            "If true, a custom `~ConvertRefcatManager` class must exist to compute the output covariances.",
```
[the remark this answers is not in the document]
cmsaunders: Ok, I thought this would work based on some docs I was reading. I'll try to get the documentation built to see.

python/lsst/meas/algorithms/loadReferenceObjects.py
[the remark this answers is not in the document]
kfindeisen: Sorry, my mistake. This is not the first time that I'd gotten confused about what `diagonalOnly` does, and I missed this one when I was cleaning up.

As for the indent, if it's supposed to be a continuation of "If 3", [then it should probably align with the "I"](https://developer.lsst.io/restructuredtext/style.html#lists). Though I'll admit I haven't checked what the built docs look like.

tests/test_convertReferenceCatalog.py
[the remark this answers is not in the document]
kfindeisen: No, this is the test that you removed while switching `coordinate_covariance` from a precondition to a run-time check on 5ae7698.

I just looked at the final code and all the `assertRaises` are associated with config validation, so I really don't think the case of calling `ConvertRefcatManager._setCoordinateCovariance` is covered anywhere.

tests/test_referenceObjectLoader.py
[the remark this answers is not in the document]
kfindeisen: But this change removes `withParallax`? 😕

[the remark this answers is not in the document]
cmsaunders: `withParallax=True` is from one of the earlier commits on this PR. Including it was necessary because of the side-effects of how `ReferenceObjectLoader.makeMinimalSchema` worked before I refactored the ticket. Parallax is not actually tested in this class.

tests/test_convertReferenceCatalog.py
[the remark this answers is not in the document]
cmsaunders: Ok, got it. I'll find a place to test that.

python/lsst/meas/algorithms/loadReferenceObjects.py
parejkoj: We definitely want this to be a failure, not a warning. No one wanting to do this means it should fail.

doc/lsst.meas.algorithms/reference-catalog-schema.rst
```diff
@@ -30,21 +31,22 @@ Catalog Schema
...
-    - ``pm_ra_parallax_Cov``: covariance between pm_ra and parallax (rad2/year)
-    - ``pm_dec_parallax_Cov``: covariance between pm_dec and parallax (rad2/year)
+    - ``coord_ra_pm_ra_Cov``: covariance between coord_ra and pm_ra (rad^2/year)
+    - ``coord_ra_pm_dec_Cov``: covariance between coord_ra and pm_dec (rad^2/year)
+    - ``coord_ra_parallax_Cov``: covariance between coord_ra and parallax (rad^2/year)
+    - ``coord_dec_pm_ra_Cov``: covariance between coord_dec and pm_ra (rad^2/year)
+    - ``coord_dec_pm_dec_Cov``: covariance between coord_dec and pm_dec (rad^2/year)
+    - ``coord_dec_parallax_Cov``: covariance between coord_dec and parallax (rad^2/year)
+    - ``pm_ra_parallax_Cov``: covariance between pm_ra and parallax (rad^2/year)
+    - ``pm_dec_parallax_Cov``: covariance between pm_dec and parallax (rad^2/year)
```
parejkoj, line 52: Are we sure the units documented above are all correct? I guess the only one that's different is `pm_ra_pm_dec_Cov` (rad^2/year^2), which we did get right above?

cmsaunders: I realized that some of the old units were wrong -- `coord_ra_parallax_Cov` should be rad^2, not rad^2/year -- so I've changed this.

python/lsst/meas/algorithms/convertReferenceCatalog.py
```diff
@@ -64,6 +65,156 @@ def addRefCatMetadata(catalog):
...
+        )
+        schema.addField(
+            field="parallax_flag",
+            type="Flag",
+            doc="Set if parallax or parallax error is bad",
+        )
+        # Add all the off-diagonal covariance terms
+        fields = ["coord_ra", "coord_dec", "pm_ra", "pm_dec", "parallax"]
+        units = ["rad", "rad", "rad/year", "rad/year", "rad"]
+        for i in range(5):
```
parejkoj, line 198: Why are you doing these loops with indices? `for field, unit in zip(itertools.combinations(fields, r=2), itertools.combinations(units, r=2)):` and then e.g. `{field[0]}_{field[1]}_Cov`, should give you all the entries, no double loop or indices necessary.

python/lsst/meas/algorithms/convertReferenceCatalog.py
```diff
@@ -262,6 +421,18 @@ def assertAllOrNone(*names):
             raise ValueError(
                 '"epoch_name" must be specified if "pm_ra/dec_name" or "parallax_name" are specified')
 
+        # Need all the error field names set if we are including covariances.
+        if self.full_position_information:
+            # Since coordinate_covariance is True, this will only pass for the
+            # "None" case
+            assertAllOrNone("full_position_information",
```
parejkoj, line 428: I don't understand the comment above: won't this only pass for the "All" case, since all those fields must be set if `full_position_information` is True?

python/lsst/meas/algorithms/convertReferenceCatalog.py
```diff
@@ -262,6 +421,18 @@ def assertAllOrNone(*names):
             raise ValueError(
                 '"epoch_name" must be specified if "pm_ra/dec_name" or "parallax_name" are specified')
 
+        # Need all the error field names set if we are including covariances.
```
parejkoj, line 424: Why did you remove the check that if `full_position_information` is set, the manager must be overridden?

tests/convertReferenceCatalogTestBase.py
```diff
@@ -41,20 +41,31 @@ def make_coord(ra, dec):
     return lsst.geom.SpherePoint(ra, dec, lsst.geom.degrees)
 
 
-def makeConvertConfig(withMagErr=False, withRaDecErr=False, withPm=False, withPmErr=False,
-                      withParallax=False):
+class ConvertReferenceCatalogCustomClass(ConvertRefcatManager):
+    """Custom class to overload `ConvertRefcatManager._setCoordinateCovariance`
+    """
+    def _setCoordinateCovariance(self, record, row):
+        """"""
```
parejkoj, line 48: Unnecessary

tests/nopytest_convertReferenceCatalog.py
```diff
@@ -55,77 +55,73 @@ class TestConvertReferenceCatalogParallel(convertReferenceCatalogTestBase.Conver
...
+                                       handlers,
+                                       name="testRefCat",
+                                       config=loaderConfig,
+                                       log=self.logger)
+        self.checkAllRowsInRefcat(loader, skyCatalog1, config)
+        self.checkAllRowsInRefcat(loader, skyCatalog2, config)
+
+        inTempDir1.cleanup()
+        inTempDir2.cleanup()
+        dataTempDir.cleanup()
```
parejkoj, line 124: Path cleanups need to happen in `tearDown`

tests/nopytest_convertReferenceCatalog.py
```diff
@@ -250,6 +246,12 @@ def test_getCatalog(self):
         self.assertFloatsEqual(newcat[:len(catalog)]['coord_ra'], catalog['coord_ra'])
         self.assertFloatsEqual(newcat[:len(catalog)]['coord_dec'], catalog['coord_dec'])
 
+    def test_setCoordinateCovariance(self):
+        catalog = self._createFakeCatalog(nOld=10, nNew=0)
+
+        with self.assertRaises(NotImplementedError):
+            self.worker._setCoordinateCovariance(catalog[0], self.fakeInput[0])
```
parejkoj, line 253: I'm not sure this test accomplishes much: we've defined it as not implemented in the base class.

tests/test_convertReferenceCatalog.py
```diff
@@ -95,6 +106,100 @@ def test_main_args_expanded_glob(self):
...
+        self.assertIn("coord_raErr", refSchema)
+        self.assertIn("coord_decErr", refSchema)
+        self.assertIn("coord_ra_coord_dec_Cov", refSchema)
+        self.assertIn("pm_raErr", refSchema)
+        self.assertIn("pm_ra_parallax_Cov", refSchema)
+        self.assertIn("parallaxErr", refSchema)
+        self.assertEqual(refSchema['coord_raErr'].asField().getUnits(), "rad")
+        self.assertEqual(refSchema['coord_ra_coord_dec_Cov'].asField().getUnits(), "rad rad")
+        self.assertEqual(refSchema['pm_raErr'].asField().getUnits(), "rad/year")
+        self.assertEqual(refSchema['pm_dec_parallax_Cov'].asField().getUnits(), "rad/year rad")
```
parejkoj, line 172: I think these fail with how you've reworked the units handling.

tests/test_convertReferenceCatalog.py
```diff
@@ -95,6 +106,100 @@ def test_main_args_expanded_glob(self):
...
+
+    def testCheckFluxUnits(self):
+        """Test that we can identify old style fluxes in a schema."""
```
parejkoj, line 175: This needs to still live in the loader tests, it doesn't have anything to do with the converter.

tests/test_convertReferenceCatalog.py
```diff
@@ -183,6 +285,127 @@ def testValidateParallax(self):
...
+                config.manager.retarget(ConvertGaiaManager)
+                setattr(config, name, None)
+                with self.assertRaises(ValueError, msg=name):
+                    config.validate()
+
+
+class ConvertGaiaManagerTests(convertReferenceCatalogTestBase.ConvertReferenceCatalogTestBase,
+                              lsst.utils.tests.TestCase):
+    """Unittests of methods of ConvertGaiaManager
+    """
```
parejkoj, line 310: This docstring is redundant.

python/lsst/meas/algorithms/convertReferenceCatalog.py
kfindeisen: I requested it [in a previous comment](https://github.com/lsst/meas_algorithms/pull/299#discussion_r1024623834), because using type checks to see if something is implemented is an anti-pattern.

parejkoj: I understand that, but doing the check in validation makes the failure explicit ("you need to set a config to get this functionality"), whereas waiting for the not implemented method to raise will cause a cascade of exceptions from multiprocessing.

python/lsst/meas/algorithms/convertReferenceCatalog.py
cmsaunders: Yeah, should be "all"...somehow it got reversed between my brain and fingers.

python/lsst/meas/algorithms/convertReferenceCatalog.py
cmsaunders: I didn't use this because I'm not really familiar with itertools. I'm fine with changing this.

tests/test_convertReferenceCatalog.py
cmsaunders: This is already fixed.

python/lsst/meas/algorithms/convertReferenceCatalog.py
```diff
@@ -64,6 +65,156 @@ def addRefCatMetadata(catalog):
...
+def _makeSchema(filterNameList, *, addCentroid=False,
+                addIsPhotometric=False, addIsResolved=False,
+                addIsVariable=False, fullPositionInformation=False):
+    """Make a standard schema for reference object catalogs.
+
+    Parameters
+    ----------
+    filterNameList : `list` of `str`
+        List of filter names. Used to create <filterName>_flux fields.
+    addIsPhotometric : `bool`
```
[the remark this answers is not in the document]
cmsaunders: That's what copying and pasting gets you--it was missing in `makeMinimalSchema` too.

tests/nopytest_convertReferenceCatalog.py
cmsaunders: This is what was already there, and these aren't class variables.

tests/nopytest_convertReferenceCatalog.py
cmsaunders: I'm pretty sure this test was requested in earlier comments.

tests/nopytest_convertReferenceCatalog.py
parejkoj: Then we should probably make them so, otherwise if the test fails, it won't cleanup.

kfindeisen: An alternative that should make @parejkoj happy while keeping them isolated to the test that actually uses them is to use the original `TemporaryDirectory` calls as (triply-nested 😬) context managers.

parejkoj: You can put all the context managers inside the same `with`, so you don't have to nest them. That would work.

cmsaunders: @parejkoj , could you give me a code snippet for that?

parejkoj: `with tempfile.TemporaryDirectory() as path1`, with tempfile.TemporaryDirectory() as path2:`, etc.

cmsaunders: This doesn't work. It looks like you aren't supposed to repeat `with`, but even after fixing that, I'm not getting a `TemporaryDirectory()` object with this--I just get a string with the directory name. I'm just going to go back to the option of making these class variables and adding `setup` and `teardown` functions.

parejkoj: Sorry, yes, you don't repeat `with`. When used as a context manager, the directory name is the target of the `with`.

See the docs:

https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory

kfindeisen (approved):
Looks good, just one minor bug and some style comments. Thank you for your patience through the multiple iterations!

python/lsst/meas/algorithms/convertReferenceCatalog.py
```diff
@@ -64,6 +66,157 @@ def addRefCatMetadata(catalog):
...
+    filterNameList : `list` of `str`
+        List of filter names. Used to create <filterName>_flux fields.
+    addCentroid : `bool`
+        If True then add fields "centroid" and "hasCentroid".
+    addIsPhotometric : `bool`
+        If True then add field "photometric".
+    addIsResolved : `bool`
+        If True then add field "resolved".
+    addIsVariable : `bool`
+        If True then add field "variable".
```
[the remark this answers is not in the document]
cmsaunders: These parameter names are copied from the original `makeMinimalSchema`. I'm fine with changing them since we're changing so much already, unless @parejkoj knows of some historical reason for having "Is" in the name.

tests/test_convertReferenceCatalog.py
```diff
@@ -183,6 +256,127 @@ def testValidateParallax(self):
...
+            outputCatalog[name] = array
+
+        for outputRow, inputRow in zip(outputCatalog, self.fakeInput):
+            self.worker._setProperMotion(outputRow, inputRow)
+            self.worker._setParallax(outputRow, inputRow)
+            self.worker._setCoordinateCovariance(outputRow, inputRow)
+
+        coordConvert = (self.worker.coord_err_unit).to(u.radian)
+        pmConvert = (self.worker.config.pm_scale * u.milliarcsecond).to_value(u.radian)
+        parallaxConvert = (self.config.parallax_scale * u.milliarcsecond).to_value(u.radian)
```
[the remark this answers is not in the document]
cmsaunders: I changed them to all be `self.worker.config`. As written, this doesn't fail because `self.worker.config` is `self.config`, but obviously it ought to be consistent.

python/lsst/meas/algorithms/convertReferenceCatalog.py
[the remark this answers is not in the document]
cmsaunders: Thinking about this a little more, I think keeping the "Is" would be better, since these correspond to adding boolean fields. To me "addVariable", for example, makes me wonder if some scalar parameter is added, whereas "addIsVariable" more clearly implies that the resulting flag with correspond to "Is Variable" or "Is Not Variable".

PR #300, opened 2022-11-28
DM-37102: convertGaiaXpManager added

psferguson:


python/lsst/meas/algorithms/convertRefcatManager.py
```diff
@@ -454,3 +454,35 @@ def gaiaFluxToFlux(flux, zeroPoint):
         result['phot_rp_mean_fluxErr'] = result['phot_rp_mean_flux'] / input['phot_rp_mean_flux_over_error']
 
         return result
+
+class ConvertGaiaXpManager(ConvertRefcatManager):
+    """Special-case convert manager to deal with refcats that have no MagErrors
```
parejkoj, line 459: Maybe: "Special-case convert manager for Gaia XP spectrophotometry catalogs, that have fluxes/flux errors, instead of magnitudes."

python/lsst/meas/algorithms/convertRefcatManager.py
```diff
@@ -454,3 +454,35 @@ def gaiaFluxToFlux(flux, zeroPoint):
...
+        for item in self.config.mag_column_list:
+            
+            inFluxKey=item.replace('_','_flux_')
+            outFluxKey=item + '_flux'
+            
+            inFluxErrKey=item.replace('_','_flux_error_')
+            outFluxErrKey=item + '_fluxErr'
+            
+            result[outFluxKey] = (inputData[inFluxKey]*u.Watt/u.Hz/u.meter/u.meter).to(u.nJy)
+            result[outFluxErrKey] = (inputData[inFluxErrKey]*u.Watt/u.Hz/u.meter/u.meter).to(u.nJy)
```
parejkoj, line 486: I think you need `to_value(u.nJy)` here to get a float array, instead of a Quantity array.

python/lsst/meas/algorithms/convertRefcatManager.py
```diff
@@ -454,3 +454,35 @@ def gaiaFluxToFlux(flux, zeroPoint):
...
+            The values that will go into the flux and fluxErr fields in the
+            output catalog, units are nJy
+        """
+        result = {}
+        for item in self.config.mag_column_list:
+            
+            inFluxKey=item.replace('_','_flux_')
+            outFluxKey=item + '_flux'
+            
+            inFluxErrKey=item.replace('_','_flux_error_')
```
parejkoj, line 482: What is this for? The `mag_column_list` should have the actual name of the input fields, I would think. What are the names of the input flux fields in the xpSpec files?

python/lsst/meas/algorithms/convertRefcatManager.py
```diff
@@ -454,3 +454,35 @@ def gaiaFluxToFlux(flux, zeroPoint):
         result['phot_rp_mean_fluxErr'] = result['phot_rp_mean_flux'] / input['phot_rp_mean_flux_over_error']
 
         return result
+
+class ConvertGaiaXpManager(ConvertRefcatManager):
+    """Special-case convert manager to deal with refcats that have no MagErrors
+    """
+    def _getFluxes(self, inputData):
+        """Compute the flux fields that will go into the output catalog.
```
parejkoj, line 462: I would leave the whole docstring for this method empty, so that it uses the parent class docstring when building the docs, since the inputs/outputs are the same.

python/lsst/meas/algorithms/convertRefcatManager.py
```diff
@@ -454,3 +454,35 @@ def gaiaFluxToFlux(flux, zeroPoint):
...
+
+        Parameters
+        ----------
+        inputData : `numpy.ndarray`
+            The input data to compute fluxes for.
+            from Gaia Collaboration, Montegriffo et al. 2022 flux units are  W/Hz/(m^2)
```
parejkoj, line 468: Move this into the class docstring, as a a sentence

parejkoj (changes requested):
Note: the PR needs to include the ticket number; see:

https://developer.lsst.io/work/flow.html#make-a-pull-request

Also, the linter failed: that will have to be fixed before merging.

python/lsst/meas/algorithms/convertRefcatManager.py
psferguson: Example xp photometry column names are:
 'Lsst_flux_u',
 'Lsst_flux_error_u'
 I had thought we would want output to be:
 'Lsst_u_flux' or 'Lsst_u_fluxErr'
 But instead we want to preserve the original column name the fluxes come from in the xp catalog so output will be 
 'Lsst_flux_u_flux' or 'Lsst_flux_u_fluxErr'

parejkoj (approved):
Thank you, this looks better.

Though, note the missing colon in the PR title.

PR #301, opened 2022-11-30
u/parejkoj: Fix SourceDetectionTask docstrings

parejkoj:
While looking at this code with Merlin, I noticed a number of incorrect or misleading docstrings. I gave it an overall cleanup pass here.

### DIFF
--- a/python/lsst/meas/algorithms/detection.py
+++ b/python/lsst/meas/algorithms/detection.py
@@ -160,14 +160,14 @@ class SourceDetectionConfig(pexConfig.Config):
 
 
 class SourceDetectionTask(pipeBase.Task):
-    """Create the detection task.  Most arguments are simply passed onto pipe.base.Task.
+    """Detect peaks and footprints of sources in an image.
 
     Parameters
     ----------
     schema : `lsst.afw.table.Schema`
         Schema object used to create the output `lsst.afw.table.SourceCatalog`
     **kwds
-        Keyword arguments passed to `lsst.pipe.base.task.Task.__init__`
+        Keyword arguments passed to `lsst.pipe.base.Task.__init__`
 
     If schema is not None and configured for 'both' detections,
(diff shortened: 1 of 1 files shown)


python/lsst/meas/algorithms/detection.py
```diff
@@ -160,25 +160,24 @@ def setDefaults(self):
 
 
 class SourceDetectionTask(pipeBase.Task):
-    """Create the detection task.  Most arguments are simply passed onto pipe.base.Task.
+    """Detect peaks and footprints of sources on an image.
```
mfisherlevine, line 163: I'd have said `in an image` but doesn't really matter.

python/lsst/meas/algorithms/detection.py
```diff
@@ -392,7 +408,7 @@ def calculateKernelSize(self, sigma):
         return (int(sigma * self.config.nSigmaForKernel + 0.5)//2)*2 + 1  # make sure it is odd
 
     def getPsf(self, exposure, sigma=None):
-        """Retrieve the PSF for an exposure
+        """Generate or retrieve a PSF for an exposure.
```
mfisherlevine, line 411: I find the use of "generate" potentially a little confusing. I assume this means that if one is premade and lives on the object then it is returned, and if something behind it needs some sort of making, then that is done, right? But will a new user think like that, or might that suggest some sort of fitting to "generate" the PSF? Just thinking aloud though, so feel free to do as you like.

mfisherlevine (commented):
Yeah, looks great. Two tiny comments to do as you please with.

python/lsst/meas/algorithms/detection.py
parejkoj: Thanks for asking about this: I dug into it more and it left me with some questions that I've asked on #dm-science-pipelines. I've reworded to better reflect what the code actually does right now.

PR #302, opened 2022-12-02
u/parejkoj: Add source detection task doc paragraph

parejkoj:
The fact that we're always convolving with a Gaussian needs to be more obvious in the task docs.


TallJimbo (approved):
Looks good, thanks!

PR #303, opened 2022-12-14
DM-37351: Add fill_values to text file reader

cmsaunders:


tests/data/testReadTextCatalog_nullValues.csv
```diff
@@ -0,0 +1,3 @@
+name, ra, dec, counts, flux, resolved
+£object1, -5.0, 10.0, 1000, 1.1, 1
```
parejkoj, line 2: Do you want the pound-symbol at the start of this?

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -54,6 +54,12 @@ class ReadTextCatalogConfig(pexConfig.Config):
         doc=("Format of files to read, from the astropy.table I/O list here:"
              "http://docs.astropy.org/en/stable/io/unified.html#built-in-table-readers-writers")
     )
+    fill_values = pexConfig.ListField(
+        dtype=str,
+        default=[],
```
parejkoj, line 59: Default `[]` or `None`?

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -54,6 +54,12 @@ class ReadTextCatalogConfig(pexConfig.Config):
...
+        doc=("A list giving [<match_string>, '0', <optional column name>], which is used to mask"
+             " the given values in the input file.")
```
parejkoj, line 61: Please reference the astropy doc that gives more detail on this, since this is really an astropy setting.

Why can't we use `NaN` for the fill value? The Mask Tables look like they support a per-column fill_value.

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -54,6 +54,12 @@ class ReadTextCatalogConfig(pexConfig.Config):
         doc=("Format of files to read, from the astropy.table I/O list here:"
              "http://docs.astropy.org/en/stable/io/unified.html#built-in-table-readers-writers")
     )
+    fill_values = pexConfig.ListField(
+        dtype=str,
+        default=[],
+        doc=("A list giving [<match_string>, '0', <optional column name>], which is used to mask"
```
parejkoj, line 60: What is the `'0'` for here? It looks like that's the value to fill with; since there's an "optional column name", can't we have this be per-column, and thus use `NaN`?

parejkoj (commented):
Questions about docs and using NaN

tests/data/testReadTextCatalog_nullValues.csv
cmsaunders: Not needed. This was copied from the `testReadTextCatalog.csv`, but I'm not checking non-ASCII characters here, so I'll remove it.

python/lsst/meas/algorithms/readTextCatalogTask.py
cmsaunders: I changed this to `None` and added an `optional=True` so it passes validation.

python/lsst/meas/algorithms/readTextCatalogTask.py
cmsaunders: If you put `Nan` in, the columns that should be integers get converted to floats. For astropy, you can do per-column `fill_value`s, but then we need to use `pexConfig.ConfigDictField`, and will have to know all the column names and set them individually in the task configuration. It's a little annoying, but maybe worth it. I'll try implementing it.

python/lsst/meas/algorithms/readTextCatalogTask.py
cmsaunders: This is what is recommended by astropy (https://docs.astropy.org/en/stable/io/ascii/read.html#f1), but not actually required. If the `ConfigDictField` works out, I'll change this.

cmsaunders:
I spent way too long trying to get the `Table.read()` `fill_values` to work per-column, but it doesn't. There's a github issue here https://github.com/astropy/astropy/issues/900 that goes into it. You can use `fill_values` in `read()`, then go through each column and set the `fill_value` after the fact, which only changes `Table.filled`, so it starts to get pretty messy. 
I decided to take a different approach and just have a config parameter that lets you turn on replacing missing floats with some other value.

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -54,6 +54,20 @@ class ReadTextCatalogConfig(pexConfig.Config):
         doc=("Format of files to read, from the astropy.table I/O list here:"
              "http://docs.astropy.org/en/stable/io/unified.html#built-in-table-readers-writers")
     )
+    fill_values = pexConfig.ListField(
+        dtype=str,
+        default=None,
+        optional=True,
+        doc=("A list giving [<match_string>, <fill_value>], which is used to mask"
+             " the given values in the input file. '0' is suggested for the fill value in order to prevent"
+             "changing the column datatype. The default behavior is to fill empty data with zeros. See "
```
parejkoj, line 63: Missing initial space.

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -54,6 +54,20 @@ class ReadTextCatalogConfig(pexConfig.Config):
...
+             "https://docs.astropy.org/en/stable/io/ascii/read.html#bad-or-missing-values for more details.")
+    )
+    replace_missing_floats = pexConfig.Field(
+        dtype=float,
+        optional=True,
+        doc="If True, replace missing data in float columns with given value instead of zero."
```
parejkoj, line 69: The docstring doesn't match the behavior below: below, you always fill missing floats with NaN, which I'd argue is the correct behavior (instead of "with the given value"). Null float fields should become NaN for the purposes of this code. `replace_missing_floats` should then just be a boolean field.

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -84,7 +98,19 @@ def run(self, filename):
...
+        table = Table.read(filename, format=self.config.format,
+                           delimiter=self.config.delimiter,
+                           **kwargs)
+
+        # convert to a numpy array for backwards compatibility with other readers
+        arr = np.array(table.as_array())
+
+        if self.config.replace_missing_floats:
+            for column in table.columns:
+                if table.dtype[column] == np.float64:
```
parejkoj, line 113: Probably worth doing this for both float64 and float32.

parejkoj (commented):
Should the `fill_values` and `replace_missing_floats` configs refer to each other? How they interact isn't entirely obvious from the docs.

python/lsst/meas/algorithms/readTextCatalogTask.py
cmsaunders: Yeah, I started with this being a boolean, but then I thought I should just leave it up to the user, so I changed it to a float field (and then forgot to actually change the code). NaN is of course the most likely value wanted here, but I don't think there's any harm in allowing flexibility in choosing the replacement value here.

parejkoj: I think this should always fill with NaN. Making this configurable is harder for the user (e.g. they have to import numpy and set this to np.nan), and NaN is 100% the correct floating point value for "there is no value".

cmsaunders:
> Should the `fill_values` and `replace_missing_floats` configs refer to each other? How they interact isn't entirely obvious from the docs.

I thought I responded to this, but I don't see a comment. These can be used independently, but I elaborated on their interaction in the config docs.

python/lsst/meas/algorithms/readTextCatalogTask.py
```diff
@@ -54,6 +54,22 @@ class ReadTextCatalogConfig(pexConfig.Config):
...
+             " the given values in the input file. '0' is suggested for the fill value in order to prevent"
+             " changing the column datatype. The default behavior is to fill empty data with zeros. See "
+             "https://docs.astropy.org/en/stable/io/ascii/read.html#bad-or-missing-values for more details."
+             "Use `replace_missing_floats_with_nan` to change floats to NaN instead of <fill_value>.")
+    )
+    replace_missing_floats_with_nan = pexConfig.Field(
+        dtype=bool,
+        default=False,
+        doc="If True, replace missing data in float columns with NaN instead of zero. If `fill_values` is "
+            "set, this parameter with replace the floats masked by `fill_values`."
```
parejkoj, line 71: "with replace"->"will replace"

I don't think I understand that final sentence. Is this implying `replace_missing_floats_with_nan` takes precedence over `fill_values` (which I think is what we want)?

parejkoj (approved):
One final question about wording. Thanks for the fixes.

parejkoj:
The PR ticket number is incorrect: should be [DM-37351](https://jira.lsstcorp.org/browse/DM-37351)

PR #304, opened 2022-12-19
DM-37415: Add debugging log output of filenames to convertRefcat

parejkoj:


psferguson (approved):
This was very useful for identifying the catalog files that were failing the conversion.

PR #306, opened 2023-02-03
DM-37791: use pybind wrappers, merge into one pybind shared lib

mwittgen:


erykoff (approved):
This all looks good.  I did think that all the `addInheritanceDependency()` calls were supposed to be moved to `_algorithmsLib.cc`?

erykoff:
Also, what was the final decision on stub files to allow for easier backwards compatibility?

mwittgen:
> This all looks good. I did think that all the `addInheritanceDependency()` calls were supposed to be moved to `_algorithmsLib.cc`?

Indeed, will fix.

mwittgen:
> Also, what was the final decision on stub files to allow for easier backwards compatibility? Or was that about `meas_base`...

Checking with @ktlim and @timj to make sure.  For `meas_base` and `meas_algorithm` there were not many direct imports of shlibs in other packages, so I didn't add the stub. There was no concern about old config files used with newer pipeline versions.

timj:
> Also, what was the final decision on stub files

We discussed this and we decided that the config case was a red herring since we already have configs that can only be read by the software that wrote them (although I would like @mwittgen to put fixing the excessive import problem on his todo list). We also noted that if anything is doing `from .x import *` in the `__init__.py` that the `x.py` is not really part of the public API even though it could be used like that. We therefore decided it was okay to drop the .py files.

PR #309, opened 2023-02-28
DM-37955: Update measureApCorrTask to use robust median outlier rejection.

erykoff:
This replaces the default source selector to use all possible (unflagged) stars, and additional checks on the fit degree after outlier rejection.

### DIFF
--- a/python/lsst/meas/algorithms/measureApCorr.py
+++ b/python/lsst/meas/algorithms/measureApCorr.py
@@ -23,7 +23,8 @@
 
 __all__ = ("MeasureApCorrConfig", "MeasureApCorrTask")
 
-import numpy
+import numpy as np
+from scipy.stats import median_abs_deviation
 
 import lsst.pex.config
 from lsst.afw.image import ApCorrMap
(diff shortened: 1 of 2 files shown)


python/lsst/meas/algorithms/measureApCorr.py
```diff
@@ -34,32 +35,39 @@
...
-                read: {name}_instFlux, {name}_instFluxErr, {name}_flag
-                added: apcorr_{name}_used
-        """
-        self.flux = schema.find(name + "_instFlux").key
-        self.err = schema.find(name + "_instFluxErr").key
-        self.flag = schema.find(name + "_flag").key
-        self.used = schema.addField("apcorr_" + name + "_used", type="Flag",
-                                    doc="set if source was used in measuring aperture correction")
+        self.fluxName = name + "_instFlux"
+        if not schema.extract(self.fluxName):
```
TallJimbo, line 56: `if self.fluxName not in schema:` is more idiomatic, and I sure hope it's not slower.

python/lsst/meas/algorithms/measureApCorr.py
```diff
@@ -164,103 +195,126 @@ def run(self, exposure, catalog):
...
-                apCorrData = apCorrData[keep]
-                indices = indices[keep]
 
-            # Final fit after clipping
-            apCorrField = ChebyshevBoundedField.fit(bbox, x, y, apCorrData, ctrl)
+            allBad = False
+            for iteration in range(self.config.numIter):
+                resid = z - fitValues
+                # We allow floating-point slop to allow for field estimation
+                # noise in a completely flat field.
```
TallJimbo, line 253: Does this avoid a divide-by-zero somewhere?  I think that means a slightly longer comment here would be useful.

tests/test_measureApCorr.py
```diff
@@ -199,6 +207,49 @@ def testApertureMeasTens(self):
...
+        self.meas_apCorr_task.config.minDegreesOfFreedom = 4
+
+        for name in self.names:
+            nameAp = name + self.apNameStr
+            sourceCat[nameAp + "_instFlux"][0] = 100.0
+
+        with self.assertRaisesRegex(RuntimeError, "only 4 sources remain"):
+            self.meas_apCorr_task.run(catalog=sourceCat, exposure=self.exposure)
+
+        # With the measurement algorithm declared as something that might fail, should not get an exception
```
TallJimbo, line 248: Something in this sentence seems awry.

TallJimbo (approved):

python/lsst/meas/algorithms/measureApCorr.py
erykoff: I don't think so, but I've rephrased the comment to be clear this shows up in tests when the residual field is completely flat and median_abs_deviation doesn't return identically zero in that case.

tests/test_measureApCorr.py
erykoff: I copied it from another test, but I have rephrased both of them to be clearer.

PR #312, opened 2023-04-12
DM-23781: Improve Sky Object Placement

leeskelvin:


python/lsst/meas/algorithms/skyObjects.py
```diff
@@ -33,9 +35,11 @@ def generateSkyObjects(mask, seed, config):
...
     as `DETECTED`).
 
-    The algorithm for determining sky objects is random trial and error:
-    we try up to `nTrialSkySources` random positions to find `nSources`
-    sky objects.
+    Sky objects are positioned using a quasi-random Halton sequence number
+    generator. This is a deterministic sequence that mimics a random trial and
+    error approach whilst acting to minimize clustering of points for a given
+    field of view. Up to `nTrialSources` points are generated, returning the
+    first `nSources` that do not overlap with the mask.
```
laurenam, line 42: RE: 
> "acting to minimize clustering of points"

does this algorithm guarantee no overlapping positions?  If not, you might consider adding:

```
# Add doubled-in-size sky object spanSet to the avoid mask.
avoid = avoid.union(spans.dilated(int(skySourceRadius)))
```
after the line:

```
skyFootprints.append(fp)
```
at line 93  below in order to have a cumulative `avoid` mask.  It worked nicely using the old trial 'n error algorithm (but may not be necessary now?)

tests/test_dynamicDetection.py
```diff
@@ -57,7 +57,7 @@ def setUp(self):
 
         # Relative tolerance for tweak factor
         # Not sure why this isn't smaller; maybe due to use of Poisson instead of Gaussian noise?
-        self.rtol = 0.1
+        self.rtol = 0.15
```
laurenam, line 60: Might be worth a comment in the commit message about this change (not that we can really explain it at the moment...😞).

laurenam (approved):

python/lsst/meas/algorithms/skyObjects.py
leeskelvin: There is nothing explicit in the SciPy Halton Sequence implementation that prevents overlapping positions as N tends to infinity, however, it is (by design) exceptionally unlikely.

However, I do like your additional logic to prevent this and I don't think it adds too much overhead. As `SkyObjectsConfig` already has a `growMask` field, I'm tempted to recommend using this information rather than hard-coding a double-in-size condition?

tests/test_dynamicDetection.py
leeskelvin: I thought about adding one, but the initial in-line that's already there seems to capture everything I would want to say anyway. In the spirit of adding something, I've added this:
```
# It seems as if some sky objects are being placed in the extra
# background region, which is causing the offset between the expected
# factor and the measured factor to be larger than otherwise expected.
# This relative tolerance was increased from 0.1 to 0.15 on DM-23781 to
# account for this.
```

laurenam: Looks good (I did just mean adding something to the commit message, but this works too!)

python/lsst/meas/algorithms/skyObjects.py
leeskelvin: Following testing, I think I convinced myself that I prefer your initial idea of growing the sky object footprints by their own radius. I experimented with growing the `growMask` parameter, and while it works, it occurred to me that that parameter was intended to mask the flux in the wings of astrophysical sources. We want to grow sky object footprints for a different reason altogether (to avoid clustering for stats purposes). To that end, I've adopted your initial recommendation above as-is.

tests/test_dynamicDetection.py
leeskelvin: Ah, the _commit_ message! For some reason I didn't parse that the first time around. Commit message also updated with some background too.

PR #316, opened 2023-04-19
DM-38753: Use InMemoryDatasetHandle for mock ref cat handle

timj:


timj:
Classes similar to `MockRefcatDataId` are defined in a couple of tests in analysis_tools and drp_tasks. In theory we could modify the version here so it just takes a region and then remove then change the other packages so they use the one from here. Not entirely sure it's worth it but I'm happy to make the change.

erykoff (approved):

PR #317, opened 2023-04-19
DM-38769: Exclude gen3.sqlite3 from MemoryTestCase.

ktlim:


ktlim:
Doing this in `lsst.utils` instead.

andy-slac:
> Doing this in `lsst.utils` instead.

@ktlim, I posted something on JIRA (in case JIRA takes time to notify you).

PR #318, opened 2023-04-21
DM-38667: Putting a condition on the minimum number of valid pixels within the normalisation annulus

bazkiaei:
This condition results in the rejection of bright star stamps with less valid pixels within the normalisation annulus than what the user determines.

In addition, some modifications have been made to reject stamps with a negative normalisation factor.


bazkiaei:
This changes from this PR is merged to the stack through PR #328

PR #322, opened 2023-04-27
DM-38472: Reformat Bright Star Subtraction Processing Tasks

leeskelvin:


python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -397,8 +441,7 @@ def extend(self, bss):
             Other instance to concatenate.
         """
         if not isinstance(bss, BrightStarStamps):
-            raise ValueError('Can only extend with a BrightStarStamps object.  '
-                             f'Got {type(bss)}.')
+            raise ValueError("Can only extend with a BrightStarStamps object.  " f"Got {type(bss)}.")
```
arunkannawadi, line 444: The f part of this `f-string` can come up front and be a single string.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -475,30 +522,40 @@ def _checkNormalization(self, normalize, innerRadius, outerRadius):
...
+                raise AttributeError(
+                    "For stamps to be normalized (normalize=True), please provide a valid "
+                    "value (in pixels) for both innerRadius and outerRadius."
+                )
             elif noneFluxCount < nStamps:
-                raise AttributeError(f"{nFluxVals} stamps already contain an annularFlux value. For stamps to"
-                                     " be normalized, all their annularFlux must be None.")
+                raise AttributeError(
+                    f"{nFluxVals} stamps already contain an annularFlux value. For stamps to"
+                    " be normalized, all their annularFlux must be None."
```
arunkannawadi, line 541: It's a convention to keep the space at the end of the previous string.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -475,30 +522,40 @@ def _checkNormalization(self, normalize, innerRadius, outerRadius):
...
-                                     "value (in pixels) for both innerRadius and outerRadius.")
+                raise AttributeError(
+                    "For stamps to be normalized (normalize=True), please provide a valid "
+                    "value (in pixels) for both innerRadius and outerRadius."
+                )
             elif noneFluxCount < nStamps:
-                raise AttributeError(f"{nFluxVals} stamps already contain an annularFlux value. For stamps to"
-                                     " be normalized, all their annularFlux must be None.")
+                raise AttributeError(
+                    f"{nFluxVals} stamps already contain an annularFlux value. For stamps to"
```
arunkannawadi, line 540:
```suggestion
                    f"{nFluxVals} stamps already contain an annularFlux value. For stamps to "
```

python/lsst/meas/algorithms/brightStarStamps.py
arunkannawadi, line 541:
```suggestion
                    "be normalized, all their annularFlux must be None."
```

arunkannawadi (approved):
Looks good

timj:
Did black get run on this file? The file seems to have had substantial reorganizing. I'm not sure if there are code differences other than black that are not obvious.

leeskelvin:
Yes @timj, this refactor is entirely a clean up exercise. Both `black` and `isort` have been run, in addition to some basic doc string updates. This effort is to facilitate future collaborative efforts on this file between multiple authors. I wanted the file to be cleaned up to help with that effort to minimize the impact future commits would have.

timj:
@leeskelvin can you please make that explicit in the commit message? "refactor" is not, to me, the same as reformatting.

Note that because this repo doesn't have any formatting checks it is very easy for the file to start to drift away from black and isort.

leeskelvin:
I'll change the ticket name to "Reformat" accordingly, thanks.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -82,30 +81,36 @@ def factory(cls, stamp_im, metadata, idx, archive_element=None):
             needed by the constructor.
         idx : `int`
             Index into the lists in ``metadata``
-        archive_element : `lsst.afwTable.io.Persistable`, optional
+        archive_element : `lsst.afw.table.io.Persistable`, optional
```
arunkannawadi, line 84:
```suggestion
        archive_element : `~lsst.afw.table.io.Persistable`, optional
```

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -189,17 +193,29 @@ class BrightStarStamps(Stamps):
...
+    >>> starSubregions = butler.get(
+            "brightStarStamps",
+            dataId,
+            parameters={"bbox": bbox}
+        )
     """
 
-    def __init__(self, starStamps, innerRadius=None, outerRadius=None, nb90Rots=None,
-                 metadata=None, use_mask=True, use_variance=False, use_archive=False):
+    def __init__(
```
arunkannawadi, line 208: I find the placement of this weird. I'd like it either before or immediately after the `factory` method above.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -204,7 +202,7 @@ def factory(cls, stamp_im, metadata, index, archive_element=None):
             needed by the constructor.
         idx : `int`
             Index into the lists in ``metadata``
-        archive_element : `lsst.afwTable.io.Persistable`, optional
+        archive_element : `lsst.afw.table.io.Persistable`, optional
```
arunkannawadi, line 205:
```suggestion
        archive_element : `~lsst.afw.table.io.Persistable`, optional
```

python/lsst/meas/algorithms/stamps.py
```diff
@@ -251,21 +249,29 @@ def factory(cls, stamp_im, metadata, index, archive_element=None):
             needed by the constructor.
         idx : `int`
             Index into the lists in ``metadata``
-        archive_element : `afwTable.io.Persistable`, optional
+        archive_element : `afw.table.io.Persistable`, optional
```
arunkannawadi, line 252:
```suggestion
        archive_element : `~lsst.afw.table.io.Persistable`, optional
```

python/lsst/meas/algorithms/stamps.py
```diff
@@ -292,17 +298,21 @@ class StampsBase(abc.ABC, Sequence):
...
+        )
     """
 
-    def __init__(self, stamps, metadata=None, use_mask=True, use_variance=True,
-                 use_archive=False):
-        if not hasattr(stamps, '__iter__'):
-            raise ValueError('The stamps parameter must be iterable.')
+    def __init__(self, stamps, metadata=None, use_mask=True, use_variance=True, use_archive=False):
+        if not hasattr(stamps, "__iter__"):
+            raise ValueError("The stamps parameter must be iterable.")
```
arunkannawadi, line 310: Is this check necessary? We don't normally check and raise an error in our code. Moreover, this base class is probably called only within our own code, so we should be able to quickly identify the error if we call it incorrectly.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -292,17 +298,21 @@ class StampsBase(abc.ABC, Sequence):
...
         for stamp in stamps:
             if not isinstance(stamp, AbstractStamp):
-                raise ValueError('The entries in stamps must inherit from AbstractStamp.  '
-                                 f'Got {type(stamp)}.')
+                raise ValueError(
+                    "The entries in stamps must inherit from AbstractStamp.  " f"Got {type(stamp)}."
+                )
```
arunkannawadi, line 315: Similarly, this check may also not be required and is inconsistent with duck typing. But if you choose to keep it, then the second string needs to be merged with the first.

python/lsst/meas/algorithms/stamps.py
```diff
@@ -334,12 +344,10 @@ def readFitsWithOptions(cls, filename, options):
         """
         # To avoid problems since this is no longer an abstract method
         if cls is not StampsBase:
-            raise NotImplementedError(
-                f"Please implement specific FITS reader for class {cls}"
-            )
+            raise NotImplementedError(f"Please implement specific FITS reader for class {cls}")
```
arunkannawadi, line 347: Either here, or in a future ticket (comment it here with a TODO in that case), this method should be refactored. The logic looks very convoluted and I don't follow what's going on here. If `readFitsWithOptions` is called on `StampBase`, then something happens but if called on a child class (Stamps), then an error is raised? And `cls` is overridden below, but why not call the method directly on `stamp_type`?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -398,7 +413,7 @@ def getArchiveElements(self):
         Returns
         -------
         archiveElements :
-            `list` [`lsst.afwTable.io.Persistable`]
+            `list` [`lsst.afw.table.io.Persistable`]
```
arunkannawadi, line 416:
```suggestion
            `list` [~`lsst.afw.table.io.Persistable`]
```

python/lsst/meas/algorithms/stamps.py
```diff
@@ -370,8 +378,15 @@ def writeFits(self, filename):
...
-                  self.use_archive)
+        writeFits(
+            filename,
+            self._stamps,
+            self._metadata,
+            type_name,
+            self.use_mask,
+            self.use_variance,
+            self.use_archive,
+        )
```
arunkannawadi, line 389: I don't quite see why to call `writeFits` method that was defined earlier in the file, rather than implementing the method here. If you agree with here, create a ticket (perhaps the same one that refactors `readFitsWithOptions`) and comment with TODO.

arunkannawadi (approved):

python/lsst/meas/algorithms/stamps.py
timj: Modulo fixing the black string formatting problem here, I don't understand what the problem is with this test? We are requiring that everything is an AbstractStamp as a way to fail early. It's effectively promising the interface.

python/lsst/meas/algorithms/stamps.py
timj: You can probably drop it because the line below `for stamp in stamps` is going to prove whether or not it is iterable -- it clearly has to be. I believe this check was added because there were bugs being triggered by people passing in a single stamp rather than list of stamp. The error message is more explicit  than waiting for the next line to raise a generic error.

python/lsst/meas/algorithms/stamps.py
timj: I think it's related to the trampoline to allow the butler formatter to use a generic type but still end up giving you the correct type and making sure that you don't by mistake end up trying to use the abstract base class for reading.

python/lsst/meas/algorithms/stamps.py
arunkannawadi: I don't think it's Pythonic to do such `isinstance` checks, and we don't check that the object passed is of the correct type in most of our code. It makes sense only if there are other classes that will conform to this interface, yet will do something completely different. This is likely the case where the original author saw a different codebase and mimicked the same structure here. Is failing early here (occassionally) advantageous over checking this every time at run time?

timj: I think it may have been my idea to add this check to be defensive. We were getting programming bugs where people were handing the wrong thing to the constructor and it was hard to track down where the bad stamp was coming from.

arunkannawadi: All the more reason to start using static type checking more broadly :-)

If the branching had to depend on the type of the object, then an `isinstance` check makes sense. If it is simply to check if it is the right type, then these should be left to mypy. What's preventing us from enabling mypy in all our repos, with one function with type hints just to get the ball rolling?

timj: It is unlikely that the entire science pipelines is going to be mypy compliant any time soon and it's not clear to me that the science pipelines group have decided they want mypy.

I don't understand why you don't want to check that these are all subclasses of Stamp.

python/lsst/meas/algorithms/brightStarStamps.py
leeskelvin: This is the `__init__` for the BrightStarStamps class, and the first thing in this class - did you have in mind somewhere else specifically for this to go?

python/lsst/meas/algorithms/stamps.py
leeskelvin: I've removed this `hasattr` check. I agree, it doesn't quite fit here, and the for loop immediately following will trigger a `TypeError` if a non-iterable type has been passed.

python/lsst/meas/algorithms/stamps.py
leeskelvin: I'm reluctant to remove this `isinstance` check without a broader understanding of how this may be being used elsewhere in the stack. As it is, I like the idea of failing fast if the wrong type is being passed in, so I lean towards maintaining this behavior for the time being. Happy to be persuaded it's safe to remove this however.

python/lsst/meas/algorithms/brightStarStamps.py
leeskelvin: Unless you're referring to the `__len__`, `__getitem__` and `__iter__` methods below, which could all be moved up?

python/lsst/meas/algorithms/stamps.py
leeskelvin: The `writeFits` function at the top of the file is a free function, whereas this `writeFits` method is a shortcut for writing a `StampsBase` class type. I guess the original design intention was for to allow this free function to be called and used with generic stamp objects?

python/lsst/meas/algorithms/stamps.py
timj: I want it to stay. I added this with Simon because we reasoned that it was better to complain early rather than late and we really want Stamps here.

arunkannawadi: 2 against 1 here. May democracy prevail.

python/lsst/meas/algorithms/brightStarStamps.py
leeskelvin: Thanks for the in-person chat about this Arun. Following our chat, I think we've agreed that we're happy leaving this `__init__` as-is for now.

python/lsst/meas/algorithms/stamps.py
leeskelvin: I've added this in-line comment, which I hope captures the gist of the discussion above:
```
        # TO-DO: Consider refactoring this method. This class check was added
        # to allow the butler formatter to use a generic type but still end up
        # giving the correct type back, ensuring that the abstract base class
        # is not used by mistake. Perhaps this logic can be optimised.
```

python/lsst/meas/algorithms/stamps.py
leeskelvin: Following our in-person discussion, I think we're happy to leave this as-is for now.

python/lsst/meas/algorithms/brightStarStamps.py
```diff
@@ -56,56 +55,62 @@ class BrightStarStamp(AbstractStamp):
     annularFlux : `Optional[float]`
```
arunkannawadi, line 55:
```suggestion
    annularFlux : `float`, optional
```

timj: These have two different meanings. The first is that `float` or `None` are allowed. The second is that you must give a `float` but things work if you don't give anything at all (the default value will be used).

arunkannawadi: We'd want something like this then: https://github.com/lsst/ip_isr/blob/4b253454a05df26a7ddd28689a0876f006854e98/python/lsst/ip/isr/vignette.py#L87

leeskelvin: Thanks, changed here (and elsewhere that I caught it).

PR #331, opened 2023-05-22
DM-39141: Add doRequirePrimary to source selectors.

erykoff:


python/lsst/meas/algorithms/astrometrySourceSelector.py
```diff
@@ -171,6 +176,14 @@ def _isUsable(self, sourceCat):
...
             & ~sourceCat.get(self.fluxFlagKey)
 
+    def _isPrimary(self, sourceCat):
+        """Return True if this is a primary source.
+        """
+        if self.primaryKey:
+            return sourceCat.get(self.primaryKey)
+        else:
+            return np.ones(len(sourceCat), dtype=bool)
+
```
laurenam, line 186: Seeing this, I'm wondering if the same should be added to `ObjectSizeStarSelectorTask` too?

laurenam (approved):

python/lsst/meas/algorithms/astrometrySourceSelector.py
erykoff: To the best of my knowledge, that selector is never run on catalogs that have had sky sources added; it's only run at the earliest time when we're looking for stars to measure the PSF.

laurenam: Yeah, I was just thinking that, in principle, someone *could*, so it wouldn’t hurt…your call!

PR #332, opened 2023-05-23
DM-39370: Fix failure to always raise MeasureApCorrError.

erykoff:
This fixes a raise that was missed in DM-38751.


eigerx (approved):
Looks good to me. Checked out the code and did a little grep for all the RuntimeError's and none of them were about aperture correction.

PR #336, opened 2023-06-21
DM-37332: Add task to re-interpolate mask planes

enourbakhsh:


ups/meas_algorithms.table
```diff
@@ -10,6 +10,7 @@ setupRequired(pipe_base)
 setupRequired(sconsUtils)
 setupRequired(utils)
 setupRequired(cpputils)
+setupRequired(ip_isr)
```
timj, line 13: This is a bit of a red flag because ip_isr depends on meas_algorithms and so we now get a circular dependency.

enourbakhsh: Correct, and Jenkins failed for the same reason. Is there a better way to do this? My first Jenkins run without this was complaining that it could not import `lsst.ip.isr`.

timj: If your test needs ip_isr it means that the test code is probably going to have to go in ip_isr. I haven't looked in detail into what you are doing so I can't really comment further.

enourbakhsh: @arunkannawadi suggested that I copy some bits of code from `ip_isr` explicitly for now

TallJimbo: If the purpose of the test is to check consistency with `ip_isr`, the test needs to go in `ip_isr` or package that has that and `meas_algorithms` as a dependency, like `pipe_tasks`.  If the code in ip_isr just happens to be convenient for this test, just copy it.

The real test is whether it'd be a problem if you were to copy it and then the `ip_isr` behavior changed to meet ISR's needs without this copy being changed; if it's not a problem, copy away.

parejkoj: Besides the isrMock, I think several things from `lsst/ip/isr/isrFunctions.py` should be moved into afw, meas_base, or meas_algorithms. In particular `transposeMaskedImage`, `createPsf`, `growMasks`, and `makeThresholdMask` all seem like they should live in a more generally-accessible place.

That said, the fact that this depends on `ip.isr.Defects` suggests that the Task itself should live in ip_isr or pipe_tasks, not meas_algorithms.

python/lsst/meas/algorithms/reinterpolate_pixels.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+        mask = exposure.mask
+
+        if self.config.growFootprints > 0 and "SAT" in self.config.maskNameList:
+            # If we are interpolating over an area larger than the original
+            # masked region, we need to expand the original mask bit to the
+            # full area to explain why we interpolated there.
+            from lsst.ip.isr import growMasks
+
+            growMasks(mask, radius=self.config.growFootprints, maskNameList=["SAT"], maskValue="SAT")
```
TallJimbo, line 86: I think this is going to leave those mask planes grown as a side-effect, and I don't think we want that.

If this growing-as-a-side-effect behavior is also what happens in ISR, that might mean we don't need to have the capability to grow at all here, and could remove that config option and the logic that goes with it.

I haven't spotted that as being the case, but I didn't look very carefully, and I'm curious what the motivation for this grow logic was.  If we can get away without the grow here that'd be cleaner, but if there's a reason we expect to need it, we should:
- reverse it after interpolating;
- make the mask planes grown configurable.

python/lsst/meas/algorithms/reinterpolate_pixels.py
```diff
@@ -0,0 +1,140 @@
...
+
+        from lsst.ip.isr import Defects
```
TallJimbo, line 88: I hadn't realized the production code (not just the tests) depended on ip_isr; I think this is something we need to fix.

Given how thin the `Defects.fromFootprintList` method is, you should just copy its implementation here.

python/lsst/meas/algorithms/reinterpolate_pixels.py
```diff
@@ -0,0 +1,140 @@
...
+        """
+
+        psf = self._createPsf(self.config.kernelFwhm)
+        if self.config.fallbackValue is None:
+            fallbackValue = afwMath.makeStatistics(exposure.image, afwMath.MEANCLIP).getValue()
+        else:
+            fallbackValue = self.config.fallbackValue
+        if "INTRP" not in exposure.mask.getMaskPlaneDict():
+            exposure.mask.addMaskPlane("INTRP")
+        maskedImage = MaskedImageF(image=exposure.image, mask=exposure.mask, variance=exposure.variance)
```
TallJimbo, line 119: Instead of making a new `MaskedImage` here, I think you can just pass `exposure.maskedImage` to `interpolateOverDefects`, and then you won't have to assign to `exposure.image` etc. afterwards.

TallJimbo (changes requested):

python/lsst/meas/algorithms/reinterpolate_pixels.py
```diff
@@ -0,0 +1,140 @@
...
+import lsst.pex.config as pexConfig
+import lsst.pipe.base as pipeBase
+from lsst.afw.image import MaskedImageF
+
+
+class ReinterpolatePixelsConfig(pexConfig.Config):
+    """Config for ReinterpolatePixelsTask"""
+
+    kernelFwhm = pexConfig.Field[float](
+        doc="FWHM of double Gaussian smoothing kernel.",
```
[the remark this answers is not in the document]
enourbakhsh: When I passed `None` for the psf parameter in `interpolateOverDefects()`, a TypeError occurred due to incompatible function arguments. The psf parameter specifically requires the `lsst.afw.detection.Psf` type.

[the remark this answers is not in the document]
enourbakhsh: Should we consider passing a dummy psf instead? Doing so would allow us to remove the `kernelFwhm` from the configuration options.

python/lsst/meas/algorithms/reinterpolate_pixels.py
enourbakhsh: I noticed that `Defects.fromFootprintList` returns an instance of the `lsst.ip.isr.Defects` class which belongs to ip_isr. Doesn't copying the implementation of `Defects.fromFootprintList` here mean that I need to remake an entire `lsst.ip.isr.Defects` class here? I'm unsure if this aligns with our intentions.

python/lsst/meas/algorithms/reinterpolate_pixels.py
enourbakhsh: Regarding the motivation for the grow logic, I'm not entirely certain. I followed the logic in `interpolateFromMask()` from `ip_isr` [here](https://github.com/lsst/ip_isr/blob/c3c1122503caa670758387223c53d247d18482f3/python/lsst/ip/isr/isrFunctions.py#L184), where it appears that the grow logic is implemented to ensure that the interpolation covers an expanded region, including the original masked area, when dealing with *saturated pixels*. However, the ticket description suggests that this task needs to be made more reusable without relying on ISR-specific logic.

enourbakhsh: In `interpolateFromMask()`, it appears that the mask planes remain grown and are not reversed back to their original state after the interpolation process.

python/lsst/meas/algorithms/reinterpolate_pixels.py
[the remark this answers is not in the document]
TallJimbo: Yes, if we can't get away with `None`, we should pass a dummy PSF. `lsst.afw.detection.GaussianPsf` is the simplest implementation I know of.

python/lsst/meas/algorithms/reinterpolate_pixels.py
TallJimbo: Instead of making a `Defects` instance, just copy the bit that makes an iterable of ``lsst.meas.algorithms.Defect`, since that iterable is also what you need to pass to `interpolateOverDefects`.

python/lsst/meas/algorithms/reinterpolate_pixels.py
RobertLuptonTheGood: Growing is done for two reasons.   In the column direction you need to grow by one pixel as charge can spill from the last pixel over saturation into its neighbour away from the bright star.  In the row direction, we saw problems with CTE in the serial and "fixed" that by growing in the column direction.   The latter shouldn't be necessary for LSSTCam

python/lsst/meas/algorithms/reinterpolate_pixels.py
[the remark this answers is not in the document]
enourbakhsh: Passing `lsst.afw.detection.Psf()` worked!

python/lsst/meas/algorithms/reinterpolate_pixels.py
enourbakhsh: It worked, but I also needed to map the `lsst.meas.algorithms.Defect` function to each `lsst.afw.detection.footprintToBBoxList(fp)` for `fp` in `fpList` to get `List[lsst.meas.algorithms.Defect]`.

python/lsst/meas/algorithms/reinterpolate_pixels.py
enourbakhsh: Looks like we can assume the SAT mask has already been grown, so there is no need to do it again. However, @natelust suggested keeping the `growFootprints` for SAT mask as a config option (will rename it to `growSaturatedFootprints` for specificity), with a default value of 0. This way, if someone wants to expand the mask even further for some reason, they can. But in that case, I will not be leaving the SAT mask grown after the reinterplation finishes - I'll simply revert this extra mask expansion at the end of the `run()` method.

enourbakhsh: Also, I chatted with @czwa, and he pretty much agrees with Robert's response: "We intentionally grow saturated pixels because they tend to spill over to their neighboring pixels."

python/lsst/meas/algorithms/reinterpolate_pixels.py
enourbakhsh: Done!

python/lsst/meas/algorithms/reinterpolate_pixels.py
```diff
@@ -0,0 +1,140 @@
...
+        default=1,
+    )
+
+    maskNameList = pexConfig.ListField[str](
+        doc="Mask plane name.",
+        default=["SAT"],
+    )
+
+    fallbackValue = pexConfig.Field[float](
+        doc="Value of last resort for interpolation.",
```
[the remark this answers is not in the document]
enourbakhsh: Done!

python/lsst/meas/algorithms/reinterpolate_pixels.py
```diff
@@ -0,0 +1,140 @@
...
+        default=1.0,
+    )
+
+    growFootprints = pexConfig.Field[int](
+        doc="Number of pixels to grow footprints for pixels.",
+        default=1,
+    )
+
+    maskNameList = pexConfig.ListField[str](
+        doc="Mask plane name.",
```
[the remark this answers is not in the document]
enourbakhsh: Done!

parejkoj:
Something very bad happend with one of the commits here. There's a mangled-named file, `et --hard HEAD~q`, that's ~20,000 lines long

enourbakhsh:
> Something very bad happend with one of the commits here. There's a mangled-named file, `et --hard HEAD~q`, that's ~20,000 lines long

@parejkoj, thanks for spotting this, Fixed it!

ups/meas_algorithms.table
TallJimbo: @parejkoj's  last point is now moot: we've removed the dependency of the production code on `ip.isr.Defects`.

I agree that at least some of the things he's mentioned should probably be moved to afw or meas_algorithms; the fact that you've found a few ip_isr routines useful for testing this code does suggest that those in particular should be moved up (probably to meas_algorithms).

But moving code like that requires an RFC, so I think we should leave the test in pipe_tasks for now, and then:
 - open an RFC to move some of that general-purpose ip_isr code to meas_algorithms, with a deprecation period for removing aliases to them from ip_isr;
 - once that RFC is implemented, move the unit test for this ticket to meas_algorithms to keep it closer to the code it tests.

That should be a pretty simple RFC and hence good practice for an important procedure.

TallJimbo (approved):
I think everything from here on out should be straightforward, so I'm marking it as reviewed.  I'm happy to comment more, but don't feel I need to be involved vetting the next round of changes.  I think the next steps are:

- Remove the grow logic (I think that's better than disabling it, as it's unlikely to ever be enabled and unused code is not good).
- File some tickets: an RFC to move ip_isr utility functions to meas_algorithms, and another to make the growing of the SAT mask in ISR to be a separate routine rather than a side-effect of interpolatoin.

PR #337, opened 2023-06-21
DM-39712: Use introspection to read the correct MaskedImage stamp type

timj:


python/lsst/meas/algorithms/stamps.py
```diff
@@ -144,11 +150,26 @@ def readFitsWithOptions(filename, stamp_factory, options):
                 bbox = Box2I(Point2I(llcX, llcY), Extent2I(width, height))
                 kwargs["bbox"] = bbox
         stamp_parts = {}
+
+        # Determine the dtype from the factory. This allows a Stamp class
+        # to be defined in terms of MaskedImageD or MaskedImageI without
+        # forcing everything to floats.
+        masked_image_cls = None
+        for stamp_field in fields(stamp_factory.__self__):
```
ktlim, line 158: I guess you don't want to assume it's named `stamp_im`?

python/lsst/meas/algorithms/stamps.py
```diff
@@ -144,11 +150,26 @@ def readFitsWithOptions(filename, stamp_factory, options):
...
+            raise RuntimeError("Stamp factory does not use MaskedImage.")
+        default_dtype = np.dtype(masked_image_cls.dtype)
+
         # We need to be careful because nExtensions includes the primary HDU.
         for idx in range(nExtensions - 1):
+            dtype = None
             md = readMetadata(filename, hdu=idx + 1)
             if md["EXTNAME"] in ("IMAGE", "VARIANCE"):
                 reader = ImageFitsReader(filename, hdu=idx + 1)
+                dtype = default_dtype
```
ktlim, line 172: Variances are never int, I think.

timj: Thanks. Yes, MaskedImageD and MaskedImageI both use ImageF. I find it a bit confusing that a double precision image should come with a single precision variance but that's what is happening so I'll force dtype to None for VARIANCE as well as MASK.

python/lsst/meas/algorithms/stamps.py
timj: Correct. Looking at the `Stamp` code it seems I only require the factory method and that it's a dataclass but it might be confusing if someone calls it something other than `stamp_im`.

PR #338, opened 2023-06-23
DM-38546: fixes while implementing new CalibrateImageTask

parejkoj:


TallJimbo (approved):
"new mocker than" in a commit message seems like it's missing a word.  Otherwise looks good, and sorry for the delay!

TallJimbo (approved):

PR #350, opened 2023-08-26
DM-39796: Update task config defaults to LSST values

parejkoj:


python/lsst/meas/algorithms/findCosmicRaysConfig.py
```diff
@@ -33,7 +33,7 @@ class FindCosmicRaysConfig(pexConfig.Config):
     nCrPixelMax = pexConfig.Field(
         dtype=int,
         doc="maximum number of contaminated pixels",
-        default=10000,
+        default=1000000,
```
czwa, line 36: Can you explain this change?  I see that this is the obs_subaru value, and that obs_decam also used a higher value, so I can see the underlying logic in relation to this ticket, but this value must be too large.  For LSSTCam, this amounts to 1 CR pixel per 4x4 grid.  HSC and DECam have double the contaminated area.  I can't imagine any image with that level of contamination being scientifically useful.  I'll dig through the obs_subaru git history to find the reason for the high value, but I just want to raise this issue.

parejkoj: The large value was also set in obs_lsst, at the root-level `characterizeImage` override, so it's the defacto default now anyway (modulo the small DECam difference). I'd love to have some rationale for why it is what it is, and whether to make it something better. The goal of this ticket is to just make overrides that are defacto defaults into actual defaults.

czwa: I've added https://jira.lsstcorp.org/browse/DM-40586 as sorting out the correct value is out of scope here.

python/lsst/meas/algorithms/objectSizeStarSelector.py
```diff
@@ -45,73 +45,73 @@ class ObjectSizeStarSelectorConfig(BaseSourceSelectorTask.ConfigClass):
     doFluxLimit = pexConfig.Field(
         doc="Apply flux limit to Psf Candidate selection?",
         dtype=bool,
-        default=True,
+        default=False,
     )
     fluxMin = pexConfig.Field(
-        doc="specify the minimum psfFlux for good Psf Candidates",
+        doc="Minimum psfFlux for good Psf Candidates",
```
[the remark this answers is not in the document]
parejkoj: Good catch! I'm happy to cleanup these docstrings as part of this.

tests/test_negative.py
```diff
@@ -123,6 +123,7 @@ def test_significance(self):
 
         schema = afwTable.SourceTable.makeMinimalSchema()
         config = SourceDetectionTask.ConfigClass()
+        config.thresholdType = "stdev"
```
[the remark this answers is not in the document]
parejkoj: The docstring for the test itself explains why.

PR #351, opened 2023-10-03
DM-41026: Ensure that fits to refcat translates strings and booleans correctly.

erykoff:


tests/test_readFitsCatalog.py
```diff
@@ -98,6 +119,9 @@ def testHDU1DefaultNames(self):
         self.assertTrue(np.array_equal(self.arr1, table))
         self.assertEqual(len(table), 2)
 
+        schema = self._makeSchema(table.dtype)
+        self.assertIsNotNone(schema)
```
parejkoj, line 123: I'm confused about what this `IsNotNone` test is supposed to do. `_makeSchema` can't return None, can it?

tests/test_readFitsCatalog.py
```diff
@@ -128,6 +156,9 @@ def testHDU2(self):
         arr = task.run(FitsPath)
         self.assertTrue(np.array_equal(self.arr2, arr))
 
+        schema = self._makeSchema(arr.dtype)
+        self.assertIsNotNone(schema)
```
parejkoj, line 160: Same question here as above re: `None`.

python/lsst/meas/algorithms/readFitsCatalogTask.py
```diff
@@ -68,7 +68,7 @@ def run(self, filename):
             a numpy structured array containing the specified columns
         """
 
-        table = Table.read(filename, hdu=self.config.hdu)
+        table = Table.read(filename, hdu=self.config.hdu, character_as_bytes=False)
```
parejkoj, line 71: Please add a comment here as to why the kwarg?

tests/test_readFitsCatalog.py
```diff
@@ -88,8 +89,28 @@ def setUp(self):
         fitsTable = makeFitsTable()
         self.arr1 = Table(fitsTable[1].data).as_array()
         self.arr2 = Table(fitsTable[2].data).as_array()
+
         self.fitsTable = fitsTable
 
+    def _makeSchema(self, dtype):
```
parejkoj, line 95: I think this could be static (doesn't need any instance variables) and should have at least a short descriptive docstring.

parejkoj (commented):
Several questions

tests/test_readFitsCatalog.py
erykoff: I did some refactoring so that the same code is run in the code and tests.

parejkoj (approved):
Thanks this makes more sense.

PR #355, opened 2023-11-09
Update docs to refer to location on USDF

parejkoj:
A quick update to the refcat docs to refer people to the correct location for the htm7 index files on USDF. No one ever needs to generate their own for our existing refcats.


suberlak (approved):
Thanks for the update!

PR #357, opened 2023-11-22
DM-41846: Force psf candidate state before testing stringification

timj:
Other tests running could change this state and break the test.

### DIFF
--- a/tests/test_psfCandidate.py
+++ b/tests/test_psfCandidate.py
@@ -202,8 +202,14 @@ class CandidateMaskingTestCase(lsst.utils.tests.TestCase):
 
     def testStr(self):
         candidate = self.createCandidate()
+
+        # The setX methods are class globals and could contain values set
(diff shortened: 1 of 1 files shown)


tests/test_psfCandidate.py
```diff
@@ -202,7 +202,13 @@ def testFaintNeighborMasking(self):
 
     def testStr(self):
         candidate = self.createCandidate()
+
+        # The setX methods are class globals and could contain values set
+        # by other code. Force the values here to ensure test consistency.
         candidate.setChi2(2.0)
+        candidate.setWidth(0)
```
timj, line 209: It might be better if I set everything to non-default values so that we can know for sure we are seeing those values in the string form.

TallJimbo: No preference either way from me (I didn't even know these values had defaults).

TallJimbo (approved):

PR #359, opened 2023-12-07
DM-41285: improve sky sources for calibrateImage

parejkoj:


python/lsst/meas/algorithms/skyObjects.py
```diff
@@ -138,11 +138,27 @@ def generateSkyObjects(mask, seed, config):
 
 
 class SkyObjectsTask(Task):
-    """Generate a list of Footprints of sky objects.
+    """Generate a list of Footprints of sky sources (regions on the sky that
+    do not otherwise have detections).
+
+    Parameters
+    ----------
+    schema : `lsst.sfw.table.Schema`
```
leeskelvin, line 146: Typo, `sfw` to `afw`.

For parsed docstring purposes, this would read better as `~lsst.afw.table.Schema`.

python/lsst/meas/algorithms/skyObjects.py
```diff
@@ -138,11 +138,27 @@ def generateSkyObjects(mask, seed, config):
...
+        Schema used to create the output `lsst.afw.table.SourceCatalog`,
```
leeskelvin, line 147: Similarly here, with the `~` prefix.

python/lsst/meas/algorithms/skyObjects.py
```diff
@@ -138,11 +138,27 @@ def generateSkyObjects(mask, seed, config):
 
 
 class SkyObjectsTask(Task):
-    """Generate a list of Footprints of sky objects.
+    """Generate a list of Footprints of sky sources (regions on the sky that
+    do not otherwise have detections).
```
leeskelvin, line 142: "Generate a list of Footprints of sky objects/sources (regions on the sky that do not otherwise have detections)."

python/lsst/meas/algorithms/skyObjects.py
```diff
@@ -162,6 +178,9 @@ def run(self, mask, seed):
             objects.
         seed : `int`
             Random number generator seed.
+        catalog : `lsst.afw.table.SourceCatalog`, optional
+            Catalog to add detected footprints to; modified in-place if any
+            sky source footprints are created.
```
leeskelvin, line 183: "sky object/source"

python/lsst/meas/algorithms/skyObjects.py
parejkoj: This one makes sense, as it's in a sentence instead of the type field of the docstring.

python/lsst/meas/algorithms/skyObjects.py
leeskelvin: (Commenting post-merge, for reference only, so not expecting any changes to be made)

True, our dev guide doesn't say to shorten parameter types, but it also doesn't say not to either. This certainly crops up elsewhere in the stack, e.g. [daf_butler](https://github.com/lsst/daf_butler/blob/3e35ddf7b712e8d8088fef5484748109d5dd22a5/python/lsst/daf/butler/registry/interfaces/_datasets.py#L122) and also in other projects, e.g., [matplotlib](https://matplotlib.org/devdocs/devel/document.html#reference-types). I wouldn't mind having guidance on this in the dev guide, one way or another.

parejkoj: This is worth asking about on docs-support: I could see it argued either way.

PR #360, opened 2023-12-20
DM-42283: Fix tests to avoid deprecation and work with newer numpy.

erykoff:


tests/test_readFitsCatalog.py
```diff
@@ -98,7 +98,12 @@ def testHDU1DefaultNames(self):
         """
         task = ReadFitsCatalogTask()
         table = task.run(FitsPath)
-        self.assertTrue(np.array_equal(self.arr1, table))
+        for name in self.arr1.dtype.names:
```
ktlim, line 101: This doesn't pick up extra columns in `table`, does it?

PR #362, opened 2024-01-18
DM-42553: Add check for finite ra/dec in ReferenceSourceSelectorTask

erykoff:


tests/test_sourceSelector.py
```diff
@@ -329,6 +330,18 @@ def testUnresolved(self):
         self.config.unresolved.maximum = None
         self.check((starGalaxy > minimum).tolist())
 
+    def testFiniteRaDec(self):
```
mrawls, line 333: Please add the shortest of docstrings, i.e., "Test that non-finite RA and Dec values are caught."

mrawls (approved):
Looks great, agreed that this should be a default "on" thing.

PR #363, opened 2024-01-23
DM-42529: Accelerate the utility function(s) needed for CloughTocher2DInterpolator

arunkannawadi:


python/lsst/meas/algorithms/cloughTocher2DInterpolatorUtils.cc
```diff
@@ -0,0 +1,65 @@
...
+namespace algorithms {
+namespace {
+
+void declareCloughTocher2DInterpolatorUtils(lsst::cpputils::python::WrapperCollection &wrappers) {
+    using PyCloughTocher2DInterpolatorUtils =
+            py::class_<CloughTocher2DInterpolatorUtils, std::shared_ptr<CloughTocher2DInterpolatorUtils>>;
+    auto clsCloughTocher2DInterpolatorUtils = wrappers.wrapType(
+            PyCloughTocher2DInterpolatorUtils(wrappers.module, "CloughTocher2DInterpolatorUtils"),
+            [](auto &mod, auto &cls) {
+                // cls.def(py::init<>());  // can we get away without it?
```
arunkannawadi, line 47: Is it okay to not have an init method, given that we'd never want to construct an instance of this class?

include/lsst/meas/algorithms/CloughTocher2DInterpolatorUtils.h
```diff
@@ -0,0 +1,98 @@
...
+@brief This class contains static utility methods that are used by the CloughTocher2DInterpolatorTask
+and exists solely to provide a namespace.
+*/
+class CloughTocher2DInterpolatorUtils {
+    typedef float PixelT;
+
+public:
+    /**
+     * @brief Find the positions of bad pixels as defined by the masks, and also
+     find the location and pixel value of good pixels around the bad pixels.
```
TallJimbo, line 50:
```suggestion
     * find the location and pixel value of good pixels around the bad pixels.
```

python/lsst/meas/algorithms/cloughTocher2DInterpolatorUtils.cc
TallJimbo: Yes, no `__init__` is appropriate here.

TallJimbo (approved):

PR #364, opened 2024-02-08
DM-39842: Convert MeasureApCorr to new exceptions

parejkoj:


TallJimbo (approved):

python/lsst/meas/algorithms/measureApCorr.py
```diff
@@ -29,14 +29,50 @@
...
+        self.iteration = iteration
+
+    @property
+    def metadata(self):
+        metadata = {"name": self.name,
+                    "nSources": self.nSources,
+                    "ndof": self.ndof,
+                    }
+        # NOTE: have to do this because task metadata doesn't allow None.
+        if self.iteration is not None:
```
mrawls, line 73: Is the `is not None` part needed here? I feel like flake8 has yelled at me in the past for not just doing, e.g., `if thing` instead of `if thing is not None`

tests/test_measureApCorr.py
```diff
@@ -153,16 +153,18 @@ def testTooFewSources(self):
...
-                self.meas_apCorr_task.run(catalog=catalog, exposure=self.exposure)
-        self.assertIn("Unable to measure aperture correction for required algorithm", cm.output[0])
+        with self.assertRaisesRegex(measureApCorr.MeasureApCorrError,
+                                    "Unable to measure aperture correction for 'test1Ap'"):
+            self.meas_apCorr_task.run(catalog=catalog, exposure=self.exposure)
 
         # We now try again after declaring that the aperture correction is
-        # allowed to fail. This should run cleanly without raising an exception.
+        # allowed to fail. This should run without raising an exception, but
+        # will log a warning.
```
mrawls, line 162: This is a really nice example of the utility of this new exception handling option! However, it's not clear to me how a user would set this at runtime, do you have an example for how to do that available somewhere?

mrawls (approved):

python/lsst/meas/algorithms/measureApCorr.py
parejkoj: You're thinking of `if thing == None`, which is not good (`None` is a singleton). In this case, the check is explicitly "is this None", not any other "falsey" value.

tests/test_measureApCorr.py
parejkoj: This is specifically for the aperture correction algorithm: `allowFailure` is the config option that sets which measurements are allowed to fail. It's a feature that DRP uses, but I don't know anything about it. I'm just making the existing tests support the new exception system.

mrawls: Right, I know this is a specific situation where you now use `allowFailure`, and thus here lies the one place where all of this began to make sense to me. 😄

parejkoj: No, `allowFailure` has always been in MeasureApCorr, it has nothing to do with my changes. All the change here does is make the test pass, since the details of the exception have changed.

PR #365, opened 2024-02-14
DM-42663: Change RequireUnresolved to use moments-based classifier by default.

erykoff:


TallJimbo (approved):

python/lsst/meas/algorithms/sourceSelector.py
```diff
@@ -418,15 +418,15 @@ class RequireUnresolved(BaseLimit):
...
     """
-    name = pexConfig.Field(dtype=str, default="base_ClassificationExtendedness_value",
+    name = pexConfig.Field(dtype=str, default="base_ClassificationSizeExtendedness_value",
                            doc="Name of column for star/galaxy separation")
 
     def setDefaults(self):
         """Set default
 
-        ``base_ClassificationExtendedness_value < 0.5`` means unresolved.
+        ``base_ClassificationSizeExtendedness_value < 0.1`` means unresolved.
```
arunkannawadi, line 427: I think we should remove this comment. The value is configurable and not intrinsically meaningful. It didn't matter when `extendedness` was behaving like a boolean. Perhaps, change it to `,,, _value < maximum` is unresolved, or just lower values correspond to unresolved sources.

PR #368, opened 2024-03-14
DM-43083: Move maskStreaks from pipe_tasks

cmsaunders:


python/lsst/meas/algorithms/maskStreaks.py
```diff
@@ -0,0 +1,788 @@
...
+    @timeMethod
+    def run(self, maskedImage):
+        """Find and mask streaks in a masked image.
+
+        Finds streaks in the image and modifies maskedImage in place by adding a
+        mask plane with any identified streaks.
+
+        Parameters
+        ----------
+        maskedImage : `lsst.afw.image.maskedImage`
```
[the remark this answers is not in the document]
isullivan: Also note here that the mask will have `STREAK` added and modified in place.

python/lsst/meas/algorithms/maskStreaks.py
```diff
@@ -0,0 +1,788 @@
...
+        for line in lines:
+            line.sigma = self.config.invSigma**-1
+            lineModel = LineProfile(data, weights, line=line)
+            # Skip any lines that do not cover any data (sometimes happens because of chip gaps)
+            if lineModel.lineMaskSize == 0:
+                continue
+
+            fit, chi2, fitFailure = lineModel.fit(dChi2Tol=self.config.dChi2Tolerance, log=self.log)
+            if fitFailure:
+                self.log.warning("Streak fit failed.")
```
isullivan, line 752: Please elaborate on this warning message, with at least a "on line X out of nLines" or other identifier. If this is expected to happen, consider downgrading this to an `info` message.

python/lsst/meas/algorithms/maskStreaks.py
```diff
@@ -0,0 +1,788 @@
...
+
+            # Initial estimate should be quite close: fit is deemed unsuccessful if rho or theta
+            # change more than the allowed bin in rho or theta:
+            if ((abs(fit.rho - line.rho) > 2 * self.config.rhoBinSize)
+                    or (abs(fit.theta - line.theta) > 2 * self.config.thetaBinSize)):
+                fitFailure = True
+                self.log.warning("Streak fit moved too far from initial estimate. Line will be dropped.")
```
isullivan, line 759: As above, please add some identifier for which line is being dropped.

isullivan (approved):
I had a few cleanup comments, but also would like to see `setDetectionMask` moved out of `maskStreaks` to `testUtils`.

python/lsst/meas/algorithms/maskStreaks.py
```diff
@@ -0,0 +1,788 @@
...
+import scipy
+import textwrap
+import copy
+from skimage.feature import canny
+from sklearn.cluster import KMeans
+import warnings
+from dataclasses import dataclass
+
+
+def setDetectionMask(maskedImage, forceSlowBin=False, binning=None, detectedPlane="DETECTED",
```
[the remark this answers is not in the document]
cmsaunders: How about just putting it in `test_maskStreaks.py` since it is unlikely to be used elsewhere?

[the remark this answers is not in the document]
isullivan: That should work.

python/lsst/meas/algorithms/maskStreaks.py
cmsaunders: Same as above--I don't want to put any specifics here.

python/lsst/meas/algorithms/maskStreaks.py
isullivan: We are allowed to log statistics on streaks as long as there is no way to determine individual streak characteristics, so perhaps simply count the number of streak fit failures and log a single warning message such as `"Streak fitting failed on XX lines"`. My concern is about being spammed with unhelpful warning messages in crowded fields.

python/lsst/meas/algorithms/maskStreaks.py
isullivan: As above, these fit failures can be counted either separately or together with the previous failures, and just issue a warning after the loop over `lines`

isullivan (approved):
Looks good! I have an alternate suggestion for handling the logging, but it should be simple.

cmsaunders:
Thanks for the comments. I now count the fit failures and just log the number that didn't work at the end of the loop. I also changed the language to say that "streak profile could not be fit for x out of y detected lines", which I think is better, since a "Fit failure" is pretty much always because the detection does not actually look like a streak.

PR #373, opened 2024-04-12
DM-43885: Add down-sampling of stars for psf determiners when we have more than we need.

erykoff:


tests/test_psfDetermination.py
```diff
@@ -436,6 +436,44 @@ def testShowPsf(self):
...
+        """
+        self.setupDeterminer()
+
+        # Try with no downsampling.
+        inputList = list(np.arange(100))
+        candidateList = self.psfDeterminer.downsampleCandidates(inputList)
+        np.testing.assert_array_equal(candidateList, inputList)
+
+        # And with downsampling.
+        inputList = list(np.arange(500))
```
isullivan, line 450: This format is convenient for the test, but it's not the same data type as the function actually expects (`lsst.meas.algorithms.PsfCandidate`). I think this is fine, but since many people understand code by looking at how the unit tests work please add a comment stating that this is different from what would typically be supplied as a `candidateList`

python/lsst/meas/algorithms/psfDeterminer.py
```diff
@@ -63,6 +72,31 @@ class BasePsfDeterminerTask(pipeBase.Task, metaclass=abc.ABCMeta):
...
+        -------
+        outputCandidateList : `list` [`lsst.meas.algorithms.PsfCandidate`]
+            Down-selected candidate list.
+        """
+        if len(inputCandidateList) <= self.config.maxCandidates:
+            return inputCandidateList
+
+        rng = np.random.RandomState(seed=self.config.downsampleRandomSeed)
+
+        selection = rng.choice(len(inputCandidateList), size=self.config.maxCandidates, replace=False)
```
isullivan, line 93: Random selection is a reasonable choice, but might we want to select the top signal-to-noise candidates? Or try to select sources to give as close to a spatially uniform distribution as possible?

python/lsst/meas/algorithms/psfDeterminer.py
```diff
@@ -63,6 +72,31 @@ class BasePsfDeterminerTask(pipeBase.Task, metaclass=abc.ABCMeta):
...
+            Input candidate list.
+
+        Returns
+        -------
+        outputCandidateList : `list` [`lsst.meas.algorithms.PsfCandidate`]
+            Down-selected candidate list.
+        """
+        if len(inputCandidateList) <= self.config.maxCandidates:
+            return inputCandidateList
+
```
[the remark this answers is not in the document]
isullivan: You could also add the number of PSF candidates used (in either case) to the task metadata. This would allow you to test that it is used correctly in the psf and piff unit tests.

isullivan (approved):
Looks good! My main comment is to add a log message reporting the downselection, and to consider adding the number of psf stars used to the task metadata, which could then be used in the piff and psfex unit tests.

python/lsst/meas/algorithms/psfDeterminer.py
arunkannawadi: Driveby review here. I don't think we'd want to do S/N based selection, but spatially uniform sounds good.

But more generally, with the same fixed seed, are we expecting different selection each time because the `len(inputCandidateList)` will vary from one image to another?

python/lsst/meas/algorithms/psfDeterminer.py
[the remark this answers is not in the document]
erykoff: I hadn't realized this was already there ... 🤦

python/lsst/meas/algorithms/psfDeterminer.py
erykoff: Also discussed on slack.  signal-to-noise selection would be bad.  Adding spatial uniformity would be good, but at the moment I think that's a premature optimization (with the default config of 300 there's more than enough that a regular sampling should be fine).

The point of the fixed seed is that if you input the same list of 2000 candidates then you will get the same sampling of 300 candidates every time.  If you input a different list of candidates then you'll get a different sampling, but that's fine because it was different in the first place.

arunkannawadi: OK, I think it should be fine especially if the PSF candidates are ordered randomly to begin with.

PR #376, opened 2024-05-20
DM-44305: Implement Gaussian Process interpolation over bad pixels

PFLeget:
See this ticket --> https://rubinobs.atlassian.net/browse/DM-44305

### DIFF
--- a/python/lsst/meas/algorithms/gp_interpolation.py
+++ b/python/lsst/meas/algorithms/gp_interpolation.py
@@ -1,0 +1,510 @@
+# This file is part of meas_algorithms.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+import numpy as np
+from lsst.meas.algorithms import CloughTocher2DInterpolatorUtils as ctUtils
+from lsst.geom import Box2I, Point2I
+from lsst.afw.geom import SpanSet
+import copy
+import treegp
+
+import jax
+from jax import jit
+import jax.numpy as jnp
+
+import logging
+
+__all__ = [
+    "InterpolateOverDefectGaussianProcess",
+    "GaussianProcessJax",
+    "GaussianProcessTreegp",
+]
+
+
+def updateMaskFromArray(mask, bad_pixel, interpBit):
+    """
+    Update the mask array with the given bad pixels.
+
+    Parameters
+    ----------
+    mask : `lsst.afw.image.MaskedImage`
+        The mask image to update.
+    bad_pixel : `np.array`
+        An array-like object containing the coordinates of the bad pixels.
+        Each row should contain the x and y coordinates of a bad pixel.
+    interpBit : `int`
+        The bit value to set for the bad pixels in the mask.
+    """
+    x0 = mask.getX0()
+    y0 = mask.getY0()
+    for row in bad_pixel:
+        x = int(row[0] - x0)
+        y = int(row[1] - y0)
+        mask.array[y, x] |= interpBit
+    # TO DO --> might be better: mask.array[int(bad_pixel[:,1]-y0), int(bad_pixel[:,0]-x)] |= interpBit
+
+
+@jit
+def median_with_mad_clipping(data, mad_multiplier=2.0):
+    """
+    Calculate the median of the input data after applying Median Absolute Deviation (MAD) clipping.
+
+    The MAD clipping method is used to remove outliers from the data. The median of the data is calculated,
+    and then the MAD is calculated as the median absolute deviation from the median. The data is then clipped
+    by removing values that are outside the range of median +/- mad_multiplier * MAD. Finally, the median of
+    the clipped data is returned.
+
+    Parameters:
+    -----------
+    data : `np.array`
+        Input data array.
+    mad_multiplier : `float`, optional
+        Multiplier for the MAD value used for clipping. Default is 2.0.
+
+    Returns:
+    --------
+    median_clipped : `float`
+        Median value of the clipped data.
+
+    Examples:
+    ---------
+    >>> data = [1, 2, 3, 4, 5, 100]
+    >>> median_with_mad_clipping(data)
+    3.5
+    """
+    median = jnp.median(data)
+    mad = jnp.median(jnp.abs(data - median))
+    clipping_range = mad_multiplier * mad
+    clipped_data = jnp.clip(data, median - clipping_range, median + clipping_range)
+    median_clipped = jnp.median(clipped_data)
+    return median_clipped
+
+
+@jit
+def jax_rbf_kernel(x1, x2, sigma, correlation_length):
+    """
+    Computes the radial basis function (RBF) kernel matrix.
+
+    Parameters:
+    -----------
+    x1 : `np.array`
+        Location of training data point with shape (n_samples, n_features).
+    x2 : `np.array`
+        Location of training/test data point with shape (n_samples, n_features).
+    sigma : `float`
+        The scale parameter of the kernel.
+    correlation_length : `float`
+        The correlation length parameter of the kernel.
+
+    Returns:
+    --------
+    kernel : `np.array`
+        RBF kernel matrix with shape (n_samples, n_samples).
+    """
+    distance_squared = jnp.sum((x1[:, None, :] - x2[None, :, :]) ** 2, axis=-1)
+    kernel = (sigma**2) * jnp.exp(-0.5 * distance_squared / (correlation_length**2))
+    return kernel
+
+
+@jit
+def jax_get_alpha(y, kernel):
+    """
+    Compute the alpha vector for Gaussian Process interpolation.
+
+    Parameters:
+    -----------
+    y : `np.array`
+        The target values of the Gaussian Process.
+    kernel : `np.array`
+        The kernel matrix of the Gaussian Process.
+
+    Returns:
+    --------
+    alpha : `np.array`
+        The alpha vector computed using the Cholesky decomposition and solution.
+
+    """
+    factor = (jax.scipy.linalg.cholesky(kernel, overwrite_a=True, lower=False), False)
+    alpha = jax.scipy.linalg.cho_solve(factor, y, overwrite_b=False)
+    return alpha.reshape((len(alpha), 1))
+
+
+@jit
+def jax_get_gp_predict(kernel_rect, alpha):
+    """
+    Compute the predicted values of gp using the given kernel and alpha (cholesky solution).
+
+    Parameters:
+    -----------
+    kernel_rect : `np.array`
+        The kernel matrix.
+    alpha : `np.array`
+        The alpha vector from Cholesky solution.
+
+    Returns:
+    --------
+    `np.array`
+        The predicted values of y.
+
+    """
+    return jnp.dot(kernel_rect, alpha).T[0]
+
+
+class GaussianProcessJax:
+    """
+    Gaussian Process regression in JAX.
+    Kernel is assumed to be isotropic RBF kernel, and solved
+    using exact Cholesky decomposition.
+    The interpolation solution is obtained by solving the linear system:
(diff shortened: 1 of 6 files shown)


python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+        self.correlation_lenght = correlation_length
+        self.white_noise = white_noise
+        self.mean = mean
+        self._alpha = None
+
+    def fit(self, x_good, y_good):
+        y = y_good - self.mean
+        self._x = x_good
+        kernel = jax_rbf_kernel(x_good, self.std, self.correlation_lenght, self.white_noise)
+        self._alpha = jax_get_alpha(y, kernel)
```
cmsaunders, line 270: I think I would have this return `_x` and `_alpha` rather than setting them as class variables, then have `predict` take `_x` and `_alpha` as arguments. Generally we don't want to have the functions change the state of the class, or to rely on the functions being called in a certain order.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+# See the Rasmussen and Williams book for more details.
+# Each function is decorated with @jit to compile the function.
+# Exist package like tinygp, that is implemented in jax also, but
+# wanted to control the implementation here, as I am doing a lot
+# of custom things (setting my own hyperparameters, fine tune mean function,
+# dynamic binning, ...).
+
+
+@jit
+def jax_pdist_squared(x):
```
cmsaunders, line 95: I don't know what might be required by jax, but making this and `jax_cdist_squared` are unnecessary for me. If it's one line of code, or even several, and it's only used in one place, I don't think it needs to be a separate function.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+    y_err = jnp.ones(len(x[:, 0])) * y_err
+    kernel += jnp.eye(len(y_err)) * (y_err**2)
+    return kernel
+
+
+@jit
+def jax_rbf_kernel_rect(x1, x2, sigma, correlation_length):
+    """
+    Compute the radial basis function (RBF) kernel (rectangular matrix).
+    Parameters:
```
cmsaunders, line 192: Need an empty line above.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+    kernel = (sigma**2) * jnp.exp(-0.5 * distance_squared / (correlation_length**2))
+    y_err = jnp.ones(len(x[:, 0])) * y_err
+    kernel += jnp.eye(len(y_err)) * (y_err**2)
+    return kernel
+
+
+@jit
+def jax_rbf_kernel_rect(x1, x2, sigma, correlation_length):
+    """
+    Compute the radial basis function (RBF) kernel (rectangular matrix).
```
cmsaunders, line 191: I need a little more information to distinguish this from the function above.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+    Parameters:
+    -----------
+    x1 : array-like
+        The first set of input points.
+    x2 : array-like
+        The second set of input points.
```
cmsaunders, line 197: What's the difference between the first and second set of input points?

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+    correlation_length : float
+        The correlation length parameter of the kernel.
+
+    Returns:
+    --------
+    kernel_rect : ndarray
+        The computed RBF kernel (rectangular matrix).
+
+    """
+    l1 = jax_cdist_squared(x1, x2)
```
cmsaunders, line 209: A more descriptive variable name would be better here.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+    -----------
+    y : array_like
+        The target values of the Gaussian Process.
+    kernel : array_like
+        The kernel matrix of the Gaussian Process.
+
+    Returns:
+    --------
+    alpha : ndarray
+        The alpha vector computed using the Cholesky decomposition and solve.
```
cmsaunders, line 229: solve -> solution

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+
+    """
+    factor = (jax.scipy.linalg.cholesky(kernel, overwrite_a=True, lower=False), False)
+    alpha = jax.scipy.linalg.cho_solve(factor, y, overwrite_b=False)
+    return alpha.reshape((len(alpha), 1))
+
+
+@jit
+def jax_get_y_predict(kernel_rect, alpha):
```
cmsaunders, line 238: Looking at this function by itself, it's not clear what `y` is. Again though, I wouldn't make this its own function.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+        self._alpha = jax_get_alpha(y, kernel)
+
+    def predict(self, x_bad):
+        kernel_rect = jax_rbf_kernel_rect(x_bad, self._x, self.std, self.correlation_lenght)
+        y_pred = jax_get_y_predict(kernel_rect, self._alpha)
+        return y_pred + self.mean
+
+
+# Vanilla Gaussian Process regression using treegp package
+# There is no fancy O(N*log(N)) solver here, just the basic GP regression (Cholesky).
```
cmsaunders, line 279: Put this in the extended summary under the first sentence in the docstrings below. Also, I would say "the basic GP regression, which uses Cholesky decomposition."

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+        Standard deviation of the Gaussian Process kernel. Default is 1.0.
+    correlation_length : float, optional
+        Correlation length of the Gaussian Process kernel. Default is 1.0.
+    white_noise : float, optional
+        White noise level of the Gaussian Process. Default is 0.0.
+    mean : float, optional
+        Mean value of the Gaussian Process. Default is 0.0.
+
+    Methods:
+    --------
```
cmsaunders, line 296: Said this elsewhere, but you don't need to have "Methods" in the class docstring.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+
+    """
+
+    def __init__(self, std=1.0, correlation_length=1.0, white_noise=0.0, mean=0.0):
+        self.std = std
+        self.correlation_length = correlation_length
+        self.white_noise = white_noise
+        self.mean = mean
+
+    def fit(self, x_good, y_good):
```
cmsaunders, line 328: As I commented elsewhere, I would use more neutral terms, like `x_train` and `x_predict` or similar, instead of `x_good` and `x_bad`.

cmsaunders (commented):
This mostly looks good. I have a lot of comments, but they are mostly related to bringing this into conformity with the LSST stack coding standards, rather than algorithmic changes or anything like that. 

The one major thing is that there need to be unit tests for `InterpolateOverDefectGaussianProcess` and the other new classes. The standalone functions don't necessarily need their own tests as long as they are run through the unit tests for the classes.

Here are a few other general points, which I included in the comments, but didn't make a note of every time I saw them:
- For some reason the github Actions are not complaining about this, but a lot of the docstring lines are too long. The rule from the Dev Guide is 79 characters: https://developer.lsst.io/python/style.html#docstring-and-comment-line-length-must-be-less-than-or-equal-to-79-columns. For lines of code, the maximum is 110 characters.
- In the docstrings, the variable type needs to be surrounded by backquotes, i.g. `` `int` ``.
- Also in the docstrings, the summary sentence should be in the imperative sense, and should be formatted with the opening triple quotes on the same line as the sentence.
- We try to be as consistent as possible with capitalization. You have almost everything in snake_case, but a few variables are camelCase. It would be best to change those to snake_case too.
- If possible, pass the logger into the class and use that instead if `warnings`. If not, I think it would be better to use the `logging` package rather than `warnings`, just because it is what is most used in the stack.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+# y_interp = kernel_rect @ (kernel + y_err**2 * I)^-1 @ y_training.
+# See the Rasmussen and Williams book for more details.
+# Each function is decorated with @jit to compile the function.
+# Exist package like tinygp, that is implemented in jax also, but
+# wanted to control the implementation here, as I am doing a lot
+# of custom things (setting my own hyperparameters, fine tune mean function,
+# dynamic binning, ...).
+
+
+@jit
```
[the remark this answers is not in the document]
PFLeget: If I put them like this, this is because I was not able to make it work within class and there is not an obvious / fast way to make it work within class I was able to found.

python/lsst/meas/algorithms/gp_interpolation.py
PFLeget: There is not really this is just the difference between the evaluation of the kernel on the training location and the kernel between training and test location. I can re-write that as a single function. Let me do that.

python/lsst/meas/algorithms/gp_interpolation.py
PFLeget: I will run black and flake8 on it (I should have done it before submitting the PR). It should solve all those problems + the too long lines.

python/lsst/meas/algorithms/gp_interpolation.py
PFLeget: I need to make more clear what are training / test points. You have a comments latter on for this. I'll implement this accross the code.

python/lsst/meas/algorithms/gp_interpolation.py
PFLeget: I removed this and merge with the previous kernel function.

python/lsst/meas/algorithms/gp_interpolation.py
PFLeget: I did clarify, this is the GP solution it is computing.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+            # effect on the prediction.
+            bbox = bbox.dilatedBy(
+                int(self.correlation_length * self.correlation_length_cut)
+            )  # need integer as input.
+            xmin, xmax = max([glob_xmin, bbox.minX]), min(glob_xmax, bbox.maxX)
+            ymin, ymax = max([glob_ymin, bbox.minY]), min(glob_ymax, bbox.maxY)
+            localBox = Box2I(Point2I(xmin, ymin), Extent2I(xmax - xmin, ymax - ymin))
+            try:
+                sub_masked_image = self.maskedImage[localBox]
+            except IndexError:
```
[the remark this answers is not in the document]
PFLeget: To be honest I forgot why I put this in the first place. But I am sure it was happening somehow.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+                raise ValueError("Sub-masked image not found.")
+
+            sub_masked_image = self.interpolate_sub_masked_image(sub_masked_image)
+            self.maskedImage[localBox] = sub_masked_image
```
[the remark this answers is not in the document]
PFLeget: This is a really good question. I try to reproduce what was done on the other interpolation and I don't think they are touching the variance plane.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+
+        Returns:
+        --------
+        numpy.ndarray
+            The binned array of good pixels.
+        """
+
+        n_pixels = len(good_pixel[:, 0])
+        dynamic_binning = int(np.sqrt(n_pixels / self.threshold_dynamic_binning))
+        if n_pixels / self.bin_spacing**2 < n_pixels / dynamic_binning**2:
```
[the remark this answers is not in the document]
PFLeget: Make sense, but I guess after I will for got why I did this in the first place. I will let it like that if it sounds ok.

python/lsst/meas/algorithms/gp_interpolation.py
```diff
@@ -0,0 +1,657 @@
...
+                    )
+                    return sub_masked_image
+                # kernel amplitude might be better described by maximum value of good pixel given
+                # the data and not really a random gaussian field.
+                kernel_amplitude = np.max(good_pixel[:, 2:])
+            else:
+                kernel_amplitude = np.max(good_pixel[:, 2:])
+            try:
+                good_pixel = self._good_pixel_binning(copy.deepcopy(good_pixel))
+            except Exception:
```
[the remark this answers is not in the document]
PFLeget: I am honest I remember it was needed like that but I don't remember the detail. I would let it like that if this is ok.

cmsaunders (approved):

PR #379, opened 2024-06-26
DM-45010: Change RuntimeError to AlgorithmError

cmsaunders:


python/lsst/meas/algorithms/normalizedCalibrationFlux.py
```diff
@@ -31,6 +31,35 @@
...
+        Number of sources selected by the fallback source selector.
+    nCalibFluxFlag : `int`
+        Number of selected sources with raw calibration flux flag unset.
+    nRefFluxFlag : `int`
+        Number of selected sources with reference flux flag unset.
+    """
+    def __init__(self, *, initNSources, nCalibFluxFlag, nRefFluxFlag):
+        msg = "There are no valid stars to compute normalized calibration fluxes."
+        msg += (f" Of {initNSources} initially selected sources, {nCalibFluxFlag} have good raw calibration "
+                f"fluxes and {nRefFluxFlag} have good reference fluxes.")
```
parejkoj, line 49: My only concern is that this doesn't seem like it contains quite the information one needs to reconstruct the failure conditions. `n_sel==0` is the failure condition, but I don't quite know how the two "good" numbers you give in the error relate to that. Maybe that's ok, I'm just not sure.

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
```diff
@@ -253,18 +287,23 @@ def _measure_aperture_correction(self, exposure, catalog):
...
-            sel = self.fallback_source_selector.run(catalog, exposure=exposure).selected
-            sel &= (~catalog[self.config.raw_calibflux_name + "_flag"]
-                    & ~catalog[self.config.measure_ap_corr.refFluxName + "_flag"])
+            initSel = self.fallback_source_selector.run(catalog, exposure=exposure).selected
+            sel = (initSel & ~catalog[self.config.raw_calibflux_name + "_flag"]
+                   & ~catalog[self.config.measure_ap_corr.refFluxName + "_flag"])
 
             n_sel = sel.sum()
 
             if n_sel == 0:
```
parejkoj, line 300: This can be a walrus:

```suggestion
            if (n_sel := sel.sum()) == 0:
```

parejkoj (commented):
Please add a test to trigger this raising in `test_normalizedCalibrationFlux.py`. You should be able to make it fail by configuring the star selector to be excessively strict.

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
cmsaunders: The idea is to tell you which of the selection cuts is causing there to be no sources. My logic is that this is the first thing you would want to know if you encountered this error.

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
```diff
@@ -31,6 +32,35 @@
 from .sourceSelector import sourceSelectorRegistry
 
 
+class NormalizedCalibrationFluxError(lsst.pipe.base.AlgorithmError):
+    """Raised if Aperture Correction fails in a non-recoverable way.
+
+    Parameters
+    ----------
+    initNSources : `int`
```
parejkoj, line 40: Does `init` here mean `initial`? Maybe spell that out explicitly, since these names go into the metadata that is written to disk. `initialNSources` or `nInitialSources` (or `n_initial_sources`, since it seems like we're more fully trying to move to snake_case these days?).

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
```diff
@@ -31,6 +32,35 @@
...
+        Number of sources selected by the fallback source selector.
+    nCalibFluxFlag : `int`
+        Number of selected sources with raw calibration flux flag unset.
+    nRefFluxFlag : `int`
+        Number of selected sources with reference flux flag unset.
+    """
+    def __init__(self, *, initNSources, nCalibFluxFlag, nRefFluxFlag):
+        msg = "There are no valid stars to compute normalized calibration fluxes."
+        msg += (f" Of {initNSources} initially selected sources, {nCalibFluxFlag} have good raw calibration "
```
parejkoj, line 49: "raw"? Do you mean "original"? I'm not sure what "raw" means in this context.

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
```diff
@@ -31,6 +32,35 @@
...
+                f"fluxes and {nRefFluxFlag} have good reference fluxes.")
+        super().__init__(msg)
+        self.initNSources = initNSources
+        self.nCalibFluxFlag = nCalibFluxFlag
+        self.nRefFluxFlag = nRefFluxFlag
+
+    @property
+    def metadata(self):
+        metadata = {"nInitSources": self.initNSources,
```
parejkoj, line 58: I think it's worth being consistent about the variable name and the metadata.

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
```diff
@@ -99,6 +129,11 @@ class NormalizedCalibrationFluxTask(lsst.pipe.base.Task):
         Schema for the input table; will be modified in place.
     **kwargs : `dict`
         Additional kwargs to pass to lsst.pipe.base.Task.__init__()
+
+    Raises
+    ------
+    NormalizedCalibrationFluxError if there are not enough sources to
+    calculate normalization.
```
parejkoj, line 136:
```suggestion
    NormalizedCalibrationFluxError
        Raised If there are not enough sources to calculate normalization.
```

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
cmsaunders: I went with `n_initial_sources` and also changed the other variables to snake_case because I realized this file is pretty much all snake case already.

python/lsst/meas/algorithms/normalizedCalibrationFlux.py
cmsaunders: This is the language that is used elsewhere in `NormalizedCalibrationFlux`. The config that does this selection is called `raw_calibflux_name`. Do you think it would be more clear to write "good `raw_calibflux_name` fluxes" or something like that?

PR #381, opened 2024-07-08
DM-45144: Fix support for weight arrays in AccumulatorMeanStack.add_masked_image()

erykoff:


python/lsst/meas/algorithms/accumulator_mean_stack.py
```diff
@@ -106,25 +106,37 @@ def add_masked_image(self, masked_image, weight=1.0):
         good_pixels = np.where(((masked_image.mask.array & self.bit_mask_value) == 0)
                                & np.isfinite(masked_image.mask.array))
 
-        self.sum_weight[good_pixels] += weight
-        self.sum_wdata[good_pixels] += weight*masked_image.image.array[good_pixels]
+        weight_array = False
```
arunkannawadi, line 109: I know this is 100% internal, but i'd suggest calling it `is_weight_array` to denote that its meant to be boolean.

tests/test_accumulator_mean_stack.py
```diff
@@ -97,147 +97,168 @@ def make_mask_map(self, stats_ctrl):
...
-        online_masked_image = coadd_exposure.maskedImage
-
-        # The coadds match at the <1e-5 level.
-        testing.assert_array_almost_equal(online_masked_image.image.array,
-                                          afw_masked_image.image.array, decimal=5)
-        testing.assert_array_almost_equal(online_masked_image.variance.array,
-                                          afw_masked_image.variance.array)
-        testing.assert_array_equal(online_masked_image.mask.array,
-                                   afw_masked_image.mask.array)
+        for use_weight_arrays in [False, True]:
```
arunkannawadi, line 100: Here and elsewhere, I'd recommend making `use_weight_arrays` an argument to the test function and use `lsst.utils.tests.methodParameters`.

arunkannawadi (approved):
Minor improvement suggestions, and I'd like to request one unit test involving a non-trivial array as weight. I'm thinking of a toy case where the image and variance are identical (but have pixel-to-pixel variance) and passing the image as the weight produces the same stacked image as the stacked variance plane where the weight is 1.0

But more generally, what is the use case to support `weight` that is an array? Neither in adding per-detector warps to make a warp, nor in coadding warps do we want to use a weight that varies from one pixel to another.

erykoff:
cc @esheldon for the use of the weight array

arunkannawadi:
It was the 3-year ago me that made you put this comment Eli. I'm sorry, I don't think that was a good suggestion: https://github.com/lsst/meas_algorithms/pull/263#discussion_r736546163

arunkannawadi (approved):
Docstring only changes (mostly) look good to me!

PR #382, opened 2024-07-08
DM-45002: Warn instead of raising with incomplete normalization aperture correction.

erykoff:
This will make the routine always succeed in creating calibration fluxes, falling back to an identity transformation if one isn't available.  (This is similar to the regular aperture corrections that leave uncorrected values in place).

### DIFF
--- a/python/lsst/meas/algorithms/normalizedCalibrationFlux.py
+++ b/python/lsst/meas/algorithms/normalizedCalibrationFlux.py
@@ -175,16 +175,28 @@ class NormalizedCalibrationFluxTask(lsst.pipe.base.Task):
         if self.config.do_measure_ap_corr:
             ap_corr_field, ap_corr_err_field = self._measure_aperture_correction(exposure, catalog)
         else:
+            use_identity = False
             ap_corr_map = exposure.info.getApCorrMap()
             if ap_corr_map is None:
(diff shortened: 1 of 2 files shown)


python/lsst/meas/algorithms/normalizedCalibrationFlux.py
```diff
@@ -175,16 +177,30 @@ def run(self, *, exposure, catalog):
...
-                                   "do_measure_ap_corr=False but exposure has no aperture correction map.")
-            ap_corr_field = ap_corr_map.get(raw_flux_name)
-            ap_corr_err_field = ap_corr_map.get(raw_fluxerr_name)
-            if not ap_corr_field or not ap_corr_err_field:
-                raise RuntimeError("NormalizedCalibrationFlux is configured with "
-                                   f"do_measure_ap_corr=False but {raw_flux_name}/{raw_fluxerr_name} "
-                                   "not in exposure aperture correction map.")
+                self.log.warning(
+                    "Exposure does not have a valid normalization map for %r; using identity normalization.",
+                    exposure_id,
```
yalsayyad, line 185: I don't think you need to go out of your way to include the `exposure_id` in the log message because the logging system stores the dataId in every line.

yalsayyad (approved):

PR #387, opened 2024-09-06
DM-46174: Add an option to flip XY before CloughTocher2D interpolation

arunkannawadi:


tests/test_cloughTocher2DInterpolate.py
```diff
@@ -63,6 +63,14 @@ def setUp(self):
...
         self.maskedimage.image[:, 110:111] = np.nan
 
+        # Set an asymmetric region as BAD
+        self.maskedimage.mask[41:42, 63:66] = afwImage.Mask.getPlaneBitMask("BAD")
+        self.maskedimage.image[41:42, 63:66] = np.nan
+        self.maskedimage.mask[42:43, 63:65] = afwImage.Mask.getPlaneBitMask("BAD")
+        self.maskedimage.image[42:43, 63:65] = np.nan
+        self.maskedimage.mask[44, 63] = afwImage.Mask.getPlaneBitMask("BAD")
+        self.maskedimage.image[44, 63] = np.nan
+
```
arunkannawadi, line 73: I'm still working on a setup that will produce different results depending on `flipXY`.

python/lsst/meas/algorithms/cloughTocher2DInterpolator.py
```diff
@@ -129,14 +135,20 @@ def run(
...
         interpolator = CloughTocher2DInterpolator(
-            list(zip(goodpix[:, 0], goodpix[:, 1])),
+            points,
             goodpix[:, 2],
             fill_value=self.config.fillValue,
         )
 
         # Compute the interpolated values at bad pixel locations.
-        badpix[:, 2] = interpolator(badpix[:, :2])
+        col_slice = slice(1, None, -1) if self.config.flipXY else slice(2)
```
TallJimbo, line 150: I think this would be easier to read with a regular `if` block and two calls to `interpolator` with what's currently `col_slice` written out explicitly inside the square brackets.

TallJimbo (approved):

PR #388, opened 2024-09-24
DM-46486: added check for units of flux field

psferguson:
Moved version check to later and only conduct version check if fluxField has no units or incorrect units


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -715,6 +711,13 @@ class which intersect or are contained within the specified region. The
             expandedCat = expandedCat.copy(deep=True)
 
         fluxField = getRefFluxField(expandedCat.schema, filterName)
+
+        if expandedCat.schema[fluxField].asField().getUnits() != 'nJy':
```
erykoff, line 715: This should be double quotes for consistency.

erykoff (approved):

PR #389, opened 2024-09-25
DM-45899: Write a Task to compute Ex for TEx

PFLeget:
See https://rubinobs.atlassian.net/browse/DM-45899 for more informations.

### DIFF
--- a/python/lsst/meas/algorithms/computeExPsf.py
+++ b/python/lsst/meas/algorithms/computeExPsf.py
@@ -1,0 +1,120 @@
+# This file is part of meas_algorithms.
+#
+# Developed for the LSST Data Management System.
+# This product includes software developed by the LSST Project
+# (https://www.lsst.org).
+# See the COPYRIGHT file at the top-level directory of this distribution
+# for details of code ownership.
+#
+# This program is free software: you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+
+from lsst.meas.algorithms.treecorrUtils import TreecorrConfig
+from lsst.pex.config import Config, ConfigField
+from lsst.pipe.base import Task
+import lsst.pipe.base as pipeBase
+import treecorr
+import copy
+import numpy.typing as npt
+
+
+__all__ = ("ComputeExPsfTask", "ComputeExPsfConfig")
+
+
+class ComputeExPsfConfig(Config):
+    """Config class of ComputeExPsfTask."""
+
+    treecorr = ConfigField(
+        dtype=TreecorrConfig,
+        doc="treecorr config.",
+    )
+
+
+class ComputeExPsfTask(Task):
+    """Compute Ex for PSF.
+
+    Compute scalar correlation function from
+    PSF ellipticity residuals to compute TEx
+    metrics.
+
+    Parameters
+    ----------
+    de1: `np.ndarray`
+        PSF ellipticity residuals component 1.
+    de2: `np.ndarray`
+        PSF ellipticity residuals component 2.
+    ra: `np.ndarray`
+        Right ascension coordinate.
+    dec: `np.ndarray`
+        Declination coordinate.
+    units: `str`
+        In which units are ra and dec. units supported
+        are the same as the one in treecorr.
+
+    Returns
+    -------
+    struct : `lsst.pipe.base.Struct`
+            The struct contains the following data:
+        ``E1``: `float`
+            <de1 de1> scalar correlation function, compute
+            in an angular bin define in TreecorrConfig.
+        ``E2``: `float`
+            <de2 de2> scalar correlation function, compute
+            in an angular bin define in TreecorrConfig.
+        ``Ex``: `float`
+            <de1 de2> scalar cross-correlation function, compute
+            in an angular bin define in TreecorrConfig.
+    """
+
+    ConfigClass = ComputeExPsfConfig
+    _DefaultName = "computeExPsf"
+
+    def run(
+        self,
+        de1: npt.NDArray,
+        de2: npt.NDArray,
+        ra: npt.NDArray,
+        dec: npt.NDArray,
+        units: str = "arcmin",
+    ) -> pipeBase.Struct:
+
+        if units != self.config.treecorr.sep_units:
+            raise ValueError(
+                "units from ComputeExPsfConfig and"
+                "ComputeExPsfTask are not the same (%s != %s)"
+                % ((units, self.config.treecorr.sep_units))
+            )
+
+        kwargs_cat = {
+            "ra": ra,
+            "dec": dec,
+            "ra_units": units,
+            "dec_units": units,
+        }
+
+        cat1 = treecorr.Catalog(k=de1, **kwargs_cat)
+        cat2 = treecorr.Catalog(k=de2, **kwargs_cat)
+
+        config_kk = self.config.treecorr.toDict()
+
+        kk = treecorr.KKCorrelation(config_kk)
+
+        kk.process(cat1)
+        kk_E1 = copy.deepcopy(kk.xi[0])
+        kk.process(cat2)
+        kk_E2 = copy.deepcopy(kk.xi[0])
+        kk.process(cat1, cat2)
+        kk_Ex = copy.deepcopy(kk.xi[0])
+
+        return pipeBase.Struct(metric_E1=kk_E1, metric_E2=kk_E2, metric_Ex=kk_Ex)
--- a/python/lsst/meas/algorithms/computeExPsf.py
+++ b/python/lsst/meas/algorithms/computeExPsf.py
@@ -40,6 +40,15 @@ class ComputeExPsfConfig(Config):
         doc="treecorr config.",
     )
 
+    def setDefaults(self):
+        super().setDefaults()
+
+        self.treecorr.min_sep = 1.0 / 60.0
+        self.treecorr.max_sep = 5.0 / 60.0
+        self.treecorr.nbins = 1
+        self.treecorr.bin_type = "Linear"
+        self.treecorr.sep_units = "degree"
+
 
 class ComputeExPsfTask(Task):
     """Compute Ex for PSF.
--- a/python/lsst/meas/algorithms/treecorrUtils.py
+++ b/python/lsst/meas/algorithms/treecorrUtils.py
@@ -29,7 +29,7 @@ class TreecorrConfig(Config):
     The fields in this class correspond to the parameters that can be passed to
     any calls to `treecorr` methods, including catalog creation and two-point
     correlation function calculations. The default values set for the fields
-    are identical to the default values set in v4.2 of `treecorr`.
+    are identical to the default values set in v4.3 of `treecorr`.
 
     A separate config class is used instead
     of constructing a `~lsst.pex.config.DictField` so that mixed types can be
@@ -38,8 +38,9 @@ class TreecorrConfig(Config):
 
     Notes
     -----
-    This is intended to be used in CalcRhoStatistics class. It only supports
-    some of the fields that are relevant for rho-statistics calculations.
+    This is intended to be used with correlations of PSF residuals. It only supports
+    some of the fields that are relevant for rho-statistics calculations and the likes
+    of it.
     """
 
     nbins = Field[int](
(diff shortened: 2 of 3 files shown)


python/lsst/meas/algorithms/__init__.py
```diff
@@ -62,6 +62,7 @@
 from .scaleVariance import *
 from .noise_covariance import *
 from .gp_interpolation import *
+from .computeExPsf import *
```
arunkannawadi, line 65: Let's avoid this. Older modules are imported by default to not break things, but newer modules are expected to be specifically imported if required.

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,305 @@
...
+import treecorr
+import copy
+import numpy.typing as npt
+
+
+__all__ = ("ComputeExPsfTask", "TreecorrConfig")
+
+
+class TreecorrConfig(Config):
+    """A Config class that holds some of the parameters supported by treecorr.
```
arunkannawadi, line 34: We should put this as its own file. We anticipate reusing this in `analysis_tools` and it is not evident which module to find this in if we put it here.

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,305 @@
...
+
+        cat1 = treecorr.Catalog(k=de1, **kwargs_cat)
+        cat2 = treecorr.Catalog(k=de2, **kwargs_cat)
+
+        config_kk = self.config.toDict()
+
+        kk = treecorr.KKCorrelation(config_kk)
+
+        kk.process(cat1)
+        kk_E1 = copy.deepcopy(kk.xi[0])
```
arunkannawadi, line 299: Do we need to do a deepcopy here? Does `kk_E1` actually get overridden when we process the next catalog?

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,305 @@
...
+        kk.process(cat2)
+        kk_E2 = copy.deepcopy(kk.xi[0])
+        kk.process(cat1, cat2)
+        kk_Ex = copy.deepcopy(kk.xi[0])
+
+        return pipeBase.Struct(E1=kk_E1, E2=kk_E2, Ex=kk_Ex)
```
arunkannawadi, line 305: There is some awkwardness in starting a variable with upper case (as opposed to a class). I don't have a better suggestion at the moment.

tests/test_computeExPsf.py
```diff
@@ -0,0 +1,272 @@
...
+
+        # At large scale, expect the scalar two-point correlation function to
+        # be all close to 0.
+
+        task = ComputeExPsfTask(config)
+        output2 = task.run(self.de1, self.de2, ra, dec, units="arcmin")
+
+        np.testing.assert_allclose(output2.E1, 0.0, atol=1e-1)
+        np.testing.assert_allclose(output2.E2, 0.0, atol=1e-1)
+        np.testing.assert_allclose(output2.Ex, 0.0, atol=1e-1)
```
arunkannawadi, line 259: Could we include some intermediate scale as well? It'd be at least useful to check that the values from `E2` is about twice more than `E1` (or whatever factor we naively expect).

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,305 @@
...
+
+    ConfigClass = TreecorrConfig  # ComputeExPsfConfig
+    _DefaultName = "computeExPsf"
+
+    def run(
+        self,
+        de1: npt.NDArray,
+        de2: npt.NDArray,
+        ra: npt.NDArray,
+        dec: npt.NDArray,
```
arunkannawadi, line 280: I suspect we'd want to just take a catalog and extract these columns here. But that can wait until we start using this.

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,305 @@
...
+__all__ = ("ComputeExPsfTask", "TreecorrConfig")
+
+
+class TreecorrConfig(Config):
+    """A Config class that holds some of the parameters supported by treecorr.
+
+    The fields in this class correspond to the parameters that can be passed to
+    any calls to `treecorr` methods, including catalog creation and two-point
+    correlation function calculations. The default values set for the fields
+    are identical to the default values set in v4.2 of `treecorr`.
```
arunkannawadi, line 39: It'd be good to check if the defaults are consistent with the  version of `treecorr` available in the `rubin-env` / shared stack at USDF.

arunkannawadi (commented):
There's actually a guide on how to transfer code between packages: https://developer.lsst.io/stack/transferring-code.html#transferring-code-between-packages

Although the history is minimal, it'd be a good practice to follow this procedure.

python/lsst/meas/algorithms/computeExPsf.py
PFLeget: Yes and yes. I was surprised by that but this is what happened while I was doing unit tests. If you have another suggestions, please let me know.

python/lsst/meas/algorithms/computeExPsf.py
PFLeget: I guess this is just because in all tickets TEx is upper case and I guess putting lower case can introduce confusion with ellipticity  parameters. Open to suggestions and don't have a big feeling on this.

python/lsst/meas/algorithms/computeExPsf.py
PFLeget: I did and I made the corrections if needed.

python/lsst/meas/algorithms/treecorrUtils.py
```diff
@@ -0,0 +1,205 @@
+from lsst.pex.config import ChoiceField, Config, Field, FieldValidationError
```
arunkannawadi, line 1: The standard preamble is missing.

python/lsst/meas/algorithms/treecorrUtils.py
```diff
@@ -0,0 +1,205 @@
...
+    are identical to the default values set in v4.3 of `treecorr`.
+
+    A separate config class is used instead
+    of constructing a `~lsst.pex.config.DictField` so that mixed types can be
+    supported and the config can be validated at the beginning of the
+    execution.
+
+    Notes
+    -----
+    This is intended to be used in CalcRhoStatistics class. It only supports
```
arunkannawadi, line 19: It's time to generalize this and replace 'CalcRhoStatistics class' with correlations of PSF residuals or something similar.

python/lsst/meas/algorithms/treecorrUtils.py
```diff
@@ -0,0 +1,205 @@
...
+    some of the fields that are relevant for rho-statistics calculations.
```
arunkannawadi, line 20:
```suggestion
    some of the fields that are relevant for rho-statistics calculations and the likes of it.
```

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,103 @@
...
+
+from lsst.meas.algorithms.treecorrUtils import TreecorrConfig
+from lsst.pipe.base import Task
+import lsst.pipe.base as pipeBase
+import treecorr
+import copy
+import numpy.typing as npt
+
+
+__all__ = "ComputeExPsfTask"
```
arunkannawadi, line 31: THis should be a list or tuple.

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,103 @@
...
+            in an angular bin define in TreecorrConfig.
+        ``E2``: `float`
+            <de2 de2> scalar correlation function, compute
+            in an angular bin define in TreecorrConfig.
+        ``Ex``: `float`
+            <de1 de2> scalar cross-correlation function, compute
+            in an angular bin define in TreecorrConfig.
+    """
+
+    ConfigClass = TreecorrConfig
```
arunkannawadi, line 70: This `Task` needs to have a corresponding `ConfligClass`, that _has_ a `ConfigurableField` of type `TreecorrConfig`.

python/lsst/meas/algorithms/computeExPsf.py
arunkannawadi: One option would be to call them `metric_E1` instead of `E1`.

arunkannawadi (commented):
The transfer branch should move the code as it was. Suggested doc changes and updating the default values to correspond to treecorr `v4.3` should happen after the transfer branch is merged to the ticket branch.

python/lsst/meas/algorithms/computeExPsf.py
PFLeget: I guess I did what you described, but not 100% sure. If you can have a look at what I did on that, I won't mind.

arunkannawadi: You'll need to follow the pattern similar to the `Config` for `CalcRhoStatistics` here: https://github.com/lsst/analysis_tools/blob/dda7fc77d96733815aa073d4c1dc6caacc4ec067/python/lsst/analysis/tools/actions/vector/calcRhoStatistics.py#L161

arunkannawadi: Ah, I missed the new changed. You might have it all now.

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,113 @@
...
+import numpy.typing as npt
+
+
+__all__ = ("ComputeExPsfTask", "ComputeExPsfConfig")
+
+
+class ComputeExPsfConfig(Config):
+    """Config class of ComputeExPsfTask."""
+
+    treecorr_config = ConfigField(
```
arunkannawadi, line 38: I understand it can be confusing at first, but make this just `treecorr`.

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,113 @@
...
+        ra: npt.NDArray,
+        dec: npt.NDArray,
+        units: str = "arcmin",
+    ) -> pipeBase.Struct:
+
+        kwargs_cat = {
+            "ra": ra,
+            "dec": dec,
+            "ra_units": units,
+            "dec_units": units,
```
arunkannawadi, line 96: We should check that `units` here match up with the `units` in `TreecorrConfig`.

arunkannawadi (commented):
Are the default values in `TreecorrConfig` suitable for the metric calculation? I don't quite remember what binning values we settled on, and what the default ones would be.  `ComputeExPsfConfig` should probably have a `setDefaults` method that would set it to the `nbins`, `min_sep`, `max_sep` and `units` values appropriately.

python/lsst/meas/algorithms/computeExPsf.py
```diff
@@ -0,0 +1,120 @@
...
+
+
+class ComputeExPsfConfig(Config):
+    """Config class of ComputeExPsfTask."""
+
+    treecorr = ConfigField(
+        dtype=TreecorrConfig,
+        doc="treecorr config.",
+    )
+
```
arunkannawadi, line 42: I suspect this needs an explicit `validate` method that calls `self.treecorr.validate`. Because `TreecorrConfig` is not a `ConfigClass` of a sub task, I am fairly certain that it will never be validated.

arunkannawadi: There also needs to be a `setDefaults` method here that sets the `min_sep`, `max_sep` and either `nbins` or `bin_size` as appropriate here corresponding to the requirements. @jmeyers314 - what was the binning we settled on finally for TEx computations?

PFLeget: For the `validate`, I tried in the unit test and putting wrong value in `ComputeExPsfConfig` (`min_sep>max_sep`) for example and it triggered the `validate` from `TreecorrConfig`. Do I need to do more than that?

arunkannawadi: I think it'd be best to put it here, given that i)  we might want to run this dynamically outside of `computeSummaryStats` and ii) it makes sense for this class to have its own defaults for `treecorr` rather than `computeSummaryStats` setting them.

arunkannawadi (approved):
Same comment as in `analysis_tools` PR - start the commit messages with upper case. Good to go after you set the defaults here in `setDefaults`.

python/lsst/meas/algorithms/treecorrUtils.py
```diff
@@ -0,0 +1,228 @@
...
+
+    num_bootstrap = Field[int](
+        doc=(
+            "How many bootstrap samples to use for the 'bootstrap' and 'marked_bootstrap' var methods."
+        ),
+        default=500,
+        optional=True,
+    )
+
+    rng = Field[int](
```
arunkannawadi, line 206: Post merge comment: This should remain as `rng_seed`. `rng` typically refers to a `np.RandomState` instance, and that's certainly what `treecorr` means. `rng` itself is not nicely persistable and settable as a config file. `rng_seed` makes it explicit that it is the integer value used to seed that instance.

python/lsst/meas/algorithms/treecorrUtils.py
```diff
@@ -0,0 +1,228 @@
...
+        doc="Value to seed the treecorr random number generator with. Used to generate patches.",
+        default=None,
+        optional=True,
```
arunkannawadi, line 209: Post merge comment: Restore the previous default value and remove `optional=True`.

PR #398, opened 2024-11-08
DM-47385: Refine mask exclusion behavior in removeBadPixels (aka cancel original masks)

mrawls:
In principle, convolution-induced mask growth may block some good sources near zeroed-out regions, so we switched to the original masked regions. Unfortunately in practice, bogus diaSources appeared near masked regions, so we are undoing this change and  living with losing some edge-y diaSources.

### DIFF
--- a/python/lsst/meas/algorithms/detection.py
+++ b/python/lsst/meas/algorithms/detection.py
@@ -774,7 +774,7 @@ class SourceDetectionTask(pipeBase.Task):
             convolveResults = self.convolveImage(maskedImage, psf, doSmooth=doSmooth)
             middle = convolveResults.middle
             sigma = convolveResults.sigma
-            self.removeBadPixels(middle, exposure.maskedImage.mask)
+            self.removeBadPixels(middle)
 
             results = self.applyThreshold(middle, maskedImage.getBBox())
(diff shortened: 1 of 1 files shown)


python/lsst/meas/algorithms/detection.py
```diff
@@ -799,20 +799,17 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
...
+        middle : `lsst.afw.image.Exposure`
+            Score or maximum likelihood difference image.
             The image plane will be modified in place.
-        mask : `lsst.afw.image.Mask`
-            Mask to select planes from to zero the image; will be restricted
-            to the image's bounding box.
         """
-        badPixelMask = mask.getPlaneBitMask(self.config.excludeMaskPlanes)
-        badPixels = mask.subset(middle.getBBox()).array & badPixelMask > 0
+        badPixelMask = lsst.afw.image.Mask.getPlaneBitMask(self.config.excludeMaskPlanes)
```
isullivan, line 811: Don't set this back to what it was, since the global definition is not necessarily correct. Use `middle.mask.getPlaneBitMask(..`

isullivan (approved):
One small but important change requested.

PR #399, opened 2024-11-12
DM-46388: Remove unnecessary STREAK

parejkoj:


python/lsst/meas/algorithms/maskStreaks.py
```diff
@@ -600,7 +600,6 @@ def run(self, maskedImage):
         """
         streaks = self.find(maskedImage)
 
-        maskedImage.mask.addMaskPlane(self.config.streaksMaskPlane)
```
mrawls, line 603: Please wrap this in an `if` statement, i.e., if the configured `streaksMaskPlane` name is not STREAK, then create it, else use it because it already exists. I could envision a world where DRP wants to use this to make a special different one-off throwaway mask.

mrawls (changes requested):

python/lsst/meas/algorithms/maskStreaks.py
parejkoj: I've done that, but I'm skeptical of keeping around extra "have a config in case someone might possibly want it in the future" fields; that's a recipe for trouble in the future, if it never gets used but confuses someone looking through the configs. I'd rather the "mask streaks" task just always sets `STREAK`, since that's what the latter was made for. But I won't do that on this ticket.

mrawls (approved):

PR #402, opened 2024-11-27
DM-47787: Add check for NaN PSF sigma

isullivan:


brianyanny (approved):
This will raise an error if the psfSigma evaluates to a NaN.  This seems fine for now, but wonder if this will prevent further processing on this image? What will be affected downstream -- or are there enough 'other PSFs' usually available that this one failing is ok overall.

python/lsst/meas/algorithms/installGaussianPsf.py
```diff
@@ -76,6 +76,9 @@ def run(self, exposure):
         if exposure.hasPsf():
             psfModel = exposure.getPsf()
             psfSigma = psfModel.computeShape(psfModel.getAveragePosition()).getDeterminantRadius()
+            if math.isnan(psfSigma):
+                raise RuntimeError("Cannot create a Gaussian PSF model. "
+                                   "The computed exposure psfSigma is NaN!")
```
parejkoj, line 81: This is ok, but what we really need is a new `AlgorithmError` in `psfexDeterminerTask` to raise if the fitted psf model is bad. Erfan added `PsfexTooFewGoodStarsError`, so you can crib from that.

PR #405, opened 2024-12-24
DM-47738: Update for refactored meas_extensions_scarlet

fred3m:


python/lsst/meas/algorithms/setPrimaryFlags.py
```diff
@@ -295,15 +295,20 @@ def getDeblendPrimaryFlags(sources):
...
-        isIsolated = isLeaf & ((parent == 0) | parentNChild == 1)
-        isDeblendedSource = (fromBlend & isLeaf) | (isIsolated & (parent == 0))
-        isDeblendedModelSource = (parent != 0) & isLeaf
+        isLeaf = nChild == 0
+        isChild = depth > 0
+        isSibling = parentNPeaks > 1
+        isDeblendedModelSource = isLeaf & isChild
+        fromBlend = isDeblendedModelSource & isSibling
+        isIsolatedParent = (depth == 0) & (nPeaks == 1)
+        isIsolated = isIsolatedParent | ((depth == 1) & ~isSibling)
```
taranu, line 310: Shouldn't this generally be `depth >= 1`?

fred3m: No, anything with depth > 1 came from a blend, so by definition it can't be a model of an isolated source. It would be a source that is isolated in the deconvolved frame but not the observed seeing.

PR #408, opened 2025-01-15
DM-47730: Add method to get refcat schema from loader.

TallJimbo:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -765,6 +765,47 @@ def loadSkyCircle(self, ctrCoord, radius, filterName, epoch=None):
...
+            ``schema``
+                Schema of the reference catalogs returned by other 'load'
+                methods.
+            ``fluxField``
+                Name of the field containing the flux associated with
+                ``filterName``.
+        """
+        if not self.refCats:
+            raise RuntimeError("No reference tables could be found.")
+        cat = self.refCats[0].get()
```
jrmullaney, line 790: This probably flags my naivety more than anything else, but unless you feel it's obvious to everyone else, I wonder whether it'd be useful to comment on why it's ok to only use the first refcat in the list.

jrmullaney (approved):
Just a tiny query/comment about using the first refcat in the list.

I'm happy to approve.

python/lsst/meas/algorithms/loadReferenceObjects.py
TallJimbo: It's because all of the refcat shards for the same catalog should have the same schema.  I'll add a comment.

PR #409, opened 2025-01-28
DM-45545: Fixes for numpy 2

timj:
@PFLeget there is a problem with numpy 2 and test_computeExPsf.py. The problem seems to be that `multivariate_normal` is returning very different answers between numpy 1 and 2. The inputs seem to be the same. Also, using the modern RNG APIs leads to breakage even on numpy 1.

### DIFF
--- a/python/lsst/meas/algorithms/testUtils.py
+++ b/python/lsst/meas/algorithms/testUtils.py
@@ -85,9 +85,10 @@ def plantSources(bbox, kwid, sky, coordList, addPoissonNoise=True):
 
     # add Poisson noise
     if (addPoissonNoise):
-        np.random.seed(seed=1)  # make results reproducible
+        # Make results predictable over different numpy versions.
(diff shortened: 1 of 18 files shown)


timj:
@erykoff can you please review the python changes? I won't merge until we can sort out the warpedPsf issue.

erykoff (commented):
Looks good, modulo the warped psf test problem.  However, I also noticed a couple numpy deprecation warnings.  Are these in scope?

timj:
I don't mind fixing other problems. I see there are still 3 calls to np.random that need fixing in tests that don't actually set a seed at all. Running the tests I don't see a warning other than the one from kht.

I don't think we can fix the np.trapz warnings that are in other packages until we drop numpy 1.x.

timj:
I removed the final three calls to np.random.x and squashed the commits that were doing those changes.

PR #412, opened 2025-02-18
DM-48783: Add illumination correction support to SubtractBackgroundTask and SourceDetectionTask

erykoff:
This PR also adds some unit tests for SubtractBackgroundTask and the background re-estimation in SourceDetectionTask.

### DIFF
--- a/python/lsst/meas/algorithms/dynamicDetection.py
+++ b/python/lsst/meas/algorithms/dynamicDetection.py
@@ -272,7 +272,7 @@ class DynamicDetectionTask(SourceDetectionTask):
         return Struct(multiplicative=medianError/stdevMeas, additive=bgMedian)
 
     def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True, expId=None,
-                         background=None):
+                         background=None, backgroundToPhotometricRatio=None):
         """Detect footprints with a dynamic threshold
 
         This varies from the vanilla ``detectFootprints`` method because we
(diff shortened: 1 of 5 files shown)


python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -291,6 +293,8 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
                 Results from preliminary detection pass.
                 (`lsst.pipe.base.Struct`)
         """
+        if backgroundToPhotometricRatio is not None:
+            raise RuntimeError("DynamicDetectionTask does not support backgroundToPhotometricRatio.")
```
TallJimbo, line 297: It might be worth it to add a config validate check for this, too.

python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -36,6 +37,51 @@
...
+    maskedImage : `lsst.afw.image.MaskedImage`
+        Masked image (image + mask + variance) to convert from a
+        photometrically flat image to an image suitable for background
+        subtraction.
+    doApply : `bool`
+        Apply the conversion? If False, this context manager will not
+        do anything.
+    backgroundToPhotometricRatio : `lsst.afw.image.Image`, optional
+        Image to convert background-flattened image to
+        photometric-flattened image (and back).
```
TallJimbo, line 56: When you fixed the implementation on the later commit, I don't think the docs had to change, because I don't feel like "apply" implies either "multiply" or "divide".  But I also think we want the docs to talk about going from a photometric-flattened image to a backtround-flattened one and back, since that's what the context actually does.  How about:

> Image to multiply a photometrically-flattened image by to obtain a background-flattened image.

TallJimbo (approved):

PR #416, opened 2025-03-19
DM-49274: Add flag_nodata to selector bad flag lists

leeskelvin:


arunkannawadi (approved):
This should have ideally been part of DM-45621 but I notice that `matcherSourceSelector` is not flagging based on `flag_nodata`.

PR #418, opened 2025-03-27
DM-49719: Add a starId to psfCandidate.

PFLeget:


arunkannawadi (commented):
I am a little apprehensive about changing anything here. Given that `.getId` either on a `psfCandidate` or `.getSource().getId()` is available, these changes are unnecessary. Moreover, there's some global nature to these candidates - for e.g., I've observed behaviors where changing the width of one candidate changes the width of all candidates in `psfCandidateList`.  Given that, I'd be very wary of setting a `starId` and making sure that it doesn't set it for everything.

PFLeget:
Oh didn't know about `.getSource().getId()`... I tested it works just fine :) Closing this PR then.

PR #421, opened 2025-04-16
DM-45987: Convert RuntimeErrors in ObjectSizeStarSelector to AlgorithmError

parejkoj:


cmsaunders (approved):
Looks good. One small comment inline.

PR #428, opened 2025-05-10
DM-50717: Add option to filter outliers from the binned background image

isullivan:


python/lsst/meas/algorithms/subtractBackground.py
```diff
@@ -274,6 +283,20 @@ def _addStats(self, exposure, background, statsKeys=None):
...
+
+        Parameters
+        ----------
+        bbox : `lsst.geom.Box2I`
+            Bounding box of the original image.
+        background : `lsst.afw.math.BackgroundMI`
+            Fit and binned background image, which will be modified in place.
+        """
+        statsImg = background.getStatsImage()
+        statsImg.image.array = median_filter(statsImg.image.array, mode='reflect', size=3)
```
leeskelvin, line 297: I'd certainly like to see the `size=3` as a configurable here. I'm less concerned of the need to make `mode` configurable though (I think `reflect` is a good one-size-fits-all here).

leeskelvin (approved):

PR #431, opened 2025-05-27
DM-51066: Persist metadeta to Stamp instances

fred3m:


timj (approved):
Thanks. Looks good.

I am pondering whether we should have a field default factory for metadata so that it will always have a PropertyList value (so no need for users to check for `None`) but it's probably not important since we are always setting it in the factory.

PR #440, opened 2025-08-09
DM-52141: Catch zero length glint trail edge case

mrawls:


python/lsst/meas/algorithms/findGlintTrails.py
```diff
@@ -268,7 +268,7 @@ def extract(fitter, x, y, prefix=""):
                            prefix, fitter.score(x, y), stderr, sum(fitter.inlier_mask_), len(x))
             # Simple O(N^2) search for longest distance; there will never be
             # enough points in a trail a for "faster" approach to be worth it.
-            length = max(scipy.spatial.distance.pdist(np.hstack((x, y))))
+            length = max(scipy.spatial.distance.pdist(np.hstack((x, y))), default=0)
```
isullivan, line 271: This works, but should you also catch these cases with 0 length to reject as a glint? Or, do these get removed as outliers later anyway?

isullivan (approved):
One comment for you to think about, but it is OK to merge as-is.

python/lsst/meas/algorithms/findGlintTrails.py
mrawls: I am honestly not sure how this could ever trigger with a 0 length scenario, since the offending line is only ever called if the minimum number of points threshold (default 5) is exceeded. Somehow there must be a glint where none of the constituent points are classified as inliers? So 1 start, 1 end, and at least 3... non-inliers... from the RANSAC algorithm that is designed to find inliers. If we start noticing lots of zero-length glints, there may be another deeper bug here we need to find. Going to merge as-is so we at least aren't falling over in this weird situation.

PR #460, opened 2026-01-21
DM-53647: detectCoaddPeaks : Too high a fraction (1.0 > 0.9) of pixels were masked with current "bright" detection round thresholds.

laurenam:


python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -655,7 +662,7 @@ def tweakBackground(self, exposure, bgLevel, bgList=None):
             bgList.append(bgData)
         return bg
 
-    def _computeBrightDetectionMask(self, maskedImage, convolveResults):
+    def _computeBrightDetectionMask(self, maskedImage, convolveResults, maxBrightDetectionIter=15):
```
kherner: Is it better to leave the default undefined here if it's getting pulled from the config? I'm just wondering if one can avoid having to set it in two places (here and line 127) if it ever changes in the future.

kherner (approved):

python/lsst/meas/algorithms/dynamicDetection.py
laurenam: Yeah, good point!  Since it's essentially a private function anyway, I switched it so just use the config (in line with the other uses in this function), and I renamed the config to more closely follow the established conventions here (i.e. moved `Max` to the end of the name).  I also reduced the default to 10 (which is still likely overkill, but this _should_ be a total edge case situation, so shouldn't cost us much).

kherner (approved):

PR #466, opened 2026-02-16
DM-54159: Update use of butler-stored detector pointing and regions in single frame processing

laurenam:


python/lsst/meas/algorithms/loadReferenceObjects.py
```diff
@@ -651,6 +651,9 @@ class which intersect or are contained within the specified region. The
         epoch : `astropy.time.Time` or `None`, optional
             Epoch to which to correct proper motion and parallax, or `None` to
             not apply such corrections.
+        wcsForCentroids : `lsst.afw.geom.SkyWcs`, optional
+            If provided, thsi WCS will be used to convert the reference catalog
```
TallJimbo: typo `thsi`

python/lsst/meas/algorithms/loadReferenceObjects.py
TallJimbo, line 1: Commit message is missing a word near "sky positions the reference catalog".

TallJimbo (approved):

PR #467, opened 2026-03-17
DM-54397: Add task for computing a rough shapelet expansion of isolated stars.

TallJimbo:


include/lsst/meas/algorithms/SpanSetMoments.h
```diff
@@ -0,0 +1,131 @@
...
+namespace lsst {
+namespace meas {
+namespace algorithms {
+
+/**
+ *  A struct that computes the unweighted moments of the pixels in an
+ * `lsst.afw.geom.SpanSet`.
+ *
+ * This class provides low-level pixel-processing code for
+ * `ComputeRoughtPsfShapeletsTask`.
```
laurenam: Drop the “t” in `Rought`

include/lsst/meas/algorithms/SpanSetMoments.h
```diff
@@ -0,0 +1,131 @@
...
+    /// Center derived from the unweighted first moment of the image.
+    geom::Point2D center;
+
+    /// Shape derived from the unweighted second moment of the image.
+    afw::geom::ellipses::Quadrupole shape;
+
+    /// The pixels actually used to compute the moments.
+    std::shared_ptr<afw::geom::SpanSet> spans;
+
+    /// Flag if there were too many bad pixels to compute the moments.
```
laurenam: “Flag if” —> “Flag set if” to match those below?

include/lsst/meas/algorithms/SpanSetMoments.h
```diff
@@ -0,0 +1,131 @@
...
+    bool too_many_bad_pixels = false;
+
+    /// Flag set if the center did not lie within the SpanSet.
+    bool center_out_of_bounds = false;
+
+    /// Flag set if there was a bad pixel too close to the center.
+    bool bad_pixel_in_center = false;
+
+    /// Flag set if the second moments do not resolve to an ellipse.
```
laurenam: Tense is changed from above, i.e. “do not” vs. “did not” & “was”.

include/lsst/meas/algorithms/SpanSetMoments.h
```diff
@@ -0,0 +1,131 @@
...
+    bool singular_second_moments = false;
+
+    /// Test whether any failure falg is set.
```
laurenam: “falg” —> “flag”

include/lsst/meas/algorithms/SpanSetMoments.h
```diff
@@ -0,0 +1,131 @@
...
+
+    /// Return a flattened array of the y coordinates in `spans`.
+    ndarray::Array<float, 1, 1> get_y_array() const;
+
+    /**
+     * Compute the unweighted moments of an image within a SpanSet.
+     *
+     * @param[in] spans          Pixel region to use.
+     * @param[in] masked_image   Image to measure.
+     * @param[in] bad_bitmask    Mask of bad pixels to remove from `spans` to
```
laurenam: Extra “to”?

include/lsst/meas/algorithms/SpanSetMoments.h
```diff
@@ -0,0 +1,131 @@
...
+     *                           before computing the moments.
+     * @param[in] bad_pixel_mask_fraction     Maximum fraction of the pixels
+     *                                        in `spans` that can be bad
+     *                                        before giving up.
+     * @param[in] bad_pixel_exlusion_radius   Radius around the estimated
+     *                                        center where bad pixels will
+     *                                        cause the algorithm to give up.
```
laurenam: I don’t quite follow this description…

laurenam (commented):
Just some flyby comments while perusing…

python/lsst/meas/algorithms/computePsfChiSq.py
```diff
@@ -0,0 +1,90 @@
...
+class ComputePsfChiSqConfig(Config):
+    padding = Field(
+        "Number of pixels to grow the PSF model bbox by on all sides.",
+        dtype=int,
+        default=10,
+    )
+    bad_mask_planes = ListField(
+        "Mask planes to identify pixels to drop from the calculation.",
+        dtype=str,
+        default=["SAT", "SUSPECT"],
```
laurenam: Any reason not to include `BAD` and/or `NO_DATA` in the list? (I always forget what `SUSPECT` includes.)

python/lsst/meas/algorithms/computePsfChiSq.py
```diff
@@ -0,0 +1,90 @@
...
+    )
+
+
+class ComputePsfChiSqTask(Task):
```
laurenam: Docs?

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
```diff
@@ -0,0 +1,785 @@
...
+    )
+    bad_pixel_exclusion_radius = Field(
+        "If a bad pixel (according to bad_mask_planes) falls within this "
+        "radius of the unweighted centroid of a footprint, drop that footprint "
+        "entirely.",
+        dtype=float,
+        default=2.0,
+    )
+    max_footprint_area = Field(
+        "Footprints larger than this threshold are dropped before computing moments.",
```
laurenam: “Footprints whose area has a pixel count larger…”?

python/lsst/meas/algorithms/spanSetMoments.cc
```diff
@@ -0,0 +1,67 @@
...
+namespace algorithms {
+
+void wrapSpanSetMoments(lsst::cpputils::python::WrapperCollection& wrappers) {
+    using PySpanSetMoments = py::classh<SpanSetMoments>;
+    auto clsSpanSetMoments =
+            wrappers.wrapType(PySpanSetMoments(wrappers.module, "SpanSetMoments"), [](auto& mod, auto& cls) {
+                cls.def_readonly("flux", &SpanSetMoments::flux);
+                cls.def_readonly("variance", &SpanSetMoments::variance);
+                // def_readonly doesn't ever seem to do the right thing w.r.t.
+                // memory management, even when explicit return value policies.
```
laurenam: Sentence (after the comma) needs some rewording.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
```diff
@@ -0,0 +1,785 @@
...
+    import matplotlib.figure
+    import matplotlib.image
+    import matplotlib.patches
+
+
+class NoStarsForShapeletsError(AlgorithmError):
+    """Exception raised when ComputeRawPsfMomentsTask fails to find any usable
+    stars.
+    """
+
```
laurenam, line 56: Unnecessary blank line?

include/lsst/meas/algorithms/SpanSetMoments.h
TallJimbo: It's a radius that tries to act analogously to a cut on `base_PixelFlags_[bad_bitmask]Center`, but about the center the algorithm is computing itself.  I've reworded slightly.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
```diff
@@ -0,0 +1,785 @@
...
+            self.log.verbose(
+                "Applying a %s %s %f cut yields only %d sources; keeping the top %d (%s %s %f) instead.",
+                name,
+                kind,
+                threshold,
+                n,
+                min_count,
+                name,
+                kind,
+                values[sorter[min_count - 1]],
```
laurenam, line 713: In adding `ComputeRoughPsfShapeletsTask` as subtask in `calibrateImage` over on DM-54482, I got this in running the tests:

```
File "/sdf/data/rubin/user/laurenma/LSST/meas_algorithms/python/lsst/meas/algorithms/computeRoughPsfShapelets.py", line 709, in _threshold_with_bounds
    values[sorter[min_count - 1]],
           ~~~~~~^^^^^^^^^^^^^^^
IndexError: index 9 is out of bounds for axis 0 with size 4
```

I think `values[sorter[min_count - 1]]` needs to be something like `values[sorter[min(min_count, len(values)) - 1]]`.  Do we want to add a hard minimum for this task to pass?

laurenam (commented):
A few more fly-by comments...

python/lsst/meas/algorithms/computePsfChiSq.py
TallJimbo: Oops, the commit that adds this task was not supposed to be in the PR, as I didn't think it was useful in the end.  I deleted it, and then it accidentally came back in a rebase when I was transferring the code between USDF and my laptop.  I'll go delete it again.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
TallJimbo: No, I think there's always supposed to be a blank line after the docstring.  It might be one of those things that I'm used to having ruff complain about or fix while flake8 doesn't care.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
TallJimbo: The `min_count` values were supposed to be that hard minimum, but as you noticed the logic was bad.  Fixed in a new commit.

This means we'll raise an `AlgorithmError` in these cases, and that might mean `CalibrateImageTask` should catch that in a lower-level try block (rather than the big one in `runQuantum`) if it wants to proceed anyway when this new task can't do its job.  Whether that's a good idea or not depends on whether this failure mode happens on good (but e.g. crowded or super-shallow-u) data.

TallJimbo (commented):
I've just deleted a file that wasn't supposed to be here.  Sorry!

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
```diff
@@ -0,0 +1,789 @@
...
+        """Find the first peak in a 1-d distribution of radii."""
+        kde = KernelDensity(bandwidth=self.config.radius_kde_bandwidth).fit(radii.reshape(-1, 1))
+        sorted_radii = radii.copy()
+        sorted_radii.sort()
+        sorted_radii = sorted_radii[sorted_radii.searchsorted(self.config.radius_mode_min):]
+        scores = kde.score_samples(radii.reshape(-1, 1))
+        peaks, _ = scipy.signal.find_peaks(scores)
+        if not peaks.size:
+            raise NoStarsForShapeletsError("Radius distribute has no mode.")
+        return sorted_radii[peaks.min()], kde
```
ctslater, line 743: I don't understand how `peaks` can be used to index into `sorted_radii`, since the scores were created from the original `radii`?

include/lsst/meas/algorithms/SpanSetMoments.h
```diff
@@ -0,0 +1,132 @@
...
+     * Fit a common shapelet expansion to multiple sources whose moments have
+     * already been computed.
+     *
+     * The expected use case is a sample of stars with similar but not
+     * identical moments, which might plausibly have an approximately common
+     * shapelet representation while benefitting from allowing the ellipse
+     * used for the expansion to vary (i.e. to partially capture spatial
+     * variation in the PSF).
+     *
+     * @param[in] masked_image    Image ot measure.
```
ctslater:
```suggestion
     * @param[in] masked_image    Image to measure.
```

src/SpanSetMoments.cc
```diff
@@ -0,0 +1,172 @@
...
+    ndarray::Array<float, 1, 1> data_array = ndarray::allocate(n_pixels);
+    data_array.deep() = 0.0;
+    // We allocate the design matrix via its transpose, because that yields
+    // the memory order expected by the shapelet library and the better
+    // memory order for the least-squares solver.
+    ndarray::Array<float, 2, -2> design_matrix =
+            ndarray::Array<float, 2, 2>(ndarray::allocate(n_coeff, n_pixels)).transpose();
+    design_matrix.deep() = 0.0;
+    std::size_t start = 0;
+    for (auto const& source_moments : moments) {
```
ctslater: It took me a long time to decode that this is a loop over sources. The naming is tricky here because the relevant values per-source are moments, and so a vector of those can fairly be called "moments", but that doesn't leave any room in the naming scheme for saying what distinguishes these elements from each other. I'm tempted to suggest the input parameter be named `sources`? I'm not sure it's much better but it's the best alternative I can think of.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
```diff
@@ -0,0 +1,789 @@
...
+    logarithmic_shapes = Field(
+        "If True, transform the (xx, yy, xy) moments to conformal shear "
+        "and log trace radius when selecting stars in order to map the ellipse "
+        "parameter to a space with (-inf, inf) bounds on all quantities (but "
+        "a less-linear relationship to the pixel data).",
+        dtype=bool,
+        default=False,
+    )
+    radius_mode_min = Field(
+        "Minimum radius at which to start searching for the first mode. "
```
ctslater: Units are pixels?

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
```diff
@@ -0,0 +1,789 @@
...
+        dtype=float,
+        default=1.0,
+    )
+    shapelet_order = Field(
+        "Order of the shapelet expansion fit to the stars.",
+        dtype=int,
+        default=4,
+    )
+    shapelet_scale_factor = Field(
+        "Scale fractor to apply to the moments ellipse when computing the ellipse for the shapelet basis.",
```
ctslater:
```suggestion
        "Scale factor to apply to the moments ellipse when computing the ellipse for the shapelet basis.",
```
I don't really understand the purpose of this. Is this a "just in case" type of config, or is this something that's going to need measuring in downstream tickets to figure out how the moments relate to the donut size?

ctslater (approved):
Looks good, just a few questions on things I didn't understand.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
TallJimbo: Yup, that's a bug; good catch!   `scores` should have been created from `sorted_radii`.

I think that says that this logic isn't super important, and that sort of makes sense, because this is about keeping galaxies and blends out of the star sample. I think at S/N=50 we _usually_ mostly get stars without trying very hard, and the rest of the robust statistics can kind of take it from there.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
TallJimbo: :+1:, will update.

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
TallJimbo: It was a parameter I thought I might need while first writing the code, i.e. if the radii coming out of the moments looked systematically a lot smaller or larger than the donuts.  But it's not something that I think merits a lot of effort downstream now, since even the one bad visit I tested on had a lot of different PSF morphologies and the radii looked pretty good for all of them.

But if we did want to optimize this, I think it might make sense to look for the value that (on average) zeros out the (2, 0) shapelet coefficient for in-focus PSFs; 1.0 is formally the right value for Gaussian PSFs, and to the extent that our good PSFs are not Gaussian I expect the (2, 0) term to do most of the work in trying to compensate, as that's the lowest-order term that affects the size.  I would expect this to be a small adjustment.

src/SpanSetMoments.cc
TallJimbo: I've renamed the argument to `sources` and the loop variable to `source`; adding "moments" to either doesn't add much information for the reader, even if it's correct.

PR #471, opened 2026-05-05
DM-54482: Integrate PSF-shapelet subtask into calibrateImage as a donut-detector

laurenam:


TallJimbo (approved):

python/lsst/meas/algorithms/computeRoughPsfShapelets.py
```diff
@@ -459,6 +481,105 @@ def select_stars(self, catalog: SourceCatalog, seed: int) -> Struct:
...
+        shapelet_coeffs = results.shapelet.getCoefficients()
+        shapelet_coeffs_list = [float(coeff) for coeff in shapelet_coeffs]
+        shapelet_summary.psfStarShapeletCoeffs = shapelet_coeffs_list
+        shapelet_used_cat = results.catalog[
+            (results.catalog[self.shapelet_base_name + "_used"])
+            & (~results.catalog[self.shapelet_base_name + "_flag"])].copy(deep=True)
+        shapelet_summary.nShapeletStar = len(shapelet_used_cat)
+        centroid_diff_shapelet_vs_slot = np.sqrt(
+            (shapelet_used_cat["slot_Centroid_x"] - shapelet_used_cat[self.shapelet_base_name + "_x"])**2.0
+            + (shapelet_used_cat["slot_Centroid_y"] - shapelet_used_cat[self.shapelet_base_name + "_y"])**2.0
```
TallJimbo: I think this is the only point at which `ComputeRoughPsfShapeletsTask` requires `SingleFrameMeasurementTask` to have been run already.  What happens right now when we have a failure during measurement (especially an all-centroids-failed error)?  I don't really have any great ideas for how to deal with this, but while I like including the centroid offset in the score most of the time, I don't like the idea that this stops us from computing the score when we can't centroid now.

laurenam: I’ve yet to hit this case, but it’s a good point!  The only idea I have at the moment is to define two shapelet scores; one that only includes the non-atmospheric power and one also including the centroid offset whenever possible (and the latter could be set to `nan` if the centroiding in failed).

TallJimbo: Yeah, that's what I was coming around to as well.  Sorry to give you another field to add, but I think that's what we ought to do.

laurenam: No probs.  Any thoughts on naming?  I was thinking of `shapelets_score_non_atm` & `shapelets_score`, with the latter being the “authoritative” one when available (I’m not sure if it’s more or less confusing to try to add more info to the name of the latter…)

TallJimbo: How about `shapelets_only_iq_score` and `shapelets_iq_score`?

laurenam: Ok, I've implemented a second method to compute the `shapelets_iq_score` after the `measurement` run of the PSF catalog.  In doing so, it occurred to me why I never his this failure mode...I don't actually think it's possible.  The reason being that the SDSS centroid algorithm has the config:
```
# If >0; maximum distance in pixels between the footprint peak and centroid allowed before resetToPeak flag is set.
config.star_measurement.undeblended['base_SdssCentroid'].maxDistToPeak=1.0
```

which, for `calibrateImage` is set to 1 pixel.  So, if the centroider would like to wonder more than one pixel from the original peak, it reverts back to that peak.  So, while this is flagged and hence formally a centroiding failure, the `slot_Centroid_x` & `slot_Centroid_y` columns always have an entry.  I have not been checking for centroid flags in any of this since they are expected to be set for very non-Gaussian shapes (i.e. funky PSFs).

I do think it's a good idea to include both "scores", but I'm curious what your thoughts are on the above...

TallJimbo: Good point. I think there might be some very narrow room for NaN centroid columns - i.e. failures before the centroider runs - but I'm not surprised we don't see any of those in practice right now.  But I'm also in favor of keeping both scores to hedge against that.

TallJimbo (approved):

PR #473, opened 2026-05-24
DM-55017: Add accommodations for crowded fields in coadd assembly and detection

laurenam:


python/lsst/meas/algorithms/detection.py
```diff
@@ -819,8 +833,11 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
                 if ``reEstimateBackground==False``.
                 (`lsst.afw.math.BackgroundList`)
             ``factor``
-                Multiplication factor applied to the configured detection
-                threshold. (`float`)
+                Multiplication factor applied to the configured threshold
+                for positive detectin polarity. (`float`)
```
kherner: Doesn't affect anything, but there's a typo in detection ("detectin") here.

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -605,16 +619,34 @@ def detectFootprints(self, exposure, doSmooth=True, sigma=None, clearMask=True,
...
-                                     "[%.2f, %.2f].  Setting tweak to lower limit: %.2f.", bgLevel,
-                                     self.config.minBackgroundTweak, self.config.maxBackgroundTweak,
-                                     self.config.minBackgroundTweak)
-                    bgLevel = self.config.minBackgroundTweak
+                if self.config.minBackgroundTweak:
+                    minBackgroundTweak = self.config.minBackgroundTweak
+                    # Increase the minimum bg tweak allowed if the factorNeg indicates image
+                    # was very oversubtracted.
+                    if factorNeg > 5*factor:
+                        if factorNeg > 8*factor:
```
kherner, line 636: Again pardon my ignorance, but what drives the choices of 5 and 8 here? They seem reasonable on their faces anyway.

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -691,49 +723,76 @@ def _computeBrightDetectionMask(self, maskedImage, convolveResults):
...
             self.log.info("Number (%) of bright DETECTED_NEGATIVE pix: {} ({:.1f}%)".
-                          format(nPixDetNeg, 100*nPixDetNeg/nPix))
-            if nPixDetNeg/nPix > brightMaskFractionMax or nPixDet/nPix > brightMaskFractionMax:
-                self.log.warning("Too high a fraction (%.2f > %.2f) of pixels were masked with current "
-                                 "\"bright\" detection round thresholds (at nIter = %d). Increasing by "
-                                 "a factor of %.2f and trying again.", max(nPixDetNeg, nPixDet)/nPix,
-                                 brightMaskFractionMax, nIter, self.config.bisectFactor)
+                          format(nPixDetNeg, 100*nPixDetNeg/nGoodPix))
+
+            if nPixDetPos/nGoodPix > brightMaskFractionMax or nPixDetNeg/nGoodPix > brightMaskNegFractionMax:
```
kherner, line 769: Super pedantic, but is there ever any risk that nPixDetPos/nGoodPix (also for the negative version) ends up doing integer/integer division? I think it's OK as long as the underlying np.sum(...) elements aren't returning integers, but it might be a good idea to multiply one of them by 1.0 to force it.

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -691,49 +723,76 @@ def _computeBrightDetectionMask(self, maskedImage, convolveResults):
...
                                      "likely mostly masked with BAD or NO_DATA or \"bad\" in some "
                                      "other respect (so expected to likely fail further downstream).")
+                    break
+                if nPixDetPos/nGoodPix > brightMaskFractionMax:
+                    brightFactorPos *= self.config.bisectFactor
+                    self.log.warning("Too high a fraction (%.2f > %.2f) of pixels were masked as "
+                                     "DETECTED with current \"bright\" detection round thresholds "
+                                     "(at nIter = %d). Increasing to a factor of %.2f and trying again.",
+                                     nPixDetPos/nGoodPix, brightMaskFractionMax, nIter, brightFactorPos)
+                if nPixDetNeg/nGoodPix > brightMaskNegFractionMax:
```
kherner, line 782: Same comment as line 769.

python/lsst/meas/algorithms/dynamicDetection.py
```diff
@@ -691,49 +723,76 @@ def _computeBrightDetectionMask(self, maskedImage, convolveResults):
...
+                    extraFactorNeg = 1.2
+                    if nPixDetNeg/nGoodPix > min(0.98, 1.25*brightMaskNegFractionMax):
```
kherner, line 784: One more

kherner (approved):

kherner (approved):
Generally looks fine; see my question on whether one could end up in an int/int division problem in a couple of cases. Python is generally smart about that sort of thing but I learned Fortran and C/C++ first, so I think of things that way.

python/lsst/meas/algorithms/dynamicDetection.py
kherner: Never mind, sometimes I forget that Python is smarter than other languages about this.

python/lsst/meas/algorithms/dynamicDetection.py
laurenam: No ignorance on your part at all!  This is again just based on what I see when testing.  The original limits were set similarly (i.e. just based on a bunch of testing), and were imposed to avoid runaway "corrections".  The more experience I get with our pipeline's behavior with crowded fields (the ones I'm trying to accommodate here weren't even passing the initial stage of this task prior to this ticket!), the more I see where further tweaking is required/warranted so, indeed, I just set some values that looked reasonable in a bunch of test tracts.  I definitely intend to keep a close eye on this (and may eventually consider making the behavior configurable, but I'd be very hesitant to do that just yet).  [And as a side note, Jim Bosch (+ many others, I suspect), are looking forward to a day where we have something much more principled for the coadd detection such that we can do away with this convoluted task altogether.  Fred is working on that, but for now this is all we've got...]

python/lsst/meas/algorithms/detection.py
```diff
@@ -649,25 +649,31 @@ def finalizeFootprints(self, mask, results, sigma, factor=1.0, factorNeg=None, g
...
+        else:
+            nSigmaToGrow = self.config.nSigmaToGrow
         factorNeg = factor if factorNeg is None else factorNeg
         for polarity, maskName in (("positive", "DETECTED"), ("negative", "DETECTED_NEGATIVE")):
             fpSet = getattr(results, polarity)
             if fpSet is None:
                 continue
-            if self.config.nSigmaToGrow > 0:
+            if nSigmaToGrow > 0:
                 nGrow = int((self.config.nSigmaToGrow * sigma) + 0.5)
```
yalsayyad, line 677: @laurenam I know this is closed already, but heads up it looks like this line either undoes or misreports the `nSigmaToGrow = growOverride` that was added above.  (At pair coding today, I think Dan told me that growOverride wasn't in use anywhere anyway)

laurenam: Garhhh...you're right.  Will fix.  This commit was meant to remedy the issue that Dan pointed out...

PR #474, opened 2026-05-29
DM-54947: Add reset method to AccumulatorMeanStack.

TallJimbo:


tests/test_accumulator_mean_stack.py
```diff
@@ -239,6 +239,23 @@ def test_online_coadd_image(self):
...
+        online_image = coadd_exposure.image
+
+        # The unmasked coadd good pixels should match at the <1e-5 level
+        # The masked pixels will not be correct for straight image stacking.
+        good_pixels = np.where(afw_masked_image.mask.array == 0)
+
+        testing.assert_array_almost_equal(online_image.array[good_pixels],
+                                          afw_masked_image.image.array[good_pixels],
+                                          decimal=5)
+
```
arunkannawadi, line 258: One good test to add would to be make sure that after `reset` , the object is identical to a newly initiated object with the same parameters (or just implement `reset` that way). This will act as a regression test if someone included a new attribute that doesn't get reset.

TallJimbo: Did you have a more specific idea for how to do that?  The best I've got would be adding comparisons of all attributes to the test, or adding an `__eq__` (but it doesn't seem like the kind of class that should have one).  And that just punts the "forget to deal with a new attribute" to a new location.

arunkannawadi: No, unfortunately other than going to the list of attributes I don't have a concerete idea.

arunkannawadi (approved):

PR #479, opened 2026-06-17
DM-55220: add accessors for template provenance to CoaddPsf

TallJimbo:


include/lsst/meas/algorithms/CoaddPsf.h
```diff
@@ -172,7 +173,21 @@ class CoaddPsf : public afw::table::io::PersistableFacade<CoaddPsf>, public Imag
...
+     * @throws lsst::pex::exceptions::NotFoundError  There is no tract column.
+     */
+    int getTract(int index) const;
+
+    /**
+     * Get the patch ID of the component image at the given index.
+     *
+     * @throws lsst::pex::exceptions::NotFoundError  There is no patch column.
+     */
+    int getPatch(int index) const;
```
isullivan, line 190: I don't see how `getTract` and `getPatch` actually get set. `src/CoadPsf.cc` initializes with a minimal schema, which does not include patch or tract columns, so it looks like the exposure catalog won't actually have tract or patch columns.

isullivan (approved):
Only one issue, which is that it looks like `getTract` and `getPatch` can't get their tract and patch values.

include/lsst/meas/algorithms/CoaddPsf.h
TallJimbo: There are a few different constructors for `CoaddPsf`, and you're right about most of them.  But there's one - the one we actually use in `GetTemplateTask` - that provides more control over the schema.

From the perspective of `CoaddPsf`, this usage was not expected or intentional.  But it happens to be where we've stuffed a lot of useful information in all of the DP2 files.  I am looking forward to being able to replace it with something more intentional in the future when we make `GetTemplateTask` write an `lsst.images` output directly.

PR #480, opened 2026-06-22
DM-55282: Add nCrSpanMax configuration to cosmic ray finder for early failure.

erykoff:


python/lsst/meas/algorithms/findCosmicRaysConfig.py
```diff
@@ -35,6 +35,11 @@ class FindCosmicRaysConfig(pexConfig.Config):
         doc="maximum number of contaminated pixels",
         default=1000000,
     )
+    nCrSpanMax = pexConfig.Field(
+        dtype=int,
+        doc="Maximum number of contaminated spans",
+        default=100_000,
```
laurenam, line 41: I like the `_`, but consistency with above... 🫠

src/CR.cc
```diff
@@ -495,6 +496,14 @@ std::vector<std::shared_ptr<afw::detection::Footprint>> findCosmicRays(
         }
     }
 
+    /*
+     * Check if number of spans is above threshold to fail fast
+     */
+    if (spans.size() > nCrSpanMax) {
+        throw LSST_EXCEPT(pex::exceptions::LengthError,
+                          (boost::format("Too many CR spans (max %d)") % nCrSpanMax).str());
+    }
```
laurenam, line 505: I take it the 
```
 throw LSST_EXCEPT(pex::exceptions::LengthError,
                                  (boost::format("Too many CR pixels (max %d)") % nCrPixelMax).str());
```
at line ~410 doesn't trigger often enough.  I don't fully follow the code logic here, but if that initial pass is certain to detect a smaller number than the final one, is it worth considering giving it a smaller threshold for "failing fast"?

laurenam (approved):

src/CR.cc
erykoff: Oh wow, I didn't notice that ... this ticket might not be necessary ... in this case that is coming up with 472,977 pixels (first pass) and then later that is much larger.

Maybe we just need to override the max pixels to something less insane than 1,000,000.  I mean ... if we ever have 100,000 CR pixels that says something has gone terribly wrong.

